PR #8, opened 2022-03-18
fix(datagrid): do not run change detection when the datagrid height is recomputed

arturovt:


Jinnie (approved):

mathisscott (approved):

mathisscott:
Thanks, @arturovt ! 

🇺🇦

PR #9, opened 2022-03-18
feat: implement the `ClrDestroyService` to reduce subscription boilerplate

arturovt:


mathisscott (approved):

steve-haar:
I wonder if we should take the time to apply this approach in all our components so the code base will be consistent.

mathisscott:
> I wonder if we should take the time to apply this approach in all our components so the code base will be consistent.

That would be a good long-term objective, imo.

arturovt:
> I wonder if we should take the time to apply this approach in all our components so the code base will be consistent.

I can take that.

steve-haar (approved):

PR #10, opened 2022-03-18
fix(popover): run change detection only when the `Esc` is clicked

arturovt:


package.json
```diff
@@ -24,7 +24,7 @@
     "public-api:update": "api-extractor run -l",
     "start": "ng serve clr-demo",
     "test": "ng test clr-angular --source-map=false --watch=false",
-    "test:watch": "npm run --silent _test:angular -- --watch",
+    "test:watch": "ng test crl-angular --watch",
```
mathisscott, line 27: `crl-angular`?

mathisscott (changes requested):

package.json
arturovt: Whoops 😄

mathisscott (approved):

package.json
```diff
@@ -24,7 +24,7 @@
     "public-api:update": "api-extractor run -l",
     "start": "ng serve clr-demo",
     "test": "ng test clr-angular --source-map=false --watch=false",
-    "test:watch": "npm run --silent _test:angular -- --watch",
+    "test:watch": "ng test clr-angular --watch",
```
steve-haar, line 27: This is fixed in https://github.com/vmware-clarity/ng-clarity/pull/23, so please remove this from this PR.

steve-haar (changes requested):

package.json
arturovt: Done

steve-haar (approved):

PR #12, opened 2022-03-18
(draft/Don't merge) fix: datagrid > cell not associated with header vpat-5955

bdryanovski:
Fix VPAT-5955 issue


mathisscott (changes requested):
Should we add any tests for this?

bdryanovski:
@mathisscott it seems you are right, our initial test case with @Jinnie was super limited so this won't work for us.

There are cases where `page` is not the best option also `totalitems` is not even working in all cases. But everything else is throwing `ExpressionChangedAfterItHasBeenCheckedError` because the data is coming on the `ngAfterContentInit`.

Another solution is in progress at the moment so until then will mark this as draft.

PR #17, opened 2022-03-18
fix(drag-and-drop): re-enter the zone when the `clrIfDragged` creates an embedded view

arturovt:


projects/angular/src/utils/drag-and-drop/if-dragged.ts
```diff
@@ -22,20 +22,25 @@ export class ClrIfDragged<T> implements OnDestroy {
...
+        // This won't trigger another change detection if it's already being run in the zone.
+        this.ngZone.run(() => this.container.createEmbeddedView(this.template));
       })
     );
     this.subscriptions.push(
       this.dragEventListener.dragEnded.subscribe(() => {
-        this.container.clear();
+        // Note: the `dragEnded` emits outside of the Angular zone, see `DragEventListenerService#broadcast`.
+        // This won't trigger another change detection if it's already being run in the zone.
+        this.ngZone.run(() => this.container.createEmbeddedView(this.template));
```
mathisscott, line 43: It doesn't seem great that the function inside the subscription is copy-pasted. Would you consider DRYing the code out?

mathisscott (changes requested):

projects/angular/src/utils/drag-and-drop/if-dragged.ts
arturovt: I don't think it's possible to decouple the `ngZone.run`, I could move it into the method but that would overcomplicate the code much more.

mathisscott (approved):

Shijir:
Hi @arturovt, thank you for this contribution. Just to give some context here, the embedded view doesn't need to re-enter the zone because `clrIfDragged` is only used internally for `ClrDraggable` to create its drag-ghost element. It needs to be as light as possible performance-wise as its only purpose is to create _fixed_ visual duplicate of the draggable element. During the drag process, nothing should change for the drag ghost element. 

But I am also curious why you thought it needs to re-enter the zone.

steve-haar (approved):

PR #20, opened 2022-03-21
docs: fix links and dev env flow doc

bbogdanov:
Signed-off-by: Bogdan Bogdanov <bbogdanov@vmware.com>


docs/BUILD.md
```diff
@@ -25,7 +25,7 @@ projects
 
 ## Build Process
 
-The build scripts in this repository have been designed to work with Node.js v17 or later.
+The build scripts in this repository have been designed to work with Node.js v16 or later.
```
bbogdanov, line 28: Because of the `.nvmrc`

projects/demo/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@clr/demo",
+  "private": true,
```
bbogdanov, line 3: Since this is not an app we would like to set in NPM

docs/BUILD.md
Jinnie: @steve-haar had some comments comments about the version of Node we need to use. I didn't understand if the requirement was updated in the `.nvmrc` too. Would be good to have some comment from him too.

steve-haar: The project has been set up to work with node 16. This is just a mistake in the documentation. Thanks for finding it.

projects/demo/package.json
steve-haar: I don't think we want this. From my understanding, setting this will prevent the package from publishing: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#private

steve-haar (changes requested):

projects/demo/package.json
coryrylan: This is the demo project, so I think this is correct as we would want it to prevent from publishing to npm

coryrylan (approved):

ashleyryan:
I addressed some of this in my PR from Friday - you'll need to rebase: https://github.com/vmware-clarity/ng-clarity/pull/11/files

projects/demo/package.json
steve-haar: Oh, sorry, I missed that this is the demo. Yes, I agree with this.

steve-haar (approved):

PR #25, opened 2022-03-21
fix(datagrid): do not run change detection when the input is focused asynchronously

arturovt:


projects/angular/src/data/datagrid/built-in/filters/datagrid-string-filter.spec.ts
```diff
@@ -42,7 +42,7 @@ const PROVIDERS = [
 ];
 
 export default function (): void {
-  describe('DatagridStringFilter component', function () {
+  fdescribe('DatagridStringFilter component', function () {
```
steve-haar, line 45: Looks like you forgot an `fdescribe`

arturovt: Whoops

steve-haar:
Looks like there's a lint error in `datagrid-string-filter.ts`. Please run `npm run lint:fix` to fix. For future reference, there's also `npm run lint:changed:fix` for modified and uncommitted files. More details here: https://github.com/vmware-clarity/ng-clarity/blob/main/docs/BUILD.md

steve-haar (approved):

mathisscott (approved):

PR #27, opened 2022-03-21
feat(popover): mark the `scroll` event listener as passive and do not run change detection until the condition is met

arturovt:


steve-haar:
@arturovt, if the public api changes, you need to run `npm run public-api:update` to update the "golden" api file so the build will pass.

mathisscott (approved):

projects/angular/src/utils/popover/providers/popover-events.service.ts
```diff
@@ -92,26 +95,30 @@ export class ClrPopoverEventsService implements OnDestroy {
...
-          // Rather than a complex change to the focus trap I put focus on the element that was clicked
-          const clickedElement = event.target as HTMLElement;
-          clickedElement.focus();
-        }
-      });
+      this.documentClickListener = this.ngZone.runOutsideAngular(() =>
+        this.renderer.listen(this.document, 'click', (event: MouseEvent) => {
+          if (event === this.ignoredEvent) {
+            // Here we ignore the opening click event (w/o this content opens and immediately closes.
+            this.ignoredEvent = null;
```
bbogdanov, line 102: Curious why using `null` instead of `delete` as we used to?

arturovt: Hey, I wouldn't say there's super specific reason for this one, this is one of the old habbits. This was one of the recommendations from the V8 team on using `delete`:
> However, the delete operator will still result in a significant performance hit on property access if a property other than the most recently added property is deleted from the object. So overall we have to continue to recommend against using delete.

I can revert it back to `delete this.ignoredEvent`.

projects/angular/src/utils/popover/providers/popover-events.service.ts
```diff
@@ -42,16 +43,18 @@ export class ClrPopoverEventsService implements OnDestroy {
...
-  private scrollSubscription: Subscription;
+  private scrollSubscription: Subscription | null = null;
   public addScrollListener() {
     if (this.scrollToClose) {
-      this.documentScroller = fromEvent(this.document, 'scroll', { capture: true });
-      this.scrollSubscription = this.documentScroller
-        .pipe(filter(this.testForSmartPopoverContentContainer))
-        .subscribe(() => {
-          this.smartOpenService.open = false;
+      const documentScroller$ = fromEvent(this.document, 'scroll', { capture: true, passive: true });
```
Jinnie, line 49: Why do we set `passive` on the `scroll` event itself? 
If I understood it correctly, it's usually used with other events that interfere with scrolling (like `touchstart`, `click`..) to not wait for their completion to start scrolling. I don't get the idea of using in on `scroll`.

arturovt: Initially, this was intended to be used with `scroll` and `touch*` events. The purpose of passive event listeners is to unblock the composite thread (while the handler which is being executed doesn't block it).

This is forcerly set to `true` in latest Chromium versions only for the `document`:
```cpp
bool EventTarget::IsTopLevelNode() {
  ...
  if (node->IsDocumentNode())
    return true;
  }
}

if (IsWheelScrollBlockingEvent(event_type) && IsTopLevelNode()) {
  if (!options->hasPassive()) {
    options->setPassive(true);
  }
}
```
(See that it's checking if the `node` is `document` and it's a scroll event).

This is not set to `true` by default in Servo (Mozilla engine) and WebKit.

Jinnie:
Closing, as due to internal refocusing we can't afford to spend much time on these optimizations.
Thanks a lot for your effort, we may come back to these PRs in future iterations.

PR #33, opened 2022-03-22
feat(drag-and-drop): mark `touchend` as passive and setup touch listeners only on touch devices

arturovt:


mathisscott (approved):

projects/angular/src/utils/drag-and-drop/providers/drag-event-listener.service.ts
```diff
@@ -150,8 +169,11 @@ export class DragEventListenerService<T> {
       });
 
       // Listen to mouseup/touchend events.
-      this.nestedListeners.push(
-        this.renderer.listen('document', endOnEvent, (endEvent: MouseEvent | TouchEvent) => {
+      // We use `fromEvent` because `renderer.listen` does not support passing `AddEventListener` options.
+      // Note: since Chrome 56 now defaults document level `touchstart` and `touchmove` listeners to passive.
+      // The `touchend` is not passive by default, we never call `preventDefault()` on it, so we're safe making it passive too.
+      const subscription = fromEvent(document, endOnEvent, { passive: true }).subscribe(
+        (endEvent: MouseEvent | TouchEvent) => {
```
Shijir, line 176: As far as I understand, passive listeners only work for the following events: `touchstart`, `touchmove`, `wheel`, and `mousewheel`. 

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

Also, checkout this one: https://github.com/Polymer/polymer/issues/4961

Shijir (changes requested):

projects/angular/src/utils/drag-and-drop/providers/drag-event-listener.service.ts
arturovt: Is that related to scrolling since it doesn't mark `touchmove` as passive?

> Also, as far as I understand, passive listeners only work for the following events

All touch listeners can be passive.

```cpp
if (event_type == event_type_names::kTouchcancel ||
    event_type == event_type_names::kTouchend ||
    event_type == event_type_names::kTouchmove ||
    event_type == event_type_names::kTouchstart) {
  if (const LocalDOMWindow *executing_window = ExecutingWindow()) {
    if (const Document *document = executing_window->document()) {
      document->CountUse(options->passive()
                             ? WebFeature::kPassiveTouchEventListener
                             : WebFeature::kNonPassiveTouchEventListener);
    }
  }
}
```

Jinnie: Similar to Shijir I don't understand the need to make the end-events here passive. They would not be blocking a scroll as it's already after the scrolling started and we're already blocking the scroll intentionally.

arturovt: > We want users to execute their callback function when clrDragStart fires initially without instantly triggering scrolling.

That actually makes sense.

> Also, touchend is a nested event within touchstart. So making touchend passive without making touchstart passive first doesn't make any sense.

Those are unrelated things. The difference between passive and non-passive event listeners is how they're dispatched and handled. By default, all event listeners are registered with the `EventListenerProperties::kBlocking` type, and this means the event listener is non-passive, and the user may call `preventDefault()` (e.g. it would cancel scrolling on the `touchstart` event). If the event listener is registered with the `EventListenerProperties::kPassive` type, it'll first send an _acknowledge message_ to the compositor thread to resume its job and execute the handler on the main thread. The compositor thread is blocked until the handler is executed on the main thread for blocking event listeners.

PR #41, opened 2022-03-23
feat(tree-view): do not run change detection whenever the first visible child is focused

arturovt:


mathisscott (approved):

projects/angular/src/data/tree-view/tree.ts
```diff
@@ -30,45 +31,47 @@ import { ClrTreeNode } from './tree-node';
   `,
   providers: [TREE_FEATURES_PROVIDER, TreeFocusManagerService],
   host: {
-    '[attr.tabindex]': 'tabindex',
```
Jinnie, line 33: Nice! We don't need change detection for the `tabindex`

projects/angular/src/data/tree-view/tree.ts
```diff
@@ -30,45 +31,47 @@ import { ClrTreeNode } from './tree-node';
...
-      this.focusManagerService.focusFirstVisibleNode();
-
-      // when the first child gets focus,
-      // tree should no longer have tabindex of 0;
-      delete this.tabindex;
-    }
-  }
-
-  @ContentChildren(ClrTreeNode) private rootNodes: QueryList<ClrTreeNode<T>>;
+  @ContentChildren(ClrTreeNode) rootNodes: QueryList<ClrTreeNode<T>>;
```
Jinnie, line 74: Why do we need this to be public?

Jinnie (commented):
Thanks for this one!
The tree is a component that's seen heavy-duty usage in some scenarios by our teams, so any performance optimization here is more than welcome.
Only one minor question below and we can merge this.

projects/angular/src/data/tree-view/tree.ts
arturovt: Reverted.

Jinnie (approved):

projects/angular/src/data/tree-view/tree.ts
```diff
@@ -31,44 +32,46 @@ import { ClrTreeNode } from './tree-node';
...
+          this.focusManagerService.focusFirstVisibleNode();
+          // when the first child gets focus,
+          // tree should no longer have tabindex of 0.
+          renderer.removeAttribute(nativeElement, 'tabindex');
+        }
+      })
+    );
 
-  private subscriptions: Subscription[] = [];
+    this.subscriptions.push(subscription);
```
kevinbuhmann, line 63: I think this should be in `ngOnInit`, not the constructor. If, for whatever reason, the constructor is called, but the component is not actually initialized, the subscription would not be unsubscribed.

kevinbuhmann: Also, I think it generally bad practice to start long-running work in constructor.

Jinnie:
Thanks for updating this, @kevinbuhmann

projects/angular/src/data/tree-view/tree.ts
Jinnie: You are right. Sadly, these optimizations are out of our current focus area. I want to try merging what is okay, but I'll have to close the ones that impose some support risk, or require more extensive testing on our side.

PR #42, opened 2022-03-23
chore: reduce subscription boilerplate with `ClrDestroyService`

arturovt:


projects/angular/src/popover/dropdown/dropdown.ts
```diff
@@ -27,26 +27,22 @@ import { ClrPopoverToggleService } from '../../utils/popover/providers/popover-t
...
   ) {
-    this.subscriptions.push(dropdownService.changes.subscribe(value => (this.toggleService.open = value)));
-    this.subscriptions.push(toggleService.openChange.subscribe(() => this.cdr.markForCheck()));
-  }
-
-  @Input('clrCloseMenuOnItemClick') isMenuClosable = true;
-
-  ngOnDestroy() {
-    this.subscriptions.forEach(sub => sub.unsubscribe());
+    dropdownService.changes.pipe(takeUntil(destroy$)).subscribe(value => (this.toggleService.open = value));
```
mathisscott, line 45: same question regarding `changes` here...

mathisscott (changes requested):

mathisscott (approved):

projects/angular/src/accordion/stepper/stepper.ts
```diff
@@ -32,20 +32,20 @@ import { ClrStepperPanel } from './stepper-panel';
...
   @ContentChildren(ClrStepperPanel, { descendants: true })
   panels: QueryList<ClrStepperPanel>;
-  subscriptions: Subscription[] = [];
   form: FormGroupDirective | NgForm;
 
   constructor(
-    @Optional() private formGroup: FormGroupDirective,
-    @Optional() private ngForm: NgForm,
-    private stepperService: StepperService
+    @Optional() @Inject(FormGroupDirective) private formGroup: FormGroupDirective | null,
```
Jinnie, line 45: Why do we switch to using `@Inject` here?

arturovt: This is used to explicitly add the `| null` union, since the `formGroup` _might be null_, because of the `@Optional()` flag.

projects/angular/src/data/datagrid/render/main-renderer.ts
```diff
@@ -62,41 +62,41 @@ export class DatagridMainRenderer implements AfterContentInit, AfterViewInit, Af
...
+        this.resetDatagridHeight();
+      }
+    });
+
+    this.detailService.stateChange.pipe(takeUntil(destroy$)).subscribe(state => this.toggleDetailPane(state));
+
+    this.items.change.pipe(takeUntil(destroy$)).subscribe(() => (this.shouldStabilizeColumns = true));
   }
 
-  @ContentChildren(DatagridHeaderRenderer) private headers: QueryList<DatagridHeaderRenderer>;
```
Jinnie, line 84: Why making the headers and rows public?

projects/angular/src/layout/tabs/tabs.ts
```diff
@@ -95,15 +93,13 @@ import { ClrTabOverflowContent } from './tab-overflow-content';
...
-export class ClrTabs implements AfterContentInit, OnDestroy {
-  private subscriptions: Subscription[] = [];
-
+export class ClrTabs implements AfterContentInit {
   private get overflowPosition() {
     return this._tabLinkDirectives.filter(link => !link.inOverflow).length;
   }
 
   @ViewChild('tabContentViewContainer', { static: true, read: ViewContainerRef })
-  private set tabContentViewContainer(value: ViewContainerRef) {
```
Jinnie, line 106: why not private again?

projects/angular/src/data/datagrid/render/cell-renderer.ts
```diff
@@ -4,45 +4,47 @@
...
+      .subscribe(() => this.clearWidth());
 
-  private subscriptions: Subscription[] = [];
-
-  ngOnDestroy() {
-    this.subscriptions.forEach(sub => sub.unsubscribe());
-    if (this.stateSubscription) {
-      this.stateSubscription.unsubscribe();
-    }
+    this.columnStateChanges$
```
Jinnie, line 42: Nice! :)

Jinnie (commented):
Thanks for this contribution!
I finally finished reviewing it (sorry for the delay). After few minor questions, I'm okay to approve and merge the changes.
The cleanup pattern is rather interesting, I like it and would use it in future too, though TBH, I wouldn't prioritize myself refactoring it all over the codebase at this stage.
Great thanks for all this effort!

mathisscott (changes requested):
Putting on hold for a bit so the team can discuss.

projects/angular/src/data/datagrid/render/main-renderer.ts
arturovt: Reverted.

projects/angular/src/layout/tabs/tabs.ts
arturovt: Reverted.

projects/angular/src/accordion/stepper/stepper.ts
kevinbuhmann: I feel like adding `| null` annotations is entirely unrelated and shouldn't be part of this PR/commit.

Jinnie:
I respect all work done, but due to some internal processes and after discussion with the team, we prefer to refrain from merging this specific PR at this moment. It introduces support risks which, though minor, are not aligned with our current focus areas and we can't afford in this moment.
We may get back to it in future iterations, but for now I'm closing it.

PR #55, opened 2022-03-31
fix(navigation): a11y focus trap when navigation is open in responsive mode

bbogdanov:
Todo:
- [x] Check with Colin about the usage of cds-internal-close-button

# PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/clarity.api.md
+++ b/projects/angular/clarity.api.md
@@ -2457,6 +2457,8 @@ export class ClrHeader implements OnDestroy {
     // (undocumented)
     ngOnDestroy(): void;
     // (undocumented)
+    openNav(navLevel: number): void;
+    // (undocumented)
     openNavLevel: number;
     // (undocumented)
     resetNavTriggers(): void;
@@ -2468,7 +2470,7 @@ export class ClrHeader implements OnDestroy {
     get responsiveNavCommonString(): string;
     // (undocumented)
     get responsiveOverflowCommonString(): string;
-    // (undocumented)
+    // @deprecated (undocumented)
     toggleNav(navLevel: number): void;
     // (undocumented)
     static ɵcmp: i0.ɵɵComponentDeclaration<ClrHeader, "clr-header", never, {}, {}, never, ["*"]>;
@@ -2987,17 +2989,30 @@ export class ClrNavigationModule {
     static ɵmod: i0.ɵɵNgModuleDeclaration<ClrNavigationModule, [typeof i1_38.ClrHeader, typeof i2_28.ClrNavLevel, typeof i3_19.NavDetectionOompaLoompa, typeof i4_13.MainContainerWillyWonka], [typeof i6.CommonModule, typeof i3_2.ClrIconModule, typeof i8.ClrDropdownModule], [typeof i1_38.ClrHeader, typeof i2_28.ClrNavLevel, typeof i3_19.NavDetectionOompaLoompa, typeof i4_13.MainContainerWillyWonka]>;
 }
 
+// Warning: (ae-forgotten-export) The symbol "FocusTrap" needs to be exported by the entry point index.d.ts
+//
 // @public (undocumented)
-export class ClrNavLevel implements OnInit {
-    constructor(responsiveNavService: ResponsiveNavigationService, elementRef: ElementRef);
+export class ClrNavLevel extends FocusTrap implements OnInit {
+    // Warning: (ae-forgotten-export) The symbol "FocusTrapElement" needs to be exported by the entry point index.d.ts
+    constructor(platformId: any, responsiveNavService: ResponsiveNavigationService, elementRef: ElementRef<FocusTrapElement>, renderer: Renderer2, injector: Injector);
     // (undocumented)
     addNavClass(level: number): void;
     // (undocumented)
     close(): void;
     // (undocumented)
+    closeButtonAriaLabel: string;
+    // (undocumented)
+    protected hideCloseButton(): void;
+    // (undocumented)
+    protected hideNavigation(): void;
+    // (undocumented)
+    get isOpen(): boolean;
+    // (undocumented)
     get level(): number;
     // (undocumented)
     _level: number;
+    // (undocumented)
+    ngAfterViewInit(): void;
     // (undocumented)
     ngOnDestroy(): void;
     // (undocumented)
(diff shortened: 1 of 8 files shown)


projects/angular/src/utils/focus-trap/focus-trap.ts
```diff
@@ -0,0 +1,189 @@
+/*
```
bbogdanov, line 1: This is a copy of the focus trap we have in Clarity Core that is about to be deprecated. It works better than the `[clrFocusTrap]` so we might take a closer look down the road to reuse this rather than `[clrFocusTrap]`

projects/angular/clarity.api.md
```diff
@@ -546,8 +546,8 @@ export class ClrButton implements LoadingListener {
 }
 
 // @public (undocumented)
-export class ClrButtonGroup implements AfterContentInit {
-    constructor(buttonGroupNewService: ButtonInGroupService, toggleService: ClrPopoverToggleService, popoverId: string, commonStrings: ClrCommonStringsService, destroy$: ClrDestroyService);
+export class ClrButtonGroup {
```
bbogdanov, line 549: That's kinda strange but ok.

projects/angular/clarity.api.md
```diff
@@ -2364,14 +2356,14 @@ class ClrExpandableAnimationModule {
 }
 
 // @public (undocumented)
-export class ClrFocusOnViewInit implements AfterViewInit, OnDestroy {
-    constructor(el: ElementRef, platformId: any, focusOnViewInit: boolean, document: any, renderer: Renderer2, ngZone: NgZone);
+export class ClrFocusOnViewInit implements AfterViewInit {
```
bbogdanov, line 2359: That's kinda strange but ok.

projects/angular/src/utils/focus-trap/focus-trap.spec.ts
```diff
@@ -0,0 +1,434 @@
...
+  });
+});
+
+/**
+ * For some reason locally it runs ok, on the CI it says it doesn't.
+ *
+ * Have to revisit this but at this moment we don't use nested focus traps in our components, so...
+ * that should be okay for now.
+ */
+xdescribe('Nested FocusTraps: ', () => {
```
bbogdanov, line 346: Have to debug this one why it doesn't work on GitHub CI. Made a demo for it to test it manually, works fine but in order to not postpone this any more I'll would be happy to get this merged and revisit this in near future.

projects/angular/src/layout/nav/header.ts
```diff
@@ -102,10 +116,9 @@ export class ClrHeader implements OnDestroy {
     this.responsiveNavService.closeAllNavs();
   }
 
-  // toggles the nav that is open
-  toggleNav(navLevel: number) {
-    this.openNavLevel = this.openNavLevel === navLevel ? null : navLevel;
-    this.responsiveNavService.sendControlMessage(ResponsiveNavCodes.NAV_TOGGLE, navLevel);
+  openNav(navLevel: number) {
```
mathisscott, line 119: Renaming this method is a breaking change. Why do we need it to say "open" vs. "toggle"?

If that is something that _must_ happen, we should make toggle an alias to `openNav()` and make a note that it is deprecated...

projects/angular/src/utils/focus-trap/focus-trap.spec.ts
mathisscott: In `@cds/core` we don't nest focus-traps, we _layer_ them so that only the top-most focus-trap is active. The approach simplifies all this. I'm not sure nested focus traps would be considered accessible.

projects/angular/src/layout/nav/header.ts
bbogdanov: Good catch! Will do

The thing is that we don't want to toggle it but just send an `open` event to the responsive nav service cause the hamburger will need to send only `open` event. The close button inside the nav-level will trigger the close.

projects/angular/src/utils/focus-trap/focus-trap.ts
bbogdanov: That was the first thing I tried but the problem with it is that the [CdsBaseFocusTrap](https://github.com/vmware-clarity/core/blob/1470155a9581e31020db4e9c0397534142fabfb8/packages/core/src/internal/base/focus-trap.base.ts#L14) extends `LitElement` and that causes compile issues in angular when you try for example to do `ClrNavLevel extends CdsBaseFocusTrap`

mathisscott: Yes, but what we have here is a reverse engineer of the previous focus trap. I think it would make more sense to reverse engineer the new one...

colinreedmiller (approved):

projects/angular/src/utils/focus-trap/focus-trap.ts
bbogdanov: I believe we can do that but I would prefer to get into this when the v6 is released. What do you think of that idea?

projects/angular/src/utils/breakpoints/breakpoints.ts
```diff
@@ -7,3 +7,7 @@
 // iPad mini screen width
 // http://stephen.io/mediaqueries/#iPadMini
 export const DATEPICKER_ENABLE_BREAKPOINT = 768;
+export const SMALL_BREAKPOINT = 576;
```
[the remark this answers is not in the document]
bbogdanov: Tried to use the tokens from `@cds/core/tokens/tokens.json` but the compiler is giving me some hard time with `"resolveJsonModule": true`

I cannot us the tokens from `@cds/core/tokens/tokens.js` cause their values are the CSS variables that doesn't provide any number value I can count on.

projects/angular/src/utils/focus-trap/focus-trap.ts
mathisscott: It's pushing more work (a.k.a. tech debt) out into the future and may be harder (more expensive) to make these changes then. It's also accepting a known less-performant solution in the short term. Which may come back to bite us.

What is your level of confidence that this tech debt will be addressed? And when?

mathisscott (approved):

projects/angular/src/utils/focus-trap/focus-trap.ts
bbogdanov: When this one is merged and v6 is released I'll be spending time getting the new focus trap functionality plugged into the library. 

The reason for this merging it like this at this stage is to not postpone the critical bugfix anymore.

bdryanovski (approved):

PR #61, opened 2022-04-01
feat: add demo for tree-view-type-ahead

Shijir:
Signed-off-by: stsogoo <stsogoo@vmware.com>

## PR Checklist

## PR Type

**Enable Type-Ahead functionality for Tree-View by introducing `clrTypeAhead` directive.** 

`clrTypeAhead` directive works as a support/helper directive to enable a type-ahead focus functionality in Tree-View component. Its only goal is to extract the text content from its host element. This directive can extract and provide element's text content in two ways:

1. Accessing `.textContent` from its native element. Example:

```html
<clr-tree-node clrTypeAhead>Hello World<clr-tree-node>
```

2. Via input binding (**Recommend**). Example:

```html
<clr-tree-node [clrTypeAhead]='Hello World'>Hello World<clr-tree-node>
```

In this way, the user directly provides the text through input binding. This is better because the directive doesn't need to access the element to extract its text content.

In this current implementation, Type-Ahead functionality is optional in Tree-View as not all Tree-View components need this functionality. Thus, in order to enable this functionality, users need to import `ClrTypeAheadModule` together with `ClrTreeViewModule`.

**Recursive Tree-View with Type-Ahead:**

```html
<clr-tree>
    <clr-tree-node
      [clrForTypeAhead]="node.name"
      *clrRecursiveFor="let node of multiRoot">
      {{node.name}}
    </clr-tree-node>
</clr-tree>
```
  
**Declaritive Tree-View with Type-Ahead:**

```html
<clr-tree>
    <clr-tree-node clrForTypeAhead [clrExpanded]="true">
      California
      <clr-tree-node clrForTypeAhead>San Francisco</clr-tree-node>
      <clr-tree-node clrForTypeAhead>Los Angeles</clr-tree-node>
    </clr-tree-node>
    <clr-tree-node clrForTypeAhead [clrExpanded]="true">
      Washington
      <clr-tree-node clrForTypeAhead>Seattle</clr-tree-node>
    </clr-tree-node>
</clr-tree>
```

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


steve-haar (approved):

projects/angular/src/data/tree-view/tree-node.ts
```diff
@@ -188,6 +199,20 @@ export class ClrTreeNode<T> implements OnInit, OnDestroy {
     this.subscriptions.push(
       this._model.loading$.pipe(debounceTime(0)).subscribe(isLoading => (this.isModelLoading = isLoading))
     );
+
+    this.subscriptions.push(
+      this.typeAheadKeyEvent.pipe(debounceTime(TREE_TYPE_AHEAD_TIMEOUT)).subscribe((bufferedKeys: string) => {
+        this.focusManager.focusNodeStartsWith(bufferedKeys, this._model);
```
ashleyryan, line 205: does this match strictly on case? So if your text is "Eight" you won't match it with "eight"?

Edit: from the tests below it looks like this matches case insensitive just fine.

projects/angular/src/utils/focus/key-focus/util.ts
```diff
@@ -39,3 +39,9 @@ export function preventArrowKeyScroll(event: KeyboardEvent) {
     event.preventDefault();
   }
 }
+
+export function isKeyALetter(event: KeyboardEvent) {
+  const char = event.key;
+  // only letter characters differ when they switch between lowercase and uppercase.
+  return char.toLowerCase() !== char.toUpperCase() || (char >= '0' && char <= '9');
```
ashleyryan, line 46: this is including numbers? The name isn't very descriptive in that case

Shijir: You are right. I will find a better name and update it.

projects/angular/src/data/tree-view/tree-node.ts
Shijir: Right, it's case insensitive.

projects/angular/src/data/tree-view/tree-node.ts
```diff
@@ -188,6 +199,20 @@ export class ClrTreeNode<T> implements OnInit, OnDestroy {
     this.subscriptions.push(
       this._model.loading$.pipe(debounceTime(0)).subscribe(isLoading => (this.isModelLoading = isLoading))
     );
+
+    this.subscriptions.push(
+      this.typeAheadKeyEvent.pipe(debounceTime(TREE_TYPE_AHEAD_TIMEOUT)).subscribe((bufferedKeys: string) => {
```
Jinnie, line 204: Wonder if we could skip this subscription if we're not using the type ahead functionality.
On a tree with 10,000 nodes, this will be 10,000 subscriptions less.

projects/angular/src/data/tree-view/tree-node.ts
```diff
@@ -190,6 +201,20 @@ export class ClrTreeNode<T> implements OnInit, OnDestroy {
...
+          // reset once bufferedKeys are used
+          this.typeAheadKeyBuffer = '';
+        })
+      );
+    }
+  }
+
   ngOnDestroy() {
     this._model.destroy();
     this.subscriptions.forEach(sub => sub.unsubscribe());
```
mathisscott, line 220: It might be worth taking a look at [Artur's subscribe/unsubscribe work for consistency](https://github.com/vmware-clarity/ng-clarity/pull/9).

projects/angular/src/data/tree-view/tree-focus-manager.service.ts
```diff
@@ -99,6 +99,71 @@ export class TreeFocusManagerService<T> {
...
+    if (!model) {
+      return null;
+    }
+
+    if (model.textContent.startsWith(searchString)) {
+      return model;
+    }
+
+    if (model.expanded && model.children.length > 0) {
+      for (const childModel of model.children) {
```
mathisscott, line 112: The code pattern from lines 111-118 seems repeated in this file several times. This seems like a good opportunity to abstract that code pattern and dry up this code...

mathisscott (changes requested):

mathisscott (approved):

PR #73, opened 2022-04-12
feat(nav): add clrAriaCurrentLink directive for nav links

ashleyryan:
Addresses VPAT-628 to programmatically add aria-current="page" to the current nav link

## PR Checklist

## PR Type

**Enable  aria-current functionality for nav links using by introducing the  `clrAriaCurrentLink` directive.** 

`clrAriaCurrentLink` directive works as utility to denote the current link with `aria-current="page"` if it matches the active router link. It uses the `RouterLinkActive` director to subscribe to the change in active status.

It can be appended to any nav link that uses the `routerLink` and `routerLinkActive` directive. Works with the header, subheader, sidenav and vertical nav

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/layout/nav/aria-current-link.spec.ts
+++ b/projects/angular/src/layout/nav/aria-current-link.spec.ts
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Component } from '@angular/core';
+import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { Router } from '@angular/router';
+import { RouterTestingModule } from '@angular/router/testing';
+
+import { MainContainerWillyWonka } from './chocolate/main-container-willy-wonka';
+import { ClrNavigationModule } from './navigation.module';
+import { ClrAriaCurrentLink } from './aria-current-link';
+@Component({
+  template: `
+    <clr-header class="header">
+      <div class="header-nav">
+        <a clrAriaCurrentLink class="nav-link" id="link1" routerLink="/home" routerLinkActive="active"
+          ><span class="nav-text">Components</span></a
+        >
+        <a clrAriaCurrentLink class="nav-link" id="link2" routerLink="/patterns" routerLinkActive="active"
+          ><span class="nav-text">Patterns</span></a
+        >
+      </div>
+    </clr-header>
+  `,
+})
+class TestComponent {}
+
+describe('AriaCurrentLink Directive', () => {
+  let fixture: ComponentFixture<any>;
+  let router: Router;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      imports: [
+        ClrNavigationModule,
+        RouterTestingModule.withRoutes([
+          { path: 'home', component: TestComponent },
+          { path: 'patterns', component: TestComponent },
+        ]),
+      ],
+      declarations: [TestComponent, ClrAriaCurrentLink],
+      providers: [MainContainerWillyWonka],
+    });
+
+    fixture = TestBed.createComponent(TestComponent);
+    router = TestBed.inject(Router);
+    fixture.detectChanges();
+  });
+
+  afterEach(() => {
+    fixture.destroy();
+  });
+
+  it('decorates the active link with aria-current', fakeAsync(() => {
+    router.navigate(['/home']);
+    tick();
+    fixture.detectChanges();
+
+    const link1 = fixture.debugElement.query(By.css('#link1'));
+    const link2 = fixture.debugElement.query(By.css('#link2'));
+
+    expect(link1.nativeElement.getAttribute('aria-current')).toEqual('page');
+    expect(link2.nativeElement.getAttribute('aria-current')).toBeNull();
+
+    router.navigate(['/patterns']);
+    tick();
+    fixture.detectChanges();
+
+    expect(link1.nativeElement.getAttribute('aria-current')).toBeNull();
+    expect(link2.nativeElement.getAttribute('aria-current')).toEqual('page');
+  }));
+});
--- a/projects/angular/src/layout/nav/aria-current-link.ts
+++ b/projects/angular/src/layout/nav/aria-current-link.ts
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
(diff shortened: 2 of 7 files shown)


steve-haar:
@ashleyryan, you need to acknowledge public api changes by running `npm run public-api:update` for the build to pass

projects/angular/src/layout/nav/aria-current-link.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Directive, HostBinding, OnDestroy, OnInit } from '@angular/core';
+import { RouterLinkActive } from '@angular/router';
+import { Subscription } from 'rxjs';
+
+@Directive({ selector: '[clr-aria-current-link]', host: { '[attr.aria-current]': 'stepAriaCurrent' } })
```
Shijir, line 11: We usually use camelCase for directive selectors and kebab-case for component selectors. There are few cases of kebab-case directives as they are there to overload existing components with additional functionalities. 

Also, it doesn't need to be in a square bracket.

projects/angular/src/layout/nav/aria-current-link.ts
```diff
@@ -0,0 +1,32 @@
...
+export class ClrAriaCurrentLink implements OnInit, OnDestroy {
+  @HostBinding('attr.aria-current') ariaCurrent = undefined;
```
Shijir, line 13: You are setting host binding in two different places in two different ways.

```
host: { '[attr.aria-current]': 'stepAriaCurrent'. } 
```

and 

```
@HostBinding('attr.aria-current') ariaCurrent = undefined;
```

The first one is preferred as you don't need to import `HostBinding` decorator.

projects/angular/src/layout/nav/aria-current-link.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  private subscriptions: Subscription[] = [];
+
+  constructor(private rla: RouterLinkActive) {}
+
+  ngOnInit() {
+    this.subscriptions.push(this.listenForRouterLinkChange());
+  }
+
+  listenForRouterLinkChange() {
```
Shijir, line 23: We can have this subscription directly in the `this.subscriptions.push` as this is the only subscription in the class.

Shijir (changes requested):
Nice utility directive solution. I only have a few requests.

projects/angular/src/layout/nav/aria-current-link.ts
ashleyryan: I was trying the first and it didn't seem to be working, so I tried switching to the second and thought I deleted the first, Turns out the issue was that I wasn't subscribing to the activeChanged event. 

Good catch, I'll clean this up

projects/angular/src/layout/nav/aria-current-link.ts
ashleyryan: Updated. do you mean it doesn't need the square bracket here in the selector or when I use it.

I noticed that when I use it without the brackets, it shows up on the dom element when viewing the source as `clrariacurrentlink` and I wasn't sure if that's behavior we want/expect.

mathisscott: Given that this is an attribute directive, I think the square brackets are warranted.

mathisscott (approved):

projects/angular/src/layout/nav/aria-current-link.ts
Shijir: Sorry for being unclear on brackets. You want to keep the brackets in the selector, but you don't need it when you use it.

Shijir (approved):
LGTM!

ashleyryan:
@Shijir @mathisscott I added a unit test if you can spot check that and make sure I'm doing it correctly

projects/angular/src/layout/nav/aria-current-link.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    expect(linkElements.length).toBe(2);
+  });
+
+  it('decorates the active link with aria-current', fakeAsync(() => {
+    router.navigate(['/home']);
+    tick();
+    fixture.detectChanges();
+    let activeLink = fixture.debugElement.query(By.css('.active'));
+    expect(activeLink.nativeElement.href).toMatch(/home/);
+    expect(activeLink.nativeElement.getAttribute('aria-current')).toEqual('page');
```
mathisscott, line 69: would we want to make sure here and in the block below that the other link doesn't have `aria-current` on it?

mathisscott (changes requested):

projects/angular/src/layout/nav/aria-current-link.spec.ts
ashleyryan: I assumed using `query` instead of `queryAll` would throw an exception if more than one element was returned, but in hindsight that's checking the implementation too much (that aria-current relies on the active css class) instead of the specific behavior we want.

Updated

projects/angular/src/layout/nav/aria-current-link.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    tick();
+    fixture.detectChanges();
+
+    let currentPageElements = fixture.debugElement.queryAll(By.css('[aria-current="page"]'));
+    expect(currentPageElements.length).toEqual(1);
+    expect(currentPageElements[0].nativeElement.href).toMatch(/home/);
+
+    router.navigate(['/patterns']);
+    tick();
+    fixture.detectChanges();
```
Shijir, line 69: does this test need fakeAsync/tick here in this test?

Shijir (approved):

projects/angular/src/layout/nav/aria-current-link.spec.ts
mathisscott: Maybe give the two links unique classnames so that you can query them outside of them being active?

This is largely just to make sure that `aria-current` is unset as expected with the active link is changed.

projects/angular/src/layout/nav/aria-current-link.spec.ts
ashleyryan: It does. I don't exactly know why, but I think it's to give the router time to update and add the active status.

And then detectChanges is necessary after the active class is set for the dom to be updated after the subscription in the directive executes. It fails if I comment out either one.

projects/angular/src/layout/nav/aria-current-link.spec.ts
ashleyryan: Makes sense - updated.

mathisscott (approved):

PR #103, opened 2022-05-04
Topic/storybook

steve-haar:
This PR adds storybook. Since we are waiting on Chromatic GitHub app approval, for now storybook will deploy as a sub directory of the demo app published to netlify (/storybook).

This PR adds the storybook setup and adds stories for the alert component only. See demo here: https://103--clr-angular-demo.netlify.app/storybook


projects/angular/src/emphasis/alert/utils/alert-types.ts
```diff
@@ -4,5 +4,9 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-// @TODO Make this an enum
-export const ALERT_TYPES: string[] = ['info', 'warning', 'danger', 'success'];
+export enum AlertType {
```
bbogdanov, line 7: That's something I believe we need to avoid. While it is good for type checker, declaring the AlertType in your component so you can use it in the template its kinda bit too much.

What about if we use `type` instead of `enum`? We still keep the type checker working but we don't force people to use an enum

```
 export type AlertType = 'info' | 'warning' | 'danger' | 'success';
```

projects/angular/src/emphasis/alert/alert.ts
```diff
@@ -60,11 +61,11 @@ export class ClrAlert implements OnInit, OnDestroy {
   @Output('clrAlertClosedChange') _closedChanged: EventEmitter<boolean> = new EventEmitter<boolean>(false);
 
   @Input('clrAlertType')
-  set alertType(val: string) {
+  set alertType(val: AlertType) {
```
bbogdanov, line 64: I believe that's a huge breaking change that people are going to complain about right away.

projects/angular/src/emphasis/alert/utils/alert-types.ts
bbogdanov, line 7: My suggestion here is to turn this into

`export type AlertType = 'info' | 'warning' | 'danger' | 'success';`

While the enum is a good data structure to iterate over and keep the type checker protecting you enforcing people to provide it in the component where they are declaring an Alert it will be too much. 

Also this will cause a breaking change if we provide it as a constrain when providing the AlertType, which we do in this PR.

bbogdanov: I believe we can keep the enum but also provide a `type` and reuse the enum internally but not enforcing the users to use it.

projects/angular/src/emphasis/alert/alert.ts
bbogdanov, line 64: Breaking change that people will complain about right away. See my comments below on the file that declares the `AlertType`

bbogdanov (changes requested):
See my comments above

projects/angular/src/emphasis/alert/utils/alert-types.ts
bbogdanov: [TypeScript playground for reference](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcCCAbYsAqzUC8cA5AJYB2AZhMXAD4kDuAhlORQOa0PEAmz5Dtm4kAzgFcAxpOCjRxANwAoJaEiw4kiOVHxg5cQFt0WXPgCiB4wG8lce-YCSVCHCJkXxADR2HcAOqs7IJuTEGc3r4OACICQlChfHHCPn5wAMpSMnKJEtKy8j4AvipaOvCiwDCY2DB4KKEAFMymdfgAXCa19cAAlG4AfHC2SiVKldWtPY01ZiiWRgB0sYLYvQpAA)

PR #105, opened 2022-05-05
fix: update required Clarity Core version with v13

bbogdanov:
Since after 6.0.0.beta.1 release the old focus trap has been removed, if we add a version that is using some version related to @cds/core@6.0.0 the compiler will throw an error that some files are missing.

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/package.json
+++ b/projects/angular/package.json
@@ -20,6 +20,6 @@
   "peerDependencies": {
     "@angular/common": "^13.0.0",
     "@angular/core": "^13.0.0",
-    "@cds/core": "next || ^6.0.0 || ^5.6.0"
+    "@cds/core": "next || <6.0.0 || ^5.6.0"
   }
 }


ashleyryan:
Hmm, so clr users *can't* use v6 in their code? Do we have a new focus trap from Core we can use in v14+?

Maybe we copy the old code over here instead until then?

bbogdanov:
@ashleyryan  That's why I needed to update the focus trap until we release v14. We can copy the code yes. Let's discuss that on the standup today.

bbogdanov:
https://github.com/vmware-clarity/ng-clarity/blob/8f11c3c8cb60982f9cfadcca7514982d9b616bd6/projects/angular/src/utils/focus-trap/focus-trap.ts#L7 These are the files that are missing in `v6`

PR #134, opened 2022-05-26
Adding utils for key navigation in list/menu 

bbogdanov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

By [w3c specification](https://w3c.github.io/aria-practices/examples/menu-button/menu-button-links.html) role menu has to have keyboard navigation with arrow keys. At the moment the overflow menu in datagrid doesn't support this by default.

Issue Number: VPAT-600

## What is the new behavior?

By using the `clrKeyNavigationList` directive on the container all clickable/focusable elements inside the container will be able to focus with arrow key.

- [x] Yes
- [ ] No

### DIFF
--- a/projects/angular/clarity.api.md
+++ b/projects/angular/clarity.api.md
@@ -48,6 +48,7 @@ import { Subscription } from 'rxjs';
 import { TemplateRef } from '@angular/core';
 import { TrackByFunction } from '@angular/core';
 import { Type } from '@angular/core';
+import { ValueProvider } from '@angular/core';
 import { ViewContainerRef } from '@angular/core';
 
 // @public (undocumented)
@@ -112,6 +113,12 @@ export const CLR_ICON_DIRECTIVES: Type<any>[];
 
 // @public (undocumented)
 export const CLR_LAYOUT_DIRECTIVES: Type<any>[];
+
+// @public (undocumented)
+export const CLR_LIST_KEY_NAVIGATION_COFNIG_PROVIDER: ValueProvider;
+
+// @public (undocumented)
+export const CLR_LIST_KEY_NAVIGATION_CONFIG: InjectionToken<unknown>;
 
 // @public (undocumented)
 export const CLR_LOADING_BUTTON_DIRECTIVES: Type<any>[];
@@ -2790,6 +2797,35 @@ class ClrKeyFocusModule {
 }
 
 // @public (undocumented)
+export class ClrKeyNavigationList implements AfterViewInit {
+    constructor(host: ElementRef<HTMLElement>, defaultConfig: KeyNavigationListConfig, focusableElementSelectors: string[]);
+    // (undocumented)
+    set configuration(config: Partial<KeyNavigationListConfig>);
+    // (undocumented)
+    ngAfterViewInit(): void;
+    // (undocumented)
+    onClick(e: Event): void;
+    // (undocumented)
+    onKeyDown(e: KeyboardEvent): void;
+    // (undocumented)
+    static ɵdir: i0.ɵɵDirectiveDeclaration<ClrKeyNavigationList, "[clrKeyNavigationList]", never, { "configuration": "configuration"; }, {}, never>;
+    // (undocumented)
+    static ɵfac: i0.ɵɵFactoryDeclaration<ClrKeyNavigationList, never>;
+}
+
+// @public (undocumented)
+export class ClrKeyNavigationListModule {
+    // (undocumented)
+    static ɵfac: i0.ɵɵFactoryDeclaration<ClrKeyNavigationListModule, never>;
+    // (undocumented)
+    static ɵinj: i0.ɵɵInjectorDeclaration<ClrKeyNavigationListModule>;
+    // Warning: (ae-forgotten-export) The symbol "i1" needs to be exported by the entry point index.d.ts
+    //
+    // (undocumented)
+    static ɵmod: i0.ɵɵNgModuleDeclaration<ClrKeyNavigationListModule, [typeof i1_50.ClrKeyNavigationList], never, [typeof i1_50.ClrKeyNavigationList]>;
+}
+
+// @public (undocumented)
 export class ClrLabel implements OnInit, OnDestroy {
     constructor(controlIdService: ControlIdService, layoutService: LayoutService, ngControlService: NgControlService, renderer: Renderer2, el: ElementRef);
     // (undocumented)
@@ -5139,6 +5175,12 @@ export const FOCUS_ON_VIEW_INIT_DIRECTIVES: Type<any>[];
 const FOCUS_TRAP_DIRECTIVES: Type<any>[];
 
 // @public (undocumented)
+export const FOCUSABLE_ELEMENT_SELECTORS: InjectionToken<unknown>;
+
+// @public (undocumented)
+export const FOCUSABLE_ELEMENT_SELECTORS_PROVIDER: ValueProvider;
+
+// @public (undocumented)
 const FOCUSABLES = "[href]:not([tabindex=\"-1\"]), button:not([disabled]):not([tabindex=\"-1\"]), input:not([disabled]):not([tabindex=\"-1\"]), select:not([disabled]):not([tabindex=\"-1\"]), textarea:not([disabled]):not([tabindex=\"-1\"]), [contenteditable=\"true\"]:not([tabindex=\"-1\"]), [tabindex]:not([tabindex=\"-1\"]) ";
 
 // @public (undocumented)
@@ -5152,6 +5194,22 @@ export const IS_TOGGLE_PROVIDER: {
 
 // @public (undocumented)
 export function isToggleFactory(): BehaviorSubject<boolean>;
+
+// @public (undocumented)
+export interface KeyNavigationListConfig {
+    // (undocumented)
+    dir: string | null;
+    // (undocumented)
+    keyListItems: string;
+    // (undocumented)
+    layout: 'both' | 'horizontal' | 'vertical';
+    // (undocumented)
+    loop: boolean;
+    // (undocumented)
+    manageFocus: boolean;
+    // (undocumented)
+    manageTabindex: boolean;
+}
 
 // @public
 export abstract class LoadingListener {
--- a/projects/angular/src/utils/list-key-navigation/key-navigation-list.ts
+++ b/projects/angular/src/utils/list-key-navigation/key-navigation-list.ts
@@ -1,0 +1,142 @@
+/*
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { AfterViewInit, Directive, ElementRef, HostListener, Inject, Input } from '@angular/core';
+import {
+  focusElement,
+  getFlattenedFocusableItems,
+  initializeKeyListItems,
+  KeyNavigationCode,
+  setActiveKeyListItem,
+  validKeyNavigationCode,
+} from '@cds/core/internal';
+
+import { FOCUSABLE_ELEMENT_SELECTORS } from './providers/focusable-items';
+import { CLR_LIST_KEY_NAVIGATION_CONFIG } from './providers/key-navigation-list.config';
+import { getNextKeyListItem } from './utils';
+
+export interface KeyNavigationListConfig {
+  keyListItems: string;
+  layout: 'both' | 'horizontal' | 'vertical';
+  manageFocus: boolean;
+  manageTabindex: boolean;
+  loop: boolean;
+  dir: string | null;
+}
+
+@Directive({
+  selector: '[clrKeyNavigationList]',
+})
+export class ClrKeyNavigationList implements AfterViewInit {
+  private config: KeyNavigationListConfig;
+
+  @Input()
+  set configuration(config: Partial<KeyNavigationListConfig>) {
+    this.config = {
+      ...this.defaultConfig,
+      dir: this.getClosestDirAttribute(),
+      ...config,
+    };
+  }
+
+  constructor(
+    private host: ElementRef<HTMLElement>,
+    @Inject(CLR_LIST_KEY_NAVIGATION_CONFIG) private defaultConfig: KeyNavigationListConfig,
+    @Inject(FOCUSABLE_ELEMENT_SELECTORS) private focusableElementSelectors: string[]
+  ) {
+    this.config = {
+      ...this.defaultConfig,
+      dir: this.getClosestDirAttribute(),
+    };
+  }
+
+  ngAfterViewInit() {
+    this.initializeTabIndex();
+  }
+
+  @HostListener('click', ['$event'])
+  onClick(e: Event) {
+    this.clickItem(e);
+  }
+
+  @HostListener('keydown', ['$event'])
+  onKeyDown(e: KeyboardEvent) {
+    this.focusItem(e);
+  }
+
+  private initializeTabIndex() {
+    if (this.config.manageFocus && this.config.manageTabindex) {
+      initializeKeyListItems(this.getListItems());
+    }
+  }
+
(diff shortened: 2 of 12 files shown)


bbogdanov:
Most of the code is an adjustment of https://github.com/vmware-clarity/core/blob/d43375df2c754ac4afc0a09210ff9d1d63be162b/projects/core/src/internal/controllers/key-navigation-list.controller.ts#L38

projects/demo/src/app/cds-theme-select.component.ts
```diff
@@ -35,7 +35,13 @@ export class CdsThemeSelectComponent implements OnInit, OnDestroy {
     this.applyTheme(null);
   }
 
-  applyTheme(theme: string) {
+  applyTheme(event: Event) {
```
kevinbuhmann, line 38: This breaks the `ngOnInit`.

projects/demo/src/app/cds-theme-select.component.ts
```diff
@@ -35,7 +35,13 @@ export class CdsThemeSelectComponent implements OnInit, OnDestroy {
...
+    if (!event || !event.target) {
```
kevinbuhmann, line 39: When would the event or target be falsy?

projects/angular/clarity.api.md
```diff
@@ -2789,6 +2789,31 @@ class ClrKeyFocusModule {
     static ɵmod: i0.ɵɵNgModuleDeclaration<ClrKeyFocusModule, [typeof i1_13.ClrKeyFocus, typeof i2_10.ClrRovingTabindex, typeof i3_10.ClrKeyFocusItem], [typeof i6.CommonModule], [typeof i1_13.ClrKeyFocus, typeof i2_10.ClrRovingTabindex, typeof i3_10.ClrKeyFocusItem]>;
 }
 
+// @public (undocumented)
```
kevinbuhmann, line 2792: These changes should be in the commit that changes the api (the first commit).

kevinbuhmann (changes requested):

projects/demo/src/app/cds-theme-select.component.ts
bbogdanov: Good point. I made the change cause it was causing a runtime error and I did the changes to unblock me on the rest. I'll fix it tomorrow

kevinbuhmann: I would suggest this:

```typescript
onThemeChange(event: Event) {
  if (event?.target) {
    this.applyTheme(event.target.value); // may have to cast `target` to `any` here
  }
}

private applyTheme(theme: string) {
  ...
}

File: projects/demo/src/app/cds-theme-select.component.ts [hunk 4]
```diff
@@ -35,6 +39,12 @@ export class CdsThemeSelectComponent implements OnInit, OnDestroy {
     this.applyTheme(null);
   }
 
+  onThemeChange(event: SelectChangeEvent) {
+    if (event?.target) {
+      this.applyTheme(event.target.value); // may have to cast `target` to `any` here
```
kevinbuhmann, line 44: The comment should be removed.

kevinbuhmann: Also, this commit is a chore, not a fix, because it doesn't affect any library.

kevinbuhmann: Also, I'm still curious why this is needed. How can this issue be reproduced?

bbogdanov: Runtime error. Pull up the `main` branch and run it :)

projects/angular/src/utils/list-key-navigation/key-navigation-list.ts
```diff
@@ -0,0 +1,144 @@
...
+    this.config = {
+      ...this.config,
+      ...config,
+    };
+  }
+
+  ngAfterViewInit() {
+    this.initializeTabIndex();
+    this.renderer.listen(this.host.nativeElement, 'click', this.clickItem.bind(this));
+    this.renderer.listen(this.host.nativeElement, 'keydown', this.focusItem.bind(this));
```
kevinbuhmann, line 80: Why not use the `HostListener` method on the directive?

projects/angular/src/utils/list-key-navigation/key-navigation-list.ts
```diff
@@ -0,0 +1,144 @@
...
+      manageFocus: true,
+      manageTabindex: true,
+      loop: false,
+      dir: this.host.nativeElement.getAttribute('rtl'),
+      ...config,
+    };
+  }
+
+  @Input()
+  set configuration(config: Partial<KeyNavigationListConfig>) {
```
kevinbuhmann, line 70: Properties usually go before the constructor.

projects/angular/src/utils/list-key-navigation/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export interface KeyListConfig {
+  code: KeyNavigationCode;
+  loop?: boolean;
+  layout?: 'horizontal' | 'vertical' | 'both';
+  dir: string | null | undefined;
+}
+
+export function getNextKeyListItem(item: HTMLElement, items: HTMLElement[], config: KeyListConfig) {
+  // todo: cory test
```
kevinbuhmann, line 17: Is this todo still relevant?

kevinbuhmann (changes requested):
I haven't fully understood the key navigation code yet, but here are some thoughts.

projects/angular/src/utils/list-key-navigation/key-navigation-list.ts
```diff
@@ -0,0 +1,144 @@
...
+    this.config = {
+      ...this.config,
```
kevinbuhmann, line 72: I think the default config should be used here so that behavior is more consistent/predictable if this setter is called more than once.

kevinbuhmann:
```typescript
clrKeyNavigationList.configuration = { layout: 'horizontal' };

// Due to the normal nature of assignment, I think a user would assume that this assignment
// does the same thing whether or not there was some previous assignment.
clrKeyNavigationList.configuration = { manageFocus: false };
```

projects/angular/src/utils/list-key-navigation/key-navigation-list.ts
```diff
@@ -0,0 +1,144 @@
...
+  loop: boolean;
+  dir: string | null;
+}
+
+@Directive({
+  selector: '[clrKeyNavigationList]',
+})
+export class ClrKeyNavigationList {
+  private get listItems(): NodeListOf<HTMLElement> {
+    return this.host.nativeElement.querySelectorAll(
```
kevinbuhmann, line 33: I think this should be a method, not a property, due to the computational expense. Programmers are more likely to cache the results of a method, than a property, if caching is needed.

projects/angular/src/utils/list-key-navigation/key-navigation-list.ts
```diff
@@ -0,0 +1,144 @@
...
+          // todo: test for only emiting when position has changed
+          this.setActiveItem(e, this.listItems[next], this.listItems[previous]);
+        }
+      }
+    }
+  }
+
+  private getActiveItem(e: Event) {
+    return Array.from(this.listItems).find(
+      c => c === (e.target as HTMLElement).closest(this.listItems[0].tagName.toLocaleLowerCase()) ?? c === e.target
```
kevinbuhmann, line 115: See my comment about `listItems` being changed to a method. This code is unnecessarily expensive because it runs the DOM query several times. Please look at all uses of the `listItems` property for easy optimizations.

projects/angular/src/utils/list-key-navigation/providers/key-navigation-list.ts
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+import { InjectionToken, ValueProvider } from '@angular/core';
+
+export const ClrListKeyNavigationToken = new InjectionToken('ClrListKeyNavigation');
+
+export const ClrKeyNaviationListProvider = {
+  provide: ClrListKeyNavigationToken,
+  useValue: {},
+} as ValueProvider;
```
kevinbuhmann, line 14: What is this for? It doesn't seem to be injected anywhere.

projects/angular/src/utils/list-key-navigation/providers/focusable-items.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { InjectionToken, ValueProvider } from '@angular/core';
+
+export const FocusableElementsToken = new InjectionToken('FocusableElements');
+
+export const FocusableElements = {
```
kevinbuhmann, line 11: 1. These constants should be renamed to snake case for consistency with other tokens/providers.
2. The provider constant should be named as such for consistency with other providers (end with "provider").
2. I think "elements" should be "element selectors" for code readability. (I was confused by `this.focusableElements` in the directive code b/c I thought it was an array of element rather than an array of selectors.)

So `FOCUSABLE_ELEMENT_SELECTORS_TOKEN` and `FOCUSABLE_ELEMENT_SELECTORS_PROVIDER`.

projects/angular/src/utils/list-key-navigation/providers/focusable-items.ts
```diff
@@ -0,0 +1,27 @@
...
+    'select:not([disabled])',
+    'textarea:not([disabled])',
+    'iframe',
+    'object',
+    'embed',
+    '*[tabindex]', // -1 tabindex is a focusable element and needed for keyboard navigation
+    '*[contenteditable=true]',
+    '[role=button]:not([disabled])',
+  ],
+} as ValueProvider;
```
kevinbuhmann, line 27: This should be a type annotation instead of a type assertion b/c a type assertion will not generate a compile-time error here if a property is missing/misspelled.

```typescript
export const FOCUSABLE_ELEMENT_SELECTORS_PROVIDER: ValueProvider = { ... };
```

projects/angular/src/utils/list-key-navigation/providers/key-navigation-list.config.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { InjectionToken, ValueProvider } from '@angular/core';
+
+export const ClrListKeyNavigationConfigToken = new InjectionToken('ClrListKeyNavigationConfig');
+
+export const ClrListKeyNavigationConfig = {
```
kevinbuhmann, line 11: 1. These constants should be renamed to snake case for consistency with other tokens/providers.
2. The provider constant should be named as such for consistency with other providers (end with "provider").

projects/angular/src/utils/list-key-navigation/providers/key-navigation-list.config.ts
```diff
@@ -0,0 +1,21 @@
...
+  provide: ClrListKeyNavigationConfigToken,
+  useValue: {
+    keyListItems: 'keyListItems',
+    layout: 'vertical',
+    manageFocus: true,
+    manageTabindex: true,
+    loop: false,
+    dir: null,
+  },
+} as ValueProvider;
```
kevinbuhmann, line 21: This should be a type annotation instead of a type assertion b/c a type assertion will not generate a compile-time error here if a property is missing/misspelled.

```typescript
export const CLR_LIST_KEY_NAVIGATION_CONFIG_PROVIDER: ValueProvider = { ... };
```

projects/angular/src/utils/list-key-navigation/key-navigation-list.ts
```diff
@@ -0,0 +1,133 @@
...
+        if (next !== previous) {
+          this.setActiveItem(e, this.getListItems()[next], this.getListItems()[previous]);
+        }
+      }
+    }
+  }
+
+  private getActiveItem(e: Event) {
+    return Array.from(this.getListItems()).find(
+      c => c === (e.target as HTMLElement).closest(this.getListItems()[0].tagName.toLocaleLowerCase()) ?? c === e.target
```
kevinbuhmann, line 100: This code is unnecessarily expensive because it runs the DOM query several times. I recommend adding `const listItems = this.getListItems();` so that `getListItems` is only called once.

kevinbuhmann (changes requested):

projects/angular/src/utils/list-key-navigation/key-navigation-list.ts
```diff
@@ -0,0 +1,133 @@
...
+export class ClrKeyNavigationList implements AfterViewInit, OnDestroy {
+  private config: KeyNavigationListConfig;
+
+  private unlisteners: (() => void)[] = [];
+
+  @Input()
+  set configuration(config: Partial<KeyNavigationListConfig>) {
+    this.config = {
+      ...this.defaultConfig,
+      dir: this.host.nativeElement.getAttribute('rtl'),
```
kevinbuhmann, line 42: I think this should be reading the `dir` attribute. https://www.w3schools.com/tags/att_global_dir.asp

Also: What if the `dir` attribute changes at after this directive is instantiated or the `configuration` input is set? What if the `dir` attribute is set on a parent element?

I think this attribute should be read when the arrow keys are pressed in case it has changed. Also, as the `dir` attribute should be inherited, the code should look up the DOM tree until it finds it or hits the root `html` element.

projects/angular/src/utils/list-key-navigation/providers/key-navigation-list.ts
bbogdanov: Good catch. Thanks.

projects/angular/src/utils/list-key-navigation/key-navigation-list.ts
bbogdanov: Yea I thought about that as well. Will do.

kevinbuhmann (changes requested):

projects/angular/src/utils/list-key-navigation/key-navigation-list.ts
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  ngAfterViewInit() {
+    this.initializeTabIndex();
+    this.unlisteners.push(this.renderer.listen(this.host.nativeElement, 'click', this.clickItem.bind(this)));
+    this.unlisteners.push(this.renderer.listen(this.host.nativeElement, 'keydown', this.focusItem.bind(this)));
+  }
+
+  ngOnDestroy() {
+    this.unlisteners.forEach(unlisten => unlisten());
```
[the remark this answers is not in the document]
bbogdanov: I am aware of the HostListener and what it does. What the problem here?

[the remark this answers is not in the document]
kevinbuhmann: The problem, in my opinion, is the unnecessary code to manage the event listeners. I think we should use the framework for that.

projects/demo/src/app/cds-theme-select.component.ts
kevinbuhmann: I unresolved this because I can produce an error. Could you provide any more detailed reproduction steps? I am mainly curious as to how this error is possible.

bbogdanov: Strange. Getting it again and running it doesn't raise an error. Hard to reproduce. I'll remove the commit. At first I had the same impression as you - Why would that be an issue? But the runtime was raising an issue in the devtools and it was stopping me to continue on the rest.

kevinbuhmann:
Note: This was reverted in #185.

PR #169, opened 2022-06-13
(Backport) feat: keyboard navigation utils for list/menus

bbogdanov:
Backport of #134


bbogdanov:
Forcing a Core v6 installation is not a good idea to use the key navigation in v12. 

https://material.angular.io/cdk/menu/overview alternative to use with v12

Issue #147, opened 2022-06-02
Allow @clr/ui@12 to use @cds/core@6


d3lio:
## Describe the bug

Currently in our project we use Clarity UI 12 with CDS Core 6 as we want to utilise the theming capabilities of Core 6 with the variables shim without upgrading Angular and the Clarity components. The issue is that we need to have two dependencies - @cds/core 5 and 6 because @clr/ui 12 has a peer dependency @cds/core 5 which we don't use.

## Expected behavior

@clr/ui@12 should have a peer dependency @cds/core@>=5 since CDS Core is aiming to have backwards compatibility with older versions of @clr/ui.

## Versions

**Clarity version:**

- [x] v12.x
- [x] v13.x
- [ ] Other:

**Framework:**

- [x] Angular
- [ ] React
- [ ] Vue
- [ ] Other:

**Framework version:**
Angular 12

**Device:**
Any

ashleyryan:
@bbogdanov your PR is not what this issue is about. The default size of icons (when not including the core css) changed in v6 compared to v5

kevinbuhmann:
It looks like the icon issue is 138, not this issue.

ashleyryan:
oops, thanks. My bad

d3lio:
Hey guys can we get an update on this?

kevinbuhmann:
My vote is that we don't do this.

1. Angular 12 is no longer under active support by the Angular team at Google, and backporting changes sometimes fails due to bugs in the legacy view engine compiler.
2. There is a maintenance cost of ensuring that @clr/angular works with @cds/core@5 and @cds/core@6. We have already been burned at least on one this and had to revert a new feature.

bbogdanov:
@kevinbuhmann Even that Angular team doesn't actively support, the company does. Another way to look into this is to figure out a way to port the theming capabilities. 

The feature you reverted used a code that was part of `v6` but not `v5` which we know caused issues. Since that feature is not in the code the only thing left used between the two packages are the icons that haven't changed in between `v5` and `v6`. 

So what's not trivial about it?

kevinbuhmann:
@bbogdanov, it's non-trivial to ensure that we don't accidentally break something for @cds/core@5 users because we develop against @cds/core@6. After I posted my earlier comment, I found a way to build PRs against @cds/core@5 in CI, but I haven't discussed that with the team yet. It should help prevent us from getting burned again, though.

[...]

PR #170, opened 2022-06-14
(backport) feat: keyboard navigation utils for list/menus

bbogdanov:
Backport of #134


package.json
```diff
@@ -57,7 +57,7 @@
     "@angular/platform-browser": "~12.0.0",
     "@angular/platform-browser-dynamic": "~12.0.0",
     "@angular/router": "~12.0.0",
-    "@cds/core": "^5.0.0",
+    "@cds/core": ">=5.0.0 || <7.0.0",
```
bbogdanov, line 60: #147 closed in favour of this

projects/angular/package.json
```diff
@@ -20,6 +20,6 @@
   "peerDependencies": {
     "@angular/common": "^12.0.0",
     "@angular/core": "^12.0.0",
-    "@cds/core": "^5.0.0"
+    "@cds/core": ">=5.0.0 || <7.0.0"
```
bbogdanov, line 23: #147 closed in favour of this

package.json
steve-haar: Wouldn't this match any version of @cds/core? Any version is going to be either greater than 5 *or* less than 7. Am I missing something?

steve-haar: Also, this is only the devDependencies. Did you mean to update the peerDependencies in /projects/angular/package.json ?

steve-haar (changes requested):

package.json
ashleyryan: >Wouldn't this match any version of @cds/core? Any version is going to be either greater than 5 or less than 7. Am I missing something?

I think I left a comment in the original PR about this - currently, yes, it would match any version, but I wasn't sure if we wanted to exclude a hypothetical version 7 in the future since we won't know for sure. It's probably overkill though

steve-haar: If that's the case shouldn't it be `&&` instead of `||`?

steve-haar: Just read the [node-semver](https://github.com/npm/node-semver) spec and looks like it should be `>=5 <7`

projects/angular/src/utils/list-key-navigation/key-navigation-list.ts
```diff
@@ -0,0 +1,142 @@
...
+
+import { AfterViewInit, Directive, ElementRef, HostListener, Inject, Input } from '@angular/core';
+import {
+  focusElement,
+  getFlattenedFocusableItems,
+  initializeKeyListItems,
+  KeyNavigationCode,
+  setActiveKeyListItem,
+  validKeyNavigationCode,
+} from '@cds/core/internal';
```
kevinbuhmann, line 15: These imports cause problems for users even if they don't use the new directive. See https://vmware-clarity.slack.com/archives/C0JF8D2LB/p1655238275941749.

kevinbuhmann (changes requested):

package.json
```diff
@@ -57,7 +57,7 @@
     "@angular/platform-browser": "~12.0.0",
     "@angular/platform-browser-dynamic": "~12.0.0",
     "@angular/router": "~12.0.0",
-    "@cds/core": "^5.0.0",
+    "@cds/core": "5 - 6",
```
kevinbuhmann, line 60: I don't think a range is a good way to declare a dev dependency.

kevinbuhmann, line 60: I don't think a range is a good way to declare a dev dependency.

kevinbuhmann (changes requested):

projects/angular/package.json
```diff
@@ -20,6 +20,6 @@
   "peerDependencies": {
     "@angular/common": "^12.0.0",
     "@angular/core": "^12.0.0",
-    "@cds/core": "^5.0.0"
+    "@cds/core": "5 - 6"
```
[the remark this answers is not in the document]
steve-haar: no, this is equivalent to `>=5.0.0 <6.0.0` (https://github.com/npm/node-semver#advanced-range-syntax)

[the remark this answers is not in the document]
kevinbuhmann: I don't think that's what we want. Shouldn't it be `>=5.0.0 <7.0.0` to allow any 5.x or 6.x version? This may be a moot point if we end up copying the functions.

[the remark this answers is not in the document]
steve-haar: Sorry, I wrote it wrong. `5 - 6` is equivalent to `>=5.0.0 <7.0.0`

package.json
kevinbuhmann: I think we should use `^6.0.0` here as it's more clear and functionally equivalent.

steve-haar:
closing in favor of https://github.com/vmware-clarity/ng-clarity/pull/173

PR #173, opened 2022-06-14
Revert "Adding utils for key navigation in list/menu"

kevinbuhmann:
Reverts vmware-clarity/ng-clarity#134

These "@cds/core/internal" imports cause problems for users even if they don't use the new directive. See https://vmware-clarity.slack.com/archives/C0JF8D2LB/p1655238275941749.


steve-haar (approved):

bbogdanov (dismissed):
Would it be better to port the functions from v6 in clr-angular instead of reverting a feature that fixes a critical a11y issue?

Jinnie (approved):

Jinnie:
@bbogdanov it's a possibility. No problem to go in this direction after reverting the current solution. Unless we can provide the update really quick.

kevinbuhmann:
> Would it be better to port the functions from v6 in clr-angular instead of reverting a feature that fixes a critical a11y issue?

That's fine with me. I'll close this in favor of a PR that copies the internal functions.

kevinbuhmann:
This branch is protected by the wildcard rule.

PR #176, opened 2022-06-15
chore: remove direct dependency on `netlify`

kevinbuhmann:
We can execute `npx netlify-cli` without having `netlify` installed locally. This also means we don't have to checkout code or install dependencies in the demo workflow which will make deploying demo previews faster.

## PR Checklist

## PR Type

CI related changes


steve-haar (approved):

.github/workflows/demo.yml
```diff
@@ -11,14 +11,10 @@ jobs:
     timeout-minutes: 10
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v2
```
kevinbuhmann, line 15: Removing this broke the "Setup Node.js" step. I fixed this in #177.

PR #188, opened 2022-06-23
chore: add chromatic support

steve-haar:
add chromatic support

### DIFF
--- a/.github/workflows/pr-visual-regression.yml
+++ b/.github/workflows/pr-visual-regression.yml
@@ -1,0 +1,33 @@
+name: PR Visual Regression
+on:
+  workflow_run:
+    workflows: [PR Build]
+    types:
+      - completed
+jobs:
+  pr-visual-regression:
+    name: PR Visual Regression
+    if: ${{github.event.workflow_run.conclusion == 'success'}}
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-visual-regression.yml
```diff
@@ -0,0 +1,33 @@
+name: PR Visual Regression
+on:
+  workflow_run:
+    workflows: [PR Build]
+    types:
+      - completed
+jobs:
+  pr-preview:
+    name: PR Chromatic
```
kevinbuhmann, line 9: These should match. I made the others match the file name.

kevinbuhmann (changes requested):

.github/workflows/pr-visual-regression.yml
steve-haar: fixed

kevinbuhmann: The `pr-preview` key is still the same.

steve-haar: sorry, should be fixed now

kevinbuhmann (approved):

PR #195, opened 2022-06-27
add forms and labels stories

steve-haar:
add forms and labels stories

### DIFF
--- a/.storybook/stories/label/label.stories.ts
+++ b/.storybook/stories/label/label.stories.ts
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2016-2022 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { Parameters } from '@storybook/addons';
+import { Story } from '@storybook/angular';
+
+import { setupStorybook } from '../../helpers/setup-storybook.helpers';
+
+enum LabelType {
+  Default = '',
+  Purple = 'label-purple',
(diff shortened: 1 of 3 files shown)


.storybook/stories/label/label.stories.ts
```diff
@@ -0,0 +1,57 @@
...
+    <span class="label" [class.clickable]="clickable" [ngClass]="labelType">
+      {{content}}
+      <span *ngIf="badgeContent" class="badge">{{badgeContent}}</span>
+    </span>
+  `,
+  props: { ...args },
+});
+
+const defaultParameters: Parameters = {
+  title: 'Label/Label',
```
kevinbuhmann, line 35: Since labels are closely related to forms, should this be `Forms/Label`?

.storybook/stories/label/label.stories.ts
```diff
@@ -0,0 +1,57 @@
...
+  Default = '',
+  Purple = 'label-purple',
+  Blue = 'label-blue',
+  Orange = 'label-orange',
+  LightBlue = 'label-light-blue',
+  Info = 'label-info',
+  Success = 'label-success',
+  Warning = 'label-warning',
+  Error = 'label-danger',
+}
```
kevinbuhmann, line 22: This could be a simple `string[]` since the enum isn't really used.

.storybook/stories/label/label.stories.ts
steve-haar: They're clickable buttons that are styled differently. Not related to forms

PR #196, opened 2022-06-27
feat(datagrid): change focused action on arrow key press

kevinbuhmann:
fixes VPAT-600

## PR Type

Feature

VPAT-600

The arrow keys don't change the focused datagrid row action.

## What is the new behavior?

The arrow keys change the focused datagrid row action.

No.


projects/angular/src/data/datagrid/datagrid-action-overflow.ts
```diff
@@ -87,18 +100,40 @@ export class ClrDatagridActionOverflow implements OnDestroy {
...
-  private focusFirstButton(): void {
+  private initalizeFocus(): void {
     if (isPlatformBrowser(this.platformId)) {
       this.zone.runOutsideAngular(() => {
         setTimeout(() => {
-          const firstButton: HTMLButtonElement | null = document.querySelector('button.action-item');
-          if (firstButton) {
-            firstButton.focus();
-          }
+          const buttons = Array.from(document.querySelectorAll<HTMLButtonElement>('button.action-item'));
```
Jinnie, line 118: We're projecting with `ng-content`, so it can be anything. Even if we assume it would be only actionable elements, it may be also anchor elements, not only buttons

Jinnie (commented):
My main concern here is that we're introducing yet another library for a utility that we already have inside `cir-angular`
https://github.com/vmware-clarity/ng-clarity/tree/main/projects/angular/src/utils/focus/key-focus

kevinbuhmann:
> My main concern here is that we're introducing yet another library for a utility that we already have inside `cir-angular` https://github.com/vmware-clarity/ng-clarity/tree/main/projects/angular/src/utils/focus/key-focus

I will look into using this instead. I didn't know it was there.

kevinbuhmann:
> My main concern here is that we're introducing yet another library for a utility that we already have inside `cir-angular` https://github.com/vmware-clarity/ng-clarity/tree/main/projects/angular/src/utils/focus/key-focus

Thanks! I have updated this PR to use that directive.

projects/angular/src/data/datagrid/datagrid-action-overflow.ts
kevinbuhmann: I agree that this should probably be improved, but I think it's out of scope here.

Jinnie (approved):
Thanks! Looks great

PR #242, opened 2022-07-11
feat(datepicker): indicate selected day in aria label

d-kostov-dev:
fixes VPAT-3637

PR Type
Feature - Accessibility

What is the current behavior?
The currently selected date is not indicated to screen reader users. 
 
What is the new behavior?
The aria-label of the day is now showing the word "Selected".

No.

### DIFF
--- a/projects/angular/src/forms/datepicker/day.spec.ts
+++ b/projects/angular/src/forms/datepicker/day.spec.ts
@@ -142,6 +142,16 @@ export default function () {
         expect(dayBtn.attributes['aria-label'].value).toEqual(dvm.dayModel.toDateString());
       });
 
+      it('sets the correct value for the aria-label attribute when the date is selected', () => {
+        const dayBtn: HTMLButtonElement = context.clarityElement.children[0];
+        const dvm: DayViewModel = context.clarityDirective.dayView;
+
+        context.testComponent.dayView.isSelected = true;
+
+        context.detectChanges();
+        expect(dayBtn.attributes['aria-label'].value).toEqual(`${dvm.dayModel.toDateString()} - Selected`);
+      });
+
       it('sets aria-selected when the date is selected', () => {
         const button: HTMLButtonElement = context.clarityElement.children[0];
         expect(button.attributes['aria-selected'].value).toBe('false');
--- a/projects/angular/src/forms/datepicker/day.ts
+++ b/projects/angular/src/forms/datepicker/day.ts
@@ -6,6 +6,7 @@
 
 import { Component, Input } from '@angular/core';
 
+import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
 import { ClrPopoverToggleService } from '../../utils/popover/providers/popover-toggle.service';
 import { DayViewModel } from './model/day-view.model';
 import { DayModel } from './model/day.model';
@@ -36,12 +37,12 @@ import { DateNavigationService } from './providers/date-navigation.service';
 })
 export class ClrDay {
   private _dayView: DayViewModel;
-  public dayString: string;
 
   constructor(
     private _dateNavigationService: DateNavigationService,
     private _toggleService: ClrPopoverToggleService,
-    private dateFormControlService: DateFormControlService
+    private dateFormControlService: DateFormControlService,
+    private commonStrings: ClrCommonStringsService
   ) {}
 
   /**
@@ -51,11 +52,18 @@ export class ClrDay {
   @Input('clrDayView')
   public set dayView(day: DayViewModel) {
     this._dayView = day;
-    this.dayString = this._dayView.dayModel.toDateString();
   }
 
   public get dayView(): DayViewModel {
     return this._dayView;
+  }
+
+  public get dayString(): string {
+    return this.dayView.isSelected
+      ? this.commonStrings.parse(this.commonStrings.keys.datepickerSelectedLabel, {
+          FULL_DATE: this._dayView.dayModel.toDateString(),
+        })
+      : this._dayView.dayModel.toDateString();
   }
 
   /**
--- a/projects/angular/src/utils/i18n/common-strings.default.ts
+++ b/projects/angular/src/utils/i18n/common-strings.default.ts
@@ -68,6 +68,7 @@ export const commonStringsDefault: ClrCommonStrings = {
   datepickerCurrentDecade: 'Current decade',
   datepickerSelectMonthText: 'Select month, the current month is {CALENDAR_MONTH}',
   datepickerSelectYearText: 'Select year, the current year is {CALENDAR_YEAR}',
+  datepickerSelectedLabel: '{FULL_DATE} - Selected',
   // Stack View
   stackViewChanged: 'Value changed.',
   // Responsive Nav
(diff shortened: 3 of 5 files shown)


projects/angular/src/forms/datepicker/day.ts
```diff
@@ -50,7 +51,7 @@ export class ClrDay {
   @Input('clrDayView')
   public set dayView(day: DayViewModel) {
     this._dayView = day;
-    this.dayString = this._dayView.dayModel.toDateString();
+    this.dayString = `${this._dayView.dayModel.toDateString()} ${this.dayView.isSelected ? this.commonStrings.keys.datepickerSelectedLabel : ''}`;
```
ashleyryan, line 54: You can't just concatenate like this, since not every language uses this structure. You'll need to put a placeholder in the string.

I need to run this by our A11Y team before we can accept it

projects/angular/src/utils/i18n/common-strings.default.ts
```diff
@@ -68,6 +68,7 @@ export const commonStringsDefault: ClrCommonStrings = {
   datepickerCurrentDecade: 'Current decade',
   datepickerSelectMonthText: 'Select month, the current month is {CALENDAR_MONTH}',
   datepickerSelectYearText: 'Select year, the current year is {CALENDAR_YEAR}',
+  datepickerSelectedLabel: 'Selected',
```
kevinbuhmann, line 71: I think this should be consistent with `datepickerSelectMonthText` and `datepickerSelectMonthText`.

```typescript
datepickerSelectDayText: 'Select day, the current day is {CALENDER_DAY}`
```

kevinbuhmann (changes requested):

projects/angular/src/forms/datepicker/day.ts
steve-haar: Technically, this will add a trailing space to the end of the day if it's not selected. I doubt it will affect screen readers, but if we wanted to be more correct we could do something like:

```
const selectedText = this.dayView.isSelected ? ` ${this.commonStrings.keys.datepickerSelectedLabel}` : '';
this.dayString = `${this._dayView.dayModel.toDateString()}${selectedText}`
```

kevinbuhmann: We cannot use concatenation though, like Ashley said. We need a placeholder in the common string.

kevinbuhmann:
I force pushed this branch to get the PR building for a11y review and to reword the commit message to follow our conventions.

ashleyryan:
@d-kostov-dev Here is the recommendation from our A11Y expert:

> Because aria-selected is not widely supported in either td or button for screen readers we recommend using aria-label="[full date] - Selected" instead. Additionally aria-current="date" should also be provided to indicate the current date as Clarity's date picker does show in visually in bold what the current date is when it is not selected.

They included a picture too: 
image

kevinbuhmann (changes requested):
See comment. Also, please fix the build errors (including lint and test errors).

projects/angular/src/forms/datepicker/day.ts
```diff
@@ -27,6 +27,7 @@ import { DateNavigationService } from './providers/date-navigation.service';
       (click)="selectDay()"
       (focus)="onDayViewFocus()"
       [attr.aria-label]="dayString"
+      [attr.aria-current]="dayView.isTodaysDate"
```
kevinbuhmann, line 30: This isn't quite right because the value should be literally "date" as shown in the graphic. But this is actually unrelated to I indicating the selected day, so I think this should removed. I'll open a separate PR.

kevinbuhmann (changes requested):
See comment.

Also, please squash the commits into one "feat(datepicker): indicate selected day in aria-label" commit.

kevinbuhmann (changes requested):
Please also run `npm run build && npm run public-api:update` to update the public API file.

projects/angular/src/forms/datepicker/day.ts
kevinbuhmann: This is in #253.

kevinbuhmann (changes requested):
Also, please add a test for this change. You can use this if you want (after the "sets the correct value for the aria-label attribute" test):

```typescript
it('sets the correct value for the aria-label attribute when the date is selected', () => {
  const dayBtn: HTMLButtonElement = context.clarityElement.children[0];
  const dvm: DayViewModel = context.clarityDirective.dayView;

  context.testComponent.dayView.isSelected = true;

  context.detectChanges();
  expect(dayBtn.attributes['aria-label'].value).toEqual(`${dvm.dayModel.toDateString()} - SELECTED`);
});
```

projects/angular/src/forms/datepicker/day.ts
```diff
@@ -50,13 +52,20 @@ export class ClrDay {
...
     this._dayView = day;
-    this.dayString = this._dayView.dayModel.toDateString();
   }
 
   public get dayView(): DayViewModel {
     return this._dayView;
   }
 
+  public get dayString(): string {
+    return this.dayView.isSelected
```
kevinbuhmann, line 62: Why was this moved?

d-kostov-dev: The dayString is initialized with the input. And it works ok in the app because every time a date is selected the grid closes and when you open it, it initializes again, but the change detection on the unit test is not working this way and the test is not passing, that's why I moved it to getter. Maybe should be private or there is an other solution to this.

kevinbuhmann: Ah, okay.

kevinbuhmann (changes requested):
Please squash the commits and keep the first commit message. These commits are not compatible with our release process.

Also, please rebase and run `npm run build && npm run public-api:update` to fix the build error.

projects/angular/src/forms/datepicker/day.ts
```diff
@@ -25,7 +26,9 @@ import { DateNavigationService } from './providers/date-navigation.service';
       [attr.tabindex]="dayView.tabIndex"
       (click)="selectDay()"
       (focus)="onDayViewFocus()"
+      [attr.aria-current]="dayView.isTodaysDate ? 'date' : 'false'"
```
kevinbuhmann, line 29: This should not be here. Please fetch the upstream repo and rebase.

kevinbuhmann (changes requested):

kevinbuhmann:
Sorry if you are still working on the rebase, but I see the merge commit you just pushed. We don't want a merge commit. There should be just one commit without any of the `aria-current` stuff. I have already merged via a different PR.

projects/angular/src/forms/datepicker/day.spec.ts
```diff
@@ -122,12 +122,36 @@ export default function () {
         expect(context.clarityDirective.selectDay).toHaveBeenCalled();
       });
 
+      it('sets arria-current on the button correctly', function () {
```
kevinbuhmann, line 125: This should be removed.

kevinbuhmann:
I went ahead and rebased for you to remove the `aria-current` changes from this commit. When the build passes, I will merge.

ashleyryan (approved):

kevinbuhmann (approved):

PR #255, opened 2022-07-25
feat(datagrid): add input for page size select element id

pradoslavVM:
PR Type
Fix - Accessibility
What is the current behavior?
Currently you cant set an Id for the select
displaying the page size.
What is the new behavior?
Will be able to set an Id to the select
and associate it with a label using the
for attribute.
No.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/clarity.api.md
+++ b/projects/angular/clarity.api.md
@@ -942,6 +942,8 @@ export interface ClrCommonStrings {
     // (undocumented)
     datepickerPreviousMonth: string;
     // (undocumented)
+    datepickerSelectedLabel: string;
+    // (undocumented)
     datepickerSelectMonthText: string;
     // (undocumented)
     datepickerSelectYearText: string;
@@ -2061,9 +2063,9 @@ export class ClrDatepickerViewManager {
 
 // @public (undocumented)
 export class ClrDay {
-    constructor(_dateNavigationService: DateNavigationService, _toggleService: ClrPopoverToggleService, dateFormControlService: DateFormControlService);
+    constructor(_dateNavigationService: DateNavigationService, _toggleService: ClrPopoverToggleService, dateFormControlService: DateFormControlService, commonStrings: ClrCommonStringsService);
     // (undocumented)
-    dayString: string;
+    get dayString(): string;
     // Warning: (ae-forgotten-export) The symbol "DayViewModel" needs to be exported by the entry point index.d.ts
     set dayView(day: DayViewModel);
     // (undocumented)
--- a/projects/angular/src/forms/datepicker/day.spec.ts
+++ b/projects/angular/src/forms/datepicker/day.spec.ts
@@ -142,6 +142,16 @@ export default function () {
         expect(dayBtn.attributes['aria-label'].value).toEqual(dvm.dayModel.toDateString());
       });
 
+      it('sets the correct value for the aria-label attribute when the date is selected', () => {
+        const dayBtn: HTMLButtonElement = context.clarityElement.children[0];
+        const dvm: DayViewModel = context.clarityDirective.dayView;
+
+        context.testComponent.dayView.isSelected = true;
+
+        context.detectChanges();
+        expect(dayBtn.attributes['aria-label'].value).toEqual(`${dvm.dayModel.toDateString()} - Selected`);
+      });
+
       it('sets aria-selected when the date is selected', () => {
         const button: HTMLButtonElement = context.clarityElement.children[0];
         expect(button.attributes['aria-selected'].value).toBe('false');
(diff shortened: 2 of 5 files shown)


projects/angular/src/data/datagrid/datagrid-page-size.ts
```diff
@@ -13,14 +13,15 @@ import { Page } from './providers/page';
   template: `
     <ng-content></ng-content>
     <div class="clr-select-wrapper">
-      <select [class.clr-page-size-select]="true" [(ngModel)]="page.size">
+      <select id="pageSizeOptionsId" [class.clr-page-size-select]="true" [(ngModel)]="page.size">
```
kevinbuhmann, line 16: I am pretty sure you mean `[id]="pageSizeOptionsId"`. Otherwise, this will set the `id` to literally "pageSizeOptionsId" which isn't helpful.

kevinbuhmann (changes requested):
See comment. Also, please reword the commit message to follow convention (commit type is `feat` and the scope is `datagrid` for this) and be more descriptive (external users will have no way of knowing what VPAT-11934 is). I suggest "feat(datagrid): add input for page size `select` element id" or similar.

projects/angular/src/data/datagrid/datagrid-page-size.ts
pradoslavVM: Added the binding,

kevinbuhmann (changes requested):
Please squash the commits and reword the commit message as I described above. These commits are not compatible with our release process.

kevinbuhmann (changes requested):
Please put the commit message subject on the sane line as the commit type and scope so that our release process will generate the changelog properly.

projects/angular/src/data/datagrid/datagrid-page-size.ts
```diff
@@ -13,14 +13,15 @@ import { Page } from './providers/page';
...
-      <select [class.clr-page-size-select]="true" [(ngModel)]="page.size">
+      <select [id]="pageSizeOptionsId" [class.clr-page-size-select]="true" [(ngModel)]="page.size">
         <option *ngFor="let option of pageSizeOptions" [ngValue]="option">{{ option }}</option>
       </select>
     </div>
   `,
 })
 export class ClrDatagridPageSize {
   @Input('clrPageSizeOptions') pageSizeOptions: number[];
+  @Input('clrPageSizeOptionsId') pageSizeOptionsId = '';
```
kevinbuhmann, line 24: I just realized that this will cause the `id` attribute to default to `id=""` which isn't legal.

To fix this, you can inject `@Inject(UNIQUE_ID) publicuniquePageSizeOptionsId: string` and set `this. pageSizeOptionsId = uniquePageSizeOptionsId` in the constructor. This will ensure there is valid and unique default value and still allow a custom value to be set.

kevinbuhmann (changes requested):

projects/angular/src/data/datagrid/datagrid-page-size.ts
kevinbuhmann: Actually, you don't need to do the `@Inject(UNIQUE_ID)`. You can just initalize this property to `uniqueIdFactory()`.

kevinbuhmann (approved):

kevinbuhmann (changes requested):
Again, please squash the commits since this is a single change.

kevinbuhmann:
I went ahead and squashed the comment and added tests. When the build passes, I will merge.

projects/angular/src/data/datagrid/datagrid-page-size.spec.ts
```diff
@@ -56,6 +62,15 @@ export default function (): void {
           expect(context.clarityElement.textContent.trim()).toMatch('Hello world');
         });
 
+        it('displays a select with pageSizeOptionsId as id', function () {
+          const pageSizeOptionsId = 'some-id';
```
ashleyryan, line 66: this uses an autogenerated id by default, right? should we add a test for that?

ashleyryan (approved):

projects/angular/src/data/datagrid/datagrid-page-size.spec.ts
kevinbuhmann: I added a test for the default id.

ashleyryan (approved):

kevinbuhmann:
@pradoslavVM: I accidentally ran the wrong `git` command and updated your fork's `main` branch with `vm-clarity/ng-clarity@main`. Sorry about that. In the future, please don't open PRs with your fork's `main` branch as it causes much confusion. Please use a topic branch. I open a new PR to supersede this one.

PR #270, opened 2022-08-02
feat(datagrid): add input for page size select element id

kevinbuhmann:
This originated in #255 as an external contribution by @pradoslavVM. I accidentally closed that PR by running a bad `git` command.

PR Type
Fix - Accessibility
What is the current behavior?
Currently you cant set an Id for the select
displaying the page size.
What is the new behavior?
Will be able to set an Id to the select
and associate it with a label using the
for attribute.
No.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/data/datagrid/datagrid-page-size.spec.ts
+++ b/projects/angular/src/data/datagrid/datagrid-page-size.spec.ts
@@ -25,6 +25,12 @@ export default function (): void {
         context.detectChanges();
         expect(context.clarityDirective.pageSizeOptions).toEqual([10, 20, 50, 100]);
       });
+
+      it('receives an input for page size options id', function () {
+        context.testComponent.pageSizeOptionsId = 'some-id';
+        context.detectChanges();
+        expect(context.clarityDirective.pageSizeOptionsId).toEqual('some-id');
+      });
     });
 
     describe('View', function () {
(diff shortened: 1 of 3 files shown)


projects/angular/src/data/datagrid/datagrid-page-size.spec.ts
```diff
@@ -25,6 +25,12 @@ export default function (): void {
         context.detectChanges();
         expect(context.clarityDirective.pageSizeOptions).toEqual([10, 20, 50, 100]);
       });
+
+      it('receives an input for page size options id', function () {
```
ashleyryan, line 29: do we need this test here if we have it below?

ashleyryan (approved):

projects/angular/src/data/datagrid/datagrid-page-size.spec.ts
kevinbuhmann: Yeah. The pattern seems to be 1) a test that the input binding itself works and 2) a test for the behavior that sets the property directly. That way, we are effectively testing that it works either way the user sets the property.

Jinnie (approved):

PR #279, opened 2022-08-16
chore: register providers that use `Renderer2` at component in tests

kevinbuhmann:
This prevents the `this.renderer.listen is not a function` error that was being thrown in `afterAll`.

I think the fix is a combination of using a real `Renderer2` and ensuring that `ClrPopoverEventsService` is destroyed properly.

### DIFF
--- a/projects/angular/src/data/datagrid/all.spec.ts
+++ b/projects/angular/src/data/datagrid/all.spec.ts
@@ -72,6 +72,7 @@ describe('Datagrid', function () {
     TableSizeServiceSpec();
     ColumnResizerServiceSpecs();
   });
+
   describe('Components', function () {
     DatagridActionBarSpecs();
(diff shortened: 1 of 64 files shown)


Jinnie (approved):
Great! I even see small performance improvement (may be in the standard deviation range).

before:
Finished in 45.773 secs / 37.37 secs
after:
Finished in 42.566 secs / 34.804 secs

ashleyryan (approved):

kevinbuhmann:
I cleaned up some changes that weren't needed. The providers registered by `TestBed.overrideDirective` don't need to be moved to the `Component` decorator.

https://github.com/vmware-clarity/ng-clarity/compare/3b7366fe9e31363044aaf02b6557fd18efcb605c..27f198b9e6ccd9e98d35945fee930451ab0514b9

PR #313, opened 2022-09-14
fix(datagrid): pagination input should only change page on enter

Jinnie:
Signed-off-by: Ivan Donchev <idonchev@vmware.com>

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Pagination input changes current page on Tab, click-out, etc. blur events.

Issue Number: VPAT-784

## What is the new behavior?

Pagination input will only change the current page on Enter.
Blur resets the input value to the actual current page.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/data/datagrid/datagrid-pagination.spec.ts
+++ b/projects/angular/src/data/datagrid/datagrid-pagination.spec.ts
@@ -256,7 +256,7 @@ export default function (): void {
         expect(context.testComponent.current.toString()).toBe('4');
       });
 
-      it('changes the current page on blur', function () {
+      it('does not change the current page on blur', function () {
         context.testComponent.size = 10;
         context.testComponent.total = 100;
         context.testComponent.current = 1;
@@ -269,7 +269,23 @@ export default function (): void {
         current.dispatchEvent(new Event('blur'));
         // Note: the toString() wouldn't be necessary if we used input type='number',
         // but we decided to opt for type='text' for now due to limited cross-browser support
-        expect(context.testComponent.current.toString()).toBe('4');
+        expect(context.testComponent.current.toString()).toBe('1');
+      });
+
+      it('input value resets on blur', function () {
+        context.testComponent.size = 10;
+        context.testComponent.total = 100;
+        context.testComponent.current = 1;
+        context.detectChanges();
+
+        const current = context.clarityElement.querySelector('.pagination-current');
+        expect(current).not.toBeNull();
+        current.value = 4;
+        current.dispatchEvent(new Event('input'));
+        current.dispatchEvent(new Event('blur'));
+        // Note: the toString() wouldn't be necessary if we used input type='number',
+        // but we decided to opt for type='text' for now due to limited cross-browser support
+        expect(current.value.toString()).toBe('1');
       });
 
       it('ignores the current page when input value is invalid', function () {
@@ -298,7 +314,12 @@ export default function (): void {
         expect(current).not.toBeNull();
         current.value = 0;
         current.dispatchEvent(new Event('input'));
-        current.dispatchEvent(new Event('blur'));
+        current.dispatchEvent(
+          new KeyboardEvent('keydown', {
+            code: 'Enter',
+            key: 'Enter',
+          })
+        );
         // Note: the toString() wouldn't be necessary if we used input type='number',
         // but we decided to opt for type='text' for now due to limited cross-browser support
         expect(context.testComponent.current.toString()).toBe('1');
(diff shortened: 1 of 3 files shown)


projects/angular/src/data/datagrid/datagrid-pagination.spec.ts
```diff
@@ -256,7 +256,7 @@ export default function (): void {
         expect(context.testComponent.current.toString()).toBe('4');
       });
 
-      it('changes the current page on blur', function () {
+      it('does not change the current page on blur', function () {
```
ashleyryan, line 259: This seems like it would be considered a breaking change

Jinnie: It's a functionality logged as a bug. If we accept the previous one was erroneous, then it's a bug fix.
It's not API of any kind that users can build functionality upon.
Same way, if a color does not meet contrast, we change it. And though it "could" be a visual breaking change, we consider it a bug fix.

ashleyryan: We have a unit test for very specific behavior that was changed to do the complete opposite. That feels to me like a breaking behavioral change, not a visual one. 

If there wasn't a unit test for it, I wouldn't have thought twice about it, but why do we have a unit test that tests specifically for the bug behavior?

kevinbuhmann: I agree with Ashley.

projects/angular/src/data/datagrid/datagrid-pagination.ts
```diff
@@ -261,6 +261,13 @@ export class ClrDatagridPagination implements OnDestroy, OnInit {
...
 
+  public verifyCurrentPage(event: any): void {
+    const parsed = parseInt(event.target.value, 10);
+    if (parsed !== this.page.current) {
+      event.target.value = this.page.current;
+    }
+  }
+
   /**
    * We only update the pagination's current page on blur of the input field, or
```
kevinbuhmann, line 272: This comment would need to be updated.

projects/angular/src/data/datagrid/datagrid-pagination.spec.ts
kevinbuhmann: Also, I don't quite understand the bug. What are the steps to reproduce?

kevinbuhmann: Actually, sometimes the tests are just plain wrong. See my fix in #317. Although this one changes tested behavior and mine just fixes an incorrect expected value.

Jinnie: Steps to reproduce:
1. Open a datagrid with pagination in it.
2. Tab to the current page input in the footer (or click on it). 
3. Type a new page number (do not hit Enter).
4. Tab out of the input field.
5. Observe the current datagrid page changes.

This is the bug. According to A11Y, an action can not occur without active confirmation. The best solution would be to have a button next to the input, but hitting Enter can be accepted as good enough alternative, as a button will visually overload the component, which already has enough buttons.
Change of the page, happening on focus out of the input should not occur though.

CC: @AmyLiNow 

I don't quite understand the comment about the wrong tests and the provided example.
The tests here used to verify the old behavior. They were correct once this was the expected behavior.
Once it's found that this behavior needs to be changed, we also need to change the tests.

Jinnie: > ...but why do we have a unit test that tests specifically for the behavior of the bug?
Code coverage?

Jinnie: I don't feel convinced that this specific change needs major release.
Though it's a good food for our old "what is a breaking change" talk.

projects/angular/src/data/datagrid/datagrid-pagination.ts
Jinnie: Done

projects/angular/src/data/datagrid/datagrid-pagination.spec.ts
ashleyryan: > The tests here used to verify the old behavior. They were correct once this was the expected behavior.
Once it's found that this behavior needs to be changed, we also need to change the tests.

Sure, the behavior needs to change because of A11Y, but that's still a _breaking behavioral change_. That was the explicit and intended behavior before - we even had unit tests for it. Now we're changing it to do the complete opposite - that's breaking.

If we didn't have the unit tests before, then I could believe the old behavior was just a bug and could be changed in a minor/patch release. But the fact that we did have unit tests for it means that the old behavior was intentional (even if ultimately wrong) and therefore cannot be changed until the next major version.

kevinbuhmann (changes requested):
This should be rebased and the commit message should have something like the following:

```
BREAKING CHANGE:
- The pagination input will only change the current page on enter keypress.
- The pagination input will resets the input value to the actual current page on blur.
```

kevinbuhmann (approved):

PR #333, opened 2022-10-06
feat(tree-view): enable "type ahead"/letter key navigation by default

kevinbuhmann:
## PR Checklist

## PR Type

Feature
Breaking change

Letter key navigation is not enabled by default in the tree view, but all tree views should implement this. See VPAT-604 and https://www.w3.org/WAI/ARIA/apg/example-index/treeview/treeview-1/treeview-1a.html.

## What is the new behavior?

Letter key navigation works by default in the tree view.

Yes.
- The `ClrForTypeAhead` directive and the `ClrForTypeAheadModule` have been removed.
- If you are importing the `ClrForTypeAheadModule`, you can just remove it.
- If you are using the `ClrForTypeAhead` directive class, you will need to adjust your code.
- The `clrForTypeAhead` input will still work on `clr-tree-node` elements.
- If your tree nodes have non-visible text (e.g. screen reader text), you will likely need to set the `clrForTypeAhead` input to just the visible text for the letter key navigation to work properly.


Jinnie:
Okay, I may be missing some implementation details, but isn't this an incremental change?
Why do we consider it "breaking"? Is there our users will need to change on their side of code?

kevinbuhmann:
> Okay, I may be missing some implementation details, but isn't this an incremental change? Why do we consider it "breaking"? Is there our users will need to change on their side of code?

It could break code because I removed a module. Also, it may not work correctly with tree views that have screen reader text in the node unless a developer make some code changes to make it work properly.

Jinnie (approved):

PR #356, opened 2022-10-20
fix(forms): remove touched enforcement

### LINKED ISSUE #337: Clarity controls causes extra valueChanges event to be sent on Angular 14
## Describe the bug

When a clarity control loses focus, there is code to check if the control has been touched. If not, Clarity sets the touched state and updates the validity and value of the control. If this control subscribes to valueChanges event, it gets one initial erroneous one which can cause existing reactive form code to break. 

## How to reproduce

Update to Angular 14. Have a reactive form. Subscribe to valueChanges event and focus and blur the control.

https://stackblitz.com/edit/clarity-light-theme-clr13-cds6-sksq78

## Expected behavior

To behave like it did with Angular 13. No extra call to valueChanges event.

## Versions

**Clarity version:**

- [x] v14.x

**Framework:**

- [x] Angular
- [ ] React
- [ ] Vue
- [ ] Other:

**Framework version:**
_ie: Angular 11_

## Additional notes

A workaround is to put the ClarityModule after the Angular forms module in the modules imports. If it's located in front, the blur events seem to now happen in different order than in Angular 13. Clarity get its event first and therefor the control's touched state has not yet been set. If changing the module import order, now the control is touched by the time clarity's blur event happens.

I wonder if we still need the code that checks for touched in Clarity wrapped-control.ts file. It seems it was for a workaround for numbers in certain cases.

Jinnie:
This was a bug fix for old issue with input type=number that is not reproducible anymore. It causes extra valueChanged events on blur after upgrade to Angular v14.

closes #337

Signed-off-by: Ivan Donchev <idonchev@vmware.com>

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

There is an enforcement of the "touched" form status on "blur".
It causes a "valueChange" event too early in lifecycle. It's observable after upgrading to angular v14 and importing Clarity Module before FormsModule.
The order of imports should not matter!

Issue Number: 337

## What is the new behavior?

Enforcement removed.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/forms/common/wrapped-control.ts
+++ b/projects/angular/src/forms/common/wrapped-control.ts
@@ -103,13 +103,6 @@ export class WrappedFormControl<W extends DynamicWrapper> implements OnInit, OnD
   @HostListener('blur')
   triggerValidation() {
     if (this.ifControlStateService) {
-      /**
-       * For some reason the <input type="number" /> on blur ngControl doesn't set the control to 'touched'
-       * This one is a workaround to provide the control to be 'touched' on blur and fix #4480.
-       */
-      if (this.ngControl && !this.ngControl.touched) {
-        this.markAsTouched();
-      }
       this.ifControlStateService.triggerStatusChange();
     }
   }
--- a/projects/angular/src/forms/input/input.spec.ts
+++ b/projects/angular/src/forms/input/input.spec.ts
@@ -5,8 +5,11 @@
  */
 
 import { Component } from '@angular/core';
-import { FormControl, FormGroup, Validators } from '@angular/forms';
+import { TestBed } from '@angular/core/testing';
+import { FormControl, FormGroup, FormsModule, NgControl, ReactiveFormsModule, Validators } from '@angular/forms';
+import { By } from '@angular/platform-browser';
 
+import { ClrCommonFormsModule } from '../common';
 import { ControlStandaloneSpec, ReactiveSpec, TemplateDrivenSpec } from '../tests/control.spec';
 import { ClrInput } from './input';
 import { ClrInputContainer } from './input-container';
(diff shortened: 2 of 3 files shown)


projects/angular/src/forms/common/wrapped-control.ts
```diff
@@ -103,13 +103,6 @@ export class WrappedFormControl<W extends DynamicWrapper> implements OnInit, OnD
...
   triggerValidation() {
     if (this.ifControlStateService) {
-      /**
-       * For some reason the <input type="number" /> on blur ngControl doesn't set the control to 'touched'
-       * This one is a workaround to provide the control to be 'touched' on blur and fix #4480.
-       */
-      if (this.ngControl && !this.ngControl.touched) {
-        this.markAsTouched();
-      }
       this.ifControlStateService.triggerStatusChange();
```
Jinnie, line 106: I wanted to remove this one too, because it forces `verifyOn: blur` like behavior and overrides the built-in angular forms strategies.
It is a big breaking change though and it will not be welcomed by many users that seem to prefer this type of default.
We do not have any complaint about this, so I have stashed the bigger fix for now.

projects/angular/src/forms/input/input.spec.ts
```diff
@@ -39,5 +42,48 @@ export default function (): void {
     ControlStandaloneSpec(StandaloneUseTest);
     TemplateDrivenSpec(ClrInputContainer, ClrInput, TemplateDrivenTest, 'clr-input');
     ReactiveSpec(ClrInputContainer, ClrInput, ReactiveTest, 'clr-input');
+    inputSpec('tempate-driven', ClrInputContainer, ClrInput, TemplateDrivenTest);
+    inputSpec('reactive', ClrInputContainer, ClrInput, ReactiveTest);
+  });
+}
+
+function inputSpec(description, testContainer, testControl, testComponent) {
```
Jinnie, line 50: Though these tests seem to belong in `control.spec.ts` together with the other template driven and reactive tests, I have to make them input-specific, as many controls, like checkboxes, selects etc, don't emit `valueChanges`.

projects/angular/src/forms/input/input.spec.ts
```diff
@@ -39,5 +42,48 @@ export default function (): void {
...
+      TestBed.configureTestingModule({
+        imports: [FormsModule, ClrCommonFormsModule, ReactiveFormsModule],
+        declarations: [testContainer, testControl, testComponent],
+      });
+      fixture = TestBed.createComponent(testComponent);
+      control = fixture.debugElement.query(By.directive(testControl));
+      fixture.detectChanges();
+    });
+
+    it('should handle valueChanges calls', () => {
```
Jinnie, line 64: I cannot really make this test fail, because the failure occurs when ClarityModule and FormsModule are imported in specific order. But I am adding the test as documentation of a desired behavior and a sample how to create an out-of-the-spec manual test outside the source code.

projects/angular/src/forms/input/input.spec.ts
```diff
@@ -39,5 +42,48 @@ export default function (): void {
...
+      control.nativeElement.dispatchEvent(new Event('focus'));
+      control.nativeElement.dispatchEvent(new Event('blur'));
+      fixture.detectChanges();
+      expect(valueChanges).toBe(0);
+
+      // now make sure input change triggers valueChanges
+      control.nativeElement.dispatchEvent(new Event('focus'));
+      control.nativeElement.value = 'abc';
+      control.nativeElement.checked = true;
+      control.nativeElement.selected = true;
```
kevinbuhmann, line 81: What are these lines for? You can't check or select an `input` element.

projects/angular/src/forms/input/input.spec.ts
```diff
@@ -39,5 +42,48 @@ export default function (): void {
...
+      });
+      fixture = TestBed.createComponent(testComponent);
+      control = fixture.debugElement.query(By.directive(testControl));
+      fixture.detectChanges();
+    });
+
+    it('should handle valueChanges calls', () => {
+      // control must be both invalid and blurred to register the validity
+      let valueChanges = 0;
+      control.injector.get(NgControl).control.valueChanges.subscribe(() => {
```
kevinbuhmann, line 67: Isn't this test essentially testing Angular code, not Clarity-specific code? I don't think we should have tests that essentially re-test Angular.

kevinbuhmann (changes requested):

projects/angular/src/forms/input/input.spec.ts
Jinnie: Absolutely right. This is relic from my unsuccessful attempts to make a common test in the `control.spec.ts`. 
Thanks, I'll remove them.

projects/angular/src/forms/input/input.spec.ts
Jinnie: The 'valueChanges' are either triggered by the Angular Framework itself, or forcibly by us in the onBlur handler.
I want to test the effect on this observable, not the implementation detail from our side, as we used to do.
Essentially, the bug was that this handler gets called one extra time. Technically, the test is useless, because the failure can only be reproduced when Clarity is consumed as library and imported before the FormsModule.
For me, a unit test serves 2 purposes - provide testing automation and support the maintenance by defining use case scenarios.
Usually they serve both in the same time, but I added this test only with the second purpose in mind - to define the expectation and help understanding the use case if future problems arise in this area.

Jinnie:
@kevinbuhmann do we have any more expectations about this issue?

kevinbuhmann (approved):

Issue #370, opened 2022-11-01
Problems with clr-tooltip when clrPosition is not explicitly set


tsvetomirnik:
## Describe the bug

There is a rendering issue with tooltips when the `clrPosition` property is not explicitly set (on the `clr-tooltip-content`).

## How to reproduce

It can be seen in https://clarity.design/documentation/tooltips > Angular Components > Examples > Right:

<details>
  <summary>Show GIF with flickering content</summary>
  
  ![tootip-flashing-issue](https://user-images.githubusercontent.com/2995456/199239645-786eda09-91f2-4589-a530-c4b496f6c424.gif)
</details>

![image](https://user-images.githubusercontent.com/2995456/199234924-d1c3a123-850c-41c1-ba63-e14e87b8aaa3.png)

1.  Open the stackblitz https://stackblitz.com/edit/clarity-with-clr-angular-ui-3zkuc9
3.  Hover over the first ⓘ icon

## Actual behavior

The tooltip's content flashes over the ⓘ icon. When the icon is focused the content stays open but it's misaligned.

## Expected behavior

The tooltip's content should appear on the right of the ⓘ icon (the default direction for the angular component).

## Versions

**Clarity version:**

- [ ] v12.x
- [x] v13.x

**Framework:**

- [x] Angular
- [ ] React
- [ ] Vue

**Framework version:**
Angular 14

## Additional notes

The workaround is to set the `clrPosition`:

```html
<clr-tooltip>
    <clr-icon clrTooltipTrigger shape="info-circle" size="24"></clr-icon>
    <clr-tooltip-content clrPosition="right" *clrIfOpen>
        <span>Lorem ipsum sit</span>
    </clr-tooltip-content>
</clr-tooltip>
```

PR #373, opened 2022-11-02
fix(tooltip): apply default size and position

kevinbuhmann:
## PR Checklist

## PR Type

Bugfix

Tooltip content defaults are not applied correctly.

Issue Number: #370

## What is the new behavior?

Tooltip content defaults ar applied correctly.

No.

### DIFF
--- a/projects/angular/src/popover/tooltip/tooltip-content.ts
+++ b/projects/angular/src/popover/tooltip/tooltip-content.ts
@@ -4,7 +4,7 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { Component, ElementRef, Inject, Injector, Input, Optional } from '@angular/core';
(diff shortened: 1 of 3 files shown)


projects/angular/src/popover/tooltip/tooltip-content.ts
```diff
@@ -122,6 +122,11 @@ export class ClrTooltipContent extends AbstractPopover {
     this.updateCssClass({ oldClass: `tooltip-${oldSize}`, newClass: `tooltip-${newSize}` });
   }
 
+  ngOnInit() {
+    this.size = this.size || defaultSize;
```
ashleyryan, line 126: Just for my own curiosity, what's so different about this vs the other way the default was getting set (or trying to) before?

ashleyryan (approved):

PR #376, opened 2022-11-03
feat(datagrid): include column title in filter toggle button aria-label

kevinbuhmann:
## PR Checklist

## PR Type

Feature

The `aria-label` of all datagrid filter toggle buttons was "Toggle column filter".

Issue Number: VPAT-615

## What is the new behavior?

The `aria-label` of datagrid filter toggle buttons now contains the column name (e.g. "Toggle user filter"). This improves accessibility by providing a unique/specific `aria-label` for each filter toggle button.

No.

## Other information

If your application localizes the `datagridFilterAriaLabel` string, you will need to add the `{COLUMN}` placeholder to take advantage of this feature. It will work the same as before without the placeholder, so this change is backwards compatible.

### DIFF
--- a/projects/angular/src/data/datagrid/datagrid-filter.ts
+++ b/projects/angular/src/data/datagrid/datagrid-filter.ts
@@ -11,6 +11,7 @@ import {
   EventEmitter,
   Inject,
   Input,
+  OnChanges,
   OnDestroy,
   Output,
   PLATFORM_ID,
@@ -44,7 +45,7 @@ import { DatagridFilterRegistrar } from './utils/datagrid-filter-registrar';
       class="datagrid-filter-toggle"
       type="button"
       #anchor
-      [attr.aria-label]="commonStrings.keys.datagridFilterAriaLabel"
+      [attr.aria-label]="toggleButtonAriaLabel"
       [attr.aria-expanded]="ariaExpanded"
       [attr.aria-controls]="popoverId"
       clrPopoverAnchor
@@ -79,7 +80,7 @@ import { DatagridFilterRegistrar } from './utils/datagrid-filter-registrar';
 })
 export class ClrDatagridFilter<T = any>
   extends DatagridFilterRegistrar<T, ClrDatagridFilterInterface<T>>
-  implements CustomFilter, OnDestroy
+  implements CustomFilter, OnChanges, OnDestroy
 {
   private subs: Subscription[] = [];
   ariaExpanded = false;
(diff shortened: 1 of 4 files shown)


projects/angular/src/data/datagrid/datagrid-filter.ts
```diff
@@ -145,8 +147,29 @@ export class ClrDatagridFilter<T = any>
...
   override ngOnDestroy(): void {
     super.ngOnDestroy();
     this.subs.forEach(sub => sub.unsubscribe());
   }
+
+  /**
+   * This is not in a getter to prevent "expression has changed after it was checked" errors.
+   * And it's more performant this way since it only runs on change.
+   */
+  private setToggleButtonAriaLabel() {
```
Jinnie, line 165: Some time ago we were trying to keep Clarity working in SSR scenario. I think we already diverged from this goal. But it may be worth talking about it and evaluating if we want to support it and what the cost will be.

projects/angular/src/data/datagrid/datagrid-filter.ts
```diff
@@ -90,7 +91,8 @@ export class ClrDatagridFilter<T = any>
     _filters: FiltersProvider<T>,
     public commonStrings: ClrCommonStringsService,
     private smartToggleService: ClrPopoverToggleService,
-    @Inject(PLATFORM_ID) private platformId: any
+    @Inject(PLATFORM_ID) private platformId: any,
+    private elementRef: ElementRef<HTMLElement>
```
Jinnie, line 95: Can't we make this optional, so we don't change the API too much?

Jinnie: Hmm, logically it does not make sense. I was looking at the test when I wrote the above comment, but it's highly unlikely that anyone will use the constructor directly outside of testing scenarios. Please disregard the previous comment.

Jinnie (approved):

projects/angular/src/data/datagrid/datagrid-filter.ts
kevinbuhmann: Users should definitely not be calling directive/component constructors. I consider these parameters not part of the public API.

projects/angular/src/data/datagrid/datagrid-filter.ts
kevinbuhmann: I'll look into that.

PR #413, opened 2022-12-05
Fix test:watch target

Jinnie:
It didn't resolve "isWatch" properly before


kevinbuhmann (changes requested):
I think there is a better way than relying on `--watch` being the last arg. Also, I don't think the `test` script should be repeated in the `test:watch` script.

See #414 for an alternative.

PR #419, opened 2022-12-06
feat(wizard): remove deprecated, undocumented inline wizard mode

kevinbuhmann:
## PR Checklist

## PR Type

Remove deprecated feature.

Inline wizards are deprecated.

## What is the new behavior?

Inline wizards are removed.

Yes.

- The inline wizard mode has been removed.
- The `clrModalOverrideScrollService` modal input has been removed.


Jinnie (changes requested):
This will break the website. There are some website demos that show the wizard inline.
AFAIK, this is the only reason for this code to exist.
I don't insist to have these unrealistic demos on the website, but before we remove this code, we'd better have the website updated and uploaded.

kevinbuhmann:
> This will break the website. There are some website demos that show the wizard inline.
> AFAIK, this is the only reason for this code to exist.
> I don't insist to have these unrealistic demos on the website, but before we remove this code, we'd better have the website updated and uploaded.

I didn't see any, but I'll look again. I'll working on a v15 website that would include this change.

Jinnie (approved):
Ok, it had only remained in the demo app then, and I see you removed it from there.
Thanks!

PR #427, opened 2022-12-07
fix(datagrid): prevent issues with single select radio in forms

kevinbuhmann:
See vmware/clarity#5574 for more information.

## PR Checklist

## PR Type

Bugfix

Style/layout issues if a single select datagrid with row selection is inside a form.

## What is the new behavior?

No style/layout issues if a single select datagrid with row selection is inside a form.

No.


Jinnie (approved):
This will work for now, but in future we need to reevaluate the whole "row wrapping label" approach as it leads to associating two labels with the checkbox/radio inside the row. And this one does not really define a meaningful label association, only makes the whole row clickable.

PR #431, opened 2022-12-08
feat: remove `clrFocusTrap` in favor of `@angular/cdk`

kevinbuhmann:
## PR Checklist

  - This feature is not documented.

## PR Type

Remove undocumented feature.

Various components use custom focus trap functionality, and this custom functionality is exported by the library.

## What is the new behavior?

Various components use focus trap functionality from @angular/cdk, and the custom functionality is removed.

Yes.
- The `@angular/cdk` package is now a required peer dependency.
- The `clrFocusTrap` directive has been removed. Please use the `cdkTrapFocus` directive from `@angular/cdk` instead.


bbogdanov:
There is already a newer implementation of a focus trap in the code that proved itself useful.

https://github.com/vmware-clarity/ng-clarity/blob/main/projects/angular/src/utils/focus-trap/focus-trap.ts

What is the need of introducing a new package that provides the same thing and on top of that introducing a dependency we have no control of?

kevinbuhmann:
>What is the need of introducing a new package that provides the same thing and on top of that introducing a dependency we have no control of?

This is about redundancy. I'm proposing eliminating custom code for things that Angular provides. I think it is unnecessary/redundant/inefficient for us to maintain custom implementations of drag/drop, focus trap, list key focus, etc. These things are provided by Angular. It's just not the best use of time/people/the team.

The Angular CDK is maintained by the same team that maintains the Angular framework packages, so in my view, it's just part of depending on Angular. As a user, I'd expect an Angular component library to use and rely on the Component Development Kit because that's what it's for.

projects/demo/src/app/modal/modal-trap.demo.html
```diff
@@ -6,14 +6,14 @@
...
   For accessiblity purposes, you may want to trap user focus to within the modal. You can use the
-  <code>clrFocusTrap</code> directive to keep the user tabbing from within the modal.
+  <code>cdkTrapFocus</code> directive to keep the user tabbing from within the modal.
 </p>
 
 <div class="clr-example">
   <div class="backdrop-example-container">
     <div class="modal static">
       <div class="modal-dialog" role="dialog" aria-hidden="true">
-        <div clrFocusTrap class="modal-content">
```
Jinnie, line 16: For some reason, I don't see this working while manually testing. Looks ok in code, but the trap is missing here:
https://431--storybook-clarity-design.netlify.app/demo/modal/focus-trap

Jinnie (changes requested):
Looks great, apart from the demo page failing.
It may be demo-specific, as the modal is pre-open and inline, but it would be great if you can take a look.

bbogdanov:
> > What is the need of introducing a new package that provides the same thing and on top of that introducing a dependency we have no control of?
> 
> This is about redundancy. I'm proposing eliminating custom code for things that Angular provides. I think it is unnecessary/redundant/inefficient for us to maintain custom implementations of drag/drop, focus trap, list key focus, etc. These things are provided by Angular. It's just not the best use of time/people/the team.
> 
> The Angular CDK is maintained by the same team that maintains the Angular framework packages, so in my view, it's just part of depending on Angular. As a user, I'd expect an Angular component library to use and rely on the Component Development Kit because that's what it's for.

I would advice testing these changes in a real application before the actual release cause this could be a real issue if someone depends on the `span` stuff that the `clrFocusTrap` uses to trap the user at the moment. I could assist with that.

kevinbuhmann:
> Looks great, apart from the demo page failing.
> It may be demo-specific, as the modal is pre-open and inline, but it would be great if you can take a look.

@Jinnie: I just forgot to import the `A11yModule` for the static demo. It's fixed now.

Jinnie (approved):

elesueur:
@kevinbuhmann Adding a dependency on @angular/cdk is a significant architectural decision for Clarity Angular.

In this initial step, your desire to avoid duplication of a smallish piece of code to deal with focus trapping could cause bundle sizes to increase in all consuming applications by over 100K because the cdkFocusTrap directive is part of the @angular/cdk/a11y entrypoint. I would be interested to know how well the Angular 15 build system can tree-shake the unused code out. Have you done any experiments to see how the addition of this new dependency increases application bundle size?

kevinbuhmann:
>Adding a dependency on @angular/cdk is a significant architectural decision for Clarity Angular.

The CDK is built for exactly this use case: component library development. I don't see any reason why we should maintain custom code for things the framework provides for us.

kevinbuhmann:
My last two force pushes avoid importing all of `A11yModule` which results in about 6 kb in bundle savings. This makes the bundle size _smaller_ with the `cdkTrapFocus` than `clrFocusTrap`.

https://github.com/vmware-clarity/ng-clarity/compare/2fc35fd9155542c7dbe8abeb72af1f3173d88030..95c4b8f09f46670ded12f58965764a782000f0c6

Jinnie:
I support what Kevin shared above.
For a team of our size and composition, it's really important to clean up and minimize our maintenance surface.
We expect even bigger benefits from the CDK if we're successful in replacing the many arrow key navigation solutions that we have all over Clarity, several of them custom, with their own specifics and problems, and at least two, which more generic and reusable.
We also have certain researches around column reordering and virtual scrolling, which are based on the CDK. Though they are far from productization yet, we may want to see them reprioritized at some point in future, as these are quite common requests we receive.

projects/demo/src/app/modal/modal-trap.demo.html
kevinbuhmann: https://github.com/vmware-clarity/ng-clarity/pull/431#issuecomment-1344410185

PR #302, opened 2022-08-29
fix(datagrid): detail-pane overlaping the content in smaller container

bbogdanov:
Signed-off-by: Bogdan Bogdanov <bbogdanov@vmware.com>

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: https://github.com/vmware/clarity/issues/6429

image

## What is the new behavior?

The detail-pane doesn't overlap the content of the datagrid table anymore when datagrid is in a smaller container.

image

FYI: VMC uses this approach for mostly a month and clients seems to not have issues with it anymore.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #449, opened 2023-01-06
fix(datagrid): show full border when detail pane is open

kevinbuhmann:
This fixes a regression that was introduced in #302.

## PR Checklist

## PR Type

Bugfix

The border between the column and the detail pane does not extend to the placeholder element below the rows.

image

## What is the new behavior?

The border between the column and the detail pane extends to the placeholder element below the rows.

image

No.

### DIFF
--- a/projects/angular/src/data/datagrid/_datagrid.clarity.scss
+++ b/projects/angular/src/data/datagrid/_datagrid.clarity.scss
@@ -1695,12 +1695,12 @@
 
   .datagrid-detail-open {
     & > .datagrid-outer-wrapper > .datagrid-inner-wrapper {
-      div.datagrid-table-wrapper {
+      div.datagrid-table {
         /**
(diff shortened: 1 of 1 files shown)


bbogdanov:
The issue with this will come when you have a very long text without spaces. Usually such text is very long uuid separated by dashes on the first column.

kevinbuhmann:
> The issue with this solution will come when you have a very long text without spaces. Usually such text is very long uuid separated by dashes inside the first column.

I'll double check, but that was one of my test cases. I think it worked the same as before.

kevinbuhmann:
@bbogdanov, this looks correct to me. The "arrow" on the row and the column filter button are still visible. Let me know if I missed something.

image

Jinnie (approved):

bbogdanov (approved):

PR #453, opened 2023-01-10
chore: rename `ELement` to `Element`

kevinbuhmann:


Jinnie (approved):
Any chance this was intentional to avoid collision with https://developer.mozilla.org/en-US/docs/Web/API/Element ?

I will approve it, but if think it's worth it, we may add a more descriptive name.

kevinbuhmann:
> Any chance this was intentional to avoid collision with https://developer.mozilla.org/en-US/docs/Web/API/Element ?
> 
> I will approve it, but if think it's worth it, we may add a more descriptive name.

That's a good point. I added this interface, and I didn't use casing for disambiguation. I don't think it's worth coming up with a better name though.

PR #457, opened 2023-01-10
chore: various cleanup around key events

kevinbuhmann:
## PR Checklist

## PR Type

Refactoring (no functional changes, no api changes)

No.


projects/angular/src/utils/enums/key-codes.enum.ts
```diff
@@ -26,8 +26,4 @@ export enum IEKeyCodes {
   ArrowLeft = 'Left',
   Space = 'Spacebar',
   Escape = 'Esc',
-  Home = 'Home',
-  End = 'End',
-  Enter = 'Enter',
-  Tab = 'Tab',
```
Jinnie, line 32: As this is a utility, I'd prefer to keep the unused enumeration values for completeness.

projects/angular/src/utils/key-codes/key-codes.ts
```diff
@@ -8,9 +8,4 @@ export const UP_ARROW = 38;
 export const DOWN_ARROW = 40;
 export const RIGHT_ARROW = 39;
 export const LEFT_ARROW = 37;
-export const ENTER = 13;
-export const SPACE = 32;
-export const SHIFT = 16;
-export const TAB = 9;
-export const ESC = 27;
```
Jinnie, line 15: Same here:
As this is a utility, I'd prefer to keep the unused enumeration values for completeness.
It's not a big overhead.

kevinbuhmann: These aren't used at all anymore since I changed everything to `event.key` instead of `event.keyCode`.

projects/angular/src/utils/enums/key-codes.enum.ts
kevinbuhmann: These aren't different for IE, so they're not needed. See the other enum.

Jinnie (approved):

PR #466, opened 2023-01-12
fix(modal): prevent scrolling content under modal when tabbing

kevinbuhmann:
This fixes a regression I introduced in #431.

## PR Checklist

## PR Type

Bugfix

The content behind the modal would be scrolled when the user tabs through the modal.

## What is the new behavior?

The content behind the modal is not scrolled when the user tabs through the modal.

No.


kevinbuhmann:
This was released in 15.0.0-beta.5.

PR #482, opened 2023-01-26
fix(datepicker): remove incorrect `role="heading"` from table header

kevinbuhmann:
## PR Checklist

## PR Type

Bugfix

Invalid aria markup in the datepicker table.

Issue Number: VPAT-661

## What is the new behavior?

Valid aria markup in the datepicker table.

No.


Jinnie (dismissed):

projects/angular/src/forms/datepicker/calendar.html
```diff
@@ -1,7 +1,7 @@
 <table class="calendar-table">
   <tr class="calendar-row weekdays">
-    <th *ngFor="let day of localeDays" class="calendar-cell weekday" role="heading" [attr.aria-label]="day.day">
-      {{day.narrow}}
+    <th *ngFor="let day of localeDays" class="calendar-cell weekday">
+      <span role="text" [attr.aria-label]="day.day">{{day.narrow}}</span>
```
Jinnie, line 4: What is the reason to use role="text" here?
It is invalid ARIA role, that was proposed, but not made it to official spec and is sometimes used to workaround specific screenreader issues, which I do not see in this case.

PR #426, opened 2022-12-07
feat: remove `ClrDragAndDropModule` in favor of `@angular/cdk`

kevinbuhmann:
## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features)
- [N/A] Docs have been added / updated (for bug fixes / features)
  - This feature is not documented.
- [N/A] If applicable, have a visual design approval

## PR Type

Remove undocumented feature.

## What is the current behavior?

The datagrid uses custom drag-drop functionality, and this custom functionality is exported by the library.

## What is the new behavior?

The datagrid uses drag-drop functionality from `@angular/cdk`, and the custom functionality is removed.

## Does this PR introduce a breaking change?

Yes.
- The `@angular/cdk` package is now a required peer dependency.
- The `ClrDragAndDropModule` has been removed. Please use the `DragDropModule` module from `@angular/cdk` instead.

PR #484, opened 2023-01-29
fix(drag-and-drop): clear constant collection on component destroy

Arooba-git:
Unwanted references to the FILES  collection is causing memory to leak. It is recommended to clear all large collections explicitly when component destroys.

Signed-off-by: Arooba Shahoor <arooba.shahoor@gmail.com>

Hello 👋 
As part of our project, we are using Facebook's new Memlab tool to detect memory leaks in SPA applications. 
While running the tool and analyzing the code of ng-clarity, we saw that your project does a very good job of ensuring that all async operations are cancelled when the component unmounts. However, as per Memlab execution results, we found a dangling collection that was causing the memory to leak (screenshots below).

[before]
Screen Shot 2023-01-28 at 12 10 49 PM

<br />
Hence we added the fix by clearing the collection in destructor and you can see the heap size and # of leaks reducing noticeably:
 <br />
Screen Shot 2023-01-28 at 12 43 08 PM

Note: We used `splice` instead of the more conventional approach of `.length = 0`, so as to cater for any future modifications such as changing the FILES object to a type that has a `length` property but is not assignable.

As per the contribution guidelines, we ran the unit test cases; all 3547 tests were completed successfully.
You can analyze this and other potential leak sources, if you like, by running [Memlab](https://facebook.github.io/memlab/) with a scenario file covering the maximum # of use cases.

Following is a sample of the scenario file we used (it needs to be a .js file but attaching here in .txt form):
[ng-clarity-memlab-scenario.txt](https://github.com/vmware-clarity/ng-clarity/files/10528755/ng-clarity-memlab-scenario.txt)

Note that some other reported leaks (in Memlab) originated from Angular's internal objects, and hence were ignored.

### DIFF
--- a/projects/demo/src/app/drag-and-drop/grouping.demo.ts
+++ b/projects/demo/src/app/drag-and-drop/grouping.demo.ts
@@ -4,7 +4,7 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { Component } from '@angular/core';
+import { Component, HostListener, OnDestroy } from '@angular/core';
 import { ClrDragEvent } from '@clr/angular';
 
 const FILES = [
@@ -18,7 +18,7 @@ const FILES = [
   styleUrls: ['./drag-and-drop.demo.scss'],
   templateUrl: './grouping.demo.html',
 })
-export class GroupingDemo {
+export class GroupingDemo implements OnDestroy {
   files: any[];
 
   droppedImages: any[] = [];
@@ -56,6 +56,13 @@ export class GroupingDemo {
       this.moveItem(dragEvent.dragDataTransfer, this.files, this.droppedMovies);
     } else if (this.droppedImages.indexOf(dragEvent.dragDataTransfer) > -1) {
       this.moveItem(dragEvent.dragDataTransfer, this.files, this.droppedImages);
+    }
+  }
+
+  @HostListener('unloaded')
+  ngOnDestroy(): void {
+    if (FILES && FILES.length) {
+      FILES.splice(0, FILES.length);
     }
   }
 }


projects/demo/src/app/drag-and-drop/grouping.demo.ts
```diff
@@ -58,4 +58,11 @@ export class GroupingDemo {
       this.moveItem(dragEvent.dragDataTransfer, this.files, this.droppedImages);
     }
   }
+
+  @HostListener('unloaded')
+  ngOnDestroy(): void {
+    if (FILES && FILES.length) {
+      FILES.splice(0, FILES.length);
```
kevinbuhmann, line 65: Thanks for your analysis, but we cannot accept this fix as is because it actually causes a bug in or demo app.

Since the `FILES` array is owned by the module, not the component class, this actually causes a bug in the demo app.

1. Navigate to https://484--storybook-clarity-design.netlify.app/demo/drag-and-drop/grouping.
2. Click a link for another drag and drop page (e.g. “Basic Draggable”).
3. Click the “Grouping” link.
4. Notice that there are no files to drag this time.

This problem can likely be fixed by removing the `FILE` constant array and inlining its contents in the constructor when the `files` property is initialized. This way, the `ngOnDestroy` logic can be removed entirely since the `files` class property will be dropped by garbage collection when the class instance is dropped.  

That said, it may not be worth it to test my proposed fix because this code is removed on our `beta` branch (which will be promoted `main` soon). See #426.

kevinbuhmann (changes requested):

kevinbuhmann:
This code no longer exists on our `main` branch.

PR #494, opened 2023-02-04
chore: add @clr/ui dark theme to storybook

kevinbuhmann:
## PR Checklist

- [N/A] Tests for the changes have been added (for bug fixes / features)
- [N/A] Docs have been added / updated (for bug fixes / features)
- [N/A] If applicable, have a visual design approval

## PR Type

Storybook feature.

## What is the current behavior?

- The @clr/ui dark theme is not available in the storybook.
- The @clr/ui app background color is not applied in the storybook.
- The @cds/core styles are applied in the storybook when a @cds/core theme is not selected.

## What is the new behavior?

- The @clr/ui dark theme is available in the storybook.
- The @clr/ui app background color is applied in the storybook.
- The @cds/core styles are not applied in the storybook unless a @cds/core theme is selected.

## Does this PR introduce a breaking change?

No.

PR #495, opened 2023-02-04
feat: apply app background color to `body` element

kevinbuhmann:
_I need this change to add the @clr/ui Dark Theme to the storybook. (#494)_

I don't think a `.main-container` element should be required to get the app background color. This change also applies the app background in the storybook.

## PR Checklist

## PR Type

Feature

- The app background color is applied to the `.main-container` element.
- The @clr/ui Light Theme in the storybook doesn't have the app background color.

## What is the new behavior?

- The app background color is applied to the `body` element.
- The @clr/ui Light Theme in the storybook has the app background color.

No.

### DIFF
--- a/projects/angular/src/layout/main-container/_layout.clarity.scss
+++ b/projects/angular/src/layout/main-container/_layout.clarity.scss
@@ -9,7 +9,6 @@
     display: flex;
     flex-direction: column;
     height: 100vh;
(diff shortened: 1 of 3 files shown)


kevinbuhmann:
I found a different way to solve the problem in the storybook, so I am closing this for now.

PR #497, opened 2023-02-07
fix(ci): add support for build/lint/test against multiple angular versions

williamernest:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] CI related changes

Update the pr build actions to run the build, lint and test commands against multiple versions of Angular. 

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -12,6 +12,7 @@ jobs:
     strategy:
       matrix:
         cds-core: ['', '@cds/core@^5.6.0']
+        angular: ['13', '14', '15']
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -22,10 +23,16 @@ jobs:
       - name: Install Dependencies
         run: npm ci
       - name: Install Alternate @cds/core Version
-        if: ${{matrix.cds-core != ''}}
+        if: ${{matrix.cds-core == '' && matrix.angular == '13'}}
         run: npm install ${{matrix.cds-core}}
+      - name: Install ng 14
+        if: ${{matrix.cds-core == '' && matrix.angular == '14'}}
+        run: npm install @angular/core@^14.2.12 @angular/common@^14.2.12
+      - name: Install ng 15
+        if: ${{matrix.cds-core == '' && matrix.angular == '15'}}
+        run: npm install @angular/core@^15.1.3 @angular/common@^15.1.3
       - name: Disable Docs Build
-        if: ${{matrix.cds-core != ''}}
+        if: ${{matrix.cds-core == '' && matrix.angular == '13'}}
         run: node ./scripts/clear-npm-script.js _build:demo _build:storybook
       - name: Lint
         run: npm run lint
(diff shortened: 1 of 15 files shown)


.github/workflows/pr-build.yml
```diff
@@ -22,10 +23,16 @@ jobs:
       - name: Install Dependencies
         run: npm ci
       - name: Install Alternate @cds/core Version
-        if: ${{matrix.cds-core != ''}}
+        if: ${{matrix.cds-core == '' && matrix.angular == '13'}}
         run: npm install ${{matrix.cds-core}}
```
Jinnie, line 27: This should probably be `matrix.angular`?
I'd guess that's what's leading to the current error in `PR Build / PR Build (14) `

williamernest: I've got it running now, it checks the cds-core integration in both angular versions. Not sure we need it, but it can't hurt.

Jinnie: Yeah, with this check it makes sense. Thanks

kevinbuhmann (changes requested):
The CI change is a “chore” not a “fix” because it shouldn’t trigger a release or be in the change log. (Also, there are two commits here with the same message.)

This PR has several extra things that aren’t needed to get what we want. See comments. Also, I think it would be better to just disable the storybook build but not the demo app build. This way we are building an app that consumes Clarity in Angular v14. This should give us a bit more confidence that just building the library and running the unit tests.

And for Angular v15, I think we should just go ahead and add Angular v15 to our peer dependency. (I would split the v15 change into a separate "feat" commit.)

See my PR for a smaller set of changes to accomplish this: #504

package.json
```diff
@@ -28,7 +29,7 @@
     "test:watch": "npm run --silent test -- --watch",
     "_build:angular": "ng build clr-angular --configuration production",
     "_build:demo": "ng build clr-demo --configuration production",
-    "_build:storybook": "npm run docs:json && build-storybook --output-dir ./dist/docs",
+    "_build:storybook": "npm run docs:json && STORYBOOK_ANGULAR_PROJECT=clr-angular build-storybook --output-dir ./dist/docs",
```
kevinbuhmann, line 32: This change isn't needed since we are skipping the storybook build for Angular v14 and v15.

package.json
```diff
@@ -50,7 +51,7 @@
     "_lint:styles:fix": "npm run --silent _lint:styles -- --fix",
     "_lint:styles:changed": "node ./scripts/get-changed-files.js --command \"stylelint\" --filter \"**/*.{css,scss,sass}\"",
     "_lint:styles:changed:fix": "node ./scripts/get-changed-files.js --command \"stylelint --fix\" --filter \"**/*.{css,scss,sass}\"",
-    "_start:storybook": "wait-on ./documentation.json ./dist/clr-ui/clr-ui.css && start-storybook -p 6006"
+    "_start:storybook": "wait-on ./documentation.json ./dist/clr-ui/clr-ui.css && STORYBOOK_ANGULAR_PROJECT=clr-angular start-storybook -p 6006"
```
kevinbuhmann, line 54: This change isn't needed since we are skipping the storybook build for Angular v14 and v15.

.npmrc
```diff
@@ -1 +1,2 @@
 engine-strict=true
+legacy-peer-deps=true # workaround for https://github.com/storybookjs/storybook/issues/18094
```
kevinbuhmann, line 2: This change isn't needed here.

williamernest:
Closed in favor of #504

PR #499, opened 2023-02-08
Revert "fix(vertical-nav): add aria list structure to vertical nav links"

kevinbuhmann:
This reverts commit part of #480.

Adding `role="listitem"` to `a` element overrides the link role.


Jinnie (approved):
I see. Yeah, it can be confusing

PR #504, opened 2023-02-13
feat/chore: run pull request build/test against Angular v13, v14, and v15 (and widen peer dependencies to allow Angular v15)

kevinbuhmann:


williamernest (approved):
LGTM

Issue #508, opened 2023-02-17
Shim - Dark mode shows white manage columns button


rbirkgit:
## Describe the bug

The manage column button in the grid is white in dark mode when using the shim.

## How to reproduce

Use shim and switch to dark mode

## Expected behavior

Dark button 

## Versions

**Clarity version:**

- [x] v13.14.0
- [x] v15.0.0 rc2
- [ ] Other:

williamernest:
Resolved in #514

PR #514, opened 2023-02-22
fix(datagrid): resolve manage columns and popover shim color issues

williamernest:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The manage columns button in the core light and dark theme is incorrect. 
The datagrid action overflow popover selected/hover background color is incorrect
Buttons throughout the library are incorrectly receiving a filter. 

Issue Number: #508 

## What is the new behavior?

Maps the colors of the manage columns button to the correct core design tokens. 
Fixes the datagrid action overflow popover background color by mapping it to the correct value.
Adds the focus state to the manage columns button (to align with the active state).
Removes unexpected filter on all btn elements based on state. 

- [ ] Yes
- [x] No

### DIFF
--- a/projects/ui/src/shim.cds-core.scss
+++ b/projects/ui/src/shim.cds-core.scss
@@ -521,7 +521,7 @@
   --clr-dropdown-selection-color: var(--cds-alias-object-interaction-background-selected); // var(--clr-global-selection-color);
   --clr-dropdown-box-shadow: var(--clr-popover-box-shadow-color);
 
-  --clr-dropdown-text-color: var(--clr-p1-color);
+  --clr-dropdown-text-color: var(--clr-global-font-color);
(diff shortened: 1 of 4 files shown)


projects/ui/src/shim.cds-core.scss
```diff
@@ -752,12 +764,6 @@
     }
   }
 
-  .btn:not([disabled]):hover,
```
williamernest, line 755: @colinreedmiller One thing to consider here is that users may expect all buttons across their application to apply these values, so I wanted to double check that we're comfortable removing these values from the shim.

colinreedmiller: Without usage data the question may be more about whether or not we can support the requests for assistance when teams encounter unexpected results. It may be wiser to retain these values until we have a migration plan that addresses this.

kevinbuhmann (approved):

Issue #157, opened 2022-06-08
Disabling datagrid row selection not working with `*ngFor`


zhichl:
## Describe the bug

We (NSX-ALB team) are setting:`[clrDgSelectable]="false"` for `<clr-dg-row>` as the snippet shown:
```
    <clr-dg-row
        *ngFor="let row of rows; index as i; trackBy: trackByRowName"
        [clrDgItem]="row"
        [clrDgSelectable]="false"
    >
```
However, the leading checkbox of a row in the UI is not grayed out for selection.
<img width="540" alt="image" src="https://user-images.githubusercontent.com/5053922/172533939-0982af90-e766-4590-a7a8-b49fad955eba.png">

I found a similar bug which was given a workaround: https://github.com/vmware/clarity/issues/4993.
We’ve tried the workaround but the bug is still present. Has there been a known solution for this issue?

## How to reproduce

Please check: https://stackblitz.com/edit/clarity-light-theme-clr13-cds6-hmw4an?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts

1.  Click on the link to enter the project.
2.  Check `app.component.html` line 30 to see `[clrDgSelectable]` is set to `false`, save and look at the layout on the right side for checkboxes.
3.  Go to `app.component.html`, comment out line 22 through line 26 and comment line 27 through line 31. This will bind `[clrDgSelectable]` to a function call which always returns `false`. Save and look at the layout on the right side for checkboxes.

## Expected behavior

Under either case from the reproduction part, checkboxes are not disabled.

## Versions

**Clarity project:**

ng-clarity

**Clarity version:**
- v5.x
- v6.x

**Framework:**

Angular

**Framework version:**
_Angular 12_

**Device:**

- Type: MacBook
- OS: MacOS
- Browser: Chrome

kevinbuhmann:
The `clrDgSelectable` input only works with the `*clrDgItems` smart iterator. I will look into into making it work it work with `*ngFor` as well.

zhichl:
@kevinbuhmann Any updates on how this bug will be handled?

kevinbuhmann:
Not at this time. I spent some time on this and wasn't able to figure out why it's not working in your StackBlitz. It works in our Storybook. I was going to consult with my teammate next week to see if he has any idea as he knows the datagrid better than I do.

[...]

PR #531, opened 2023-03-01
fix(datagrid): set non-selectable multi select rows to disabled on initial render

williamernest:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, when using a multi selectable datagrid, the rows do not render in their correct state (based on clrDgSelectable). Since the clrDgSelectable and item values are both Inputs, the clrDgSelectable state may be set within the component prior to the item, which results in the selectable state being lost. This change introduces a private local variable to keep track of the clrDgSelectable state, and introduces a setter for the item that sets the value when item is available. 

Issue Number: #157 

## What is the new behavior?
When rendering a multi-select datagrid, rows will properly respect the clrDgSelectable input when initially rendered. 

- [ ] Yes
- [ x] No

### DIFF
--- a/.storybook/stories/datagrid/datagrid-row.stories.ts
+++ b/.storybook/stories/datagrid/datagrid-row.stories.ts
@@ -40,7 +40,7 @@ const defaultStory: Story = args => ({
       <clr-dg-row
         *clrDgItems="let element of elements; let index = index"
         [clrDgExpanded]="clrDgExpanded && index === 0"
-        [clrDgSelectable]="clrDgSelectable && index === 0"
+        [clrDgSelectable]="index !== 0 || clrDgSelectable"
         [clrDgSelected]="clrDgSelected && index === 0"
         [clrDgDetailOpenLabel]="clrDgDetailOpenLabel"
         [clrDgDetailCloseLabel]="clrDgDetailCloseLabel"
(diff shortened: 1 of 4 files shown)


.storybook/stories/datagrid/datagrid-row.stories.ts
```diff
@@ -40,7 +40,7 @@ const defaultStory: Story = args => ({
       <clr-dg-row
         *clrDgItems="let element of elements; let index = index"
         [clrDgExpanded]="clrDgExpanded && index === 0"
-        [clrDgSelectable]="clrDgSelectable && index === 0"
```
williamernest, line 43: We should consider adding docblocks in storybook to give better descriptions to what the user is expected to see with our parameters. It's not apparent that most of these are designed to only apply to the first row.

projects/angular/src/data/datagrid/datagrid-row.spec.ts
```diff
@@ -134,6 +134,20 @@ export default function (): void {
         TestBed.get(Items).all = [{ id: 1 }, { id: 2 }];
       });
 
+      it('renders correctly if item is set after clrDgSelectable', fakeAsync(function () {
+        selectionProvider.selectionType = SelectionType.Multi;
+        const tempItem = context.testComponent.item;
+        context.testComponent.item = undefined;
+        context.detectChanges();
+        context.testComponent.clrDgSelectable = false;
```
williamernest, line 142: In some cases, the clrDgSelectable Input is called before clrDgItem  value. This test replicates this by forcing the item to be undefined, and then setting the clrDgSelectable, and then repopulating the item.

kevinbuhmann (approved):

PR #539, opened 2023-03-03
fix(datagrid): add `clrDgItemsTrackBy` workaround

### LINKED ISSUE #157: Disabling datagrid row selection not working with `*ngFor`

kevinbuhmann:
closes #157

## PR Checklist

  - I will open a website PR.

## PR Type

Bugfix

You cannot use `[clrDgSelectable]` with `*ngFor` due to use of the iterator `trackBy` function to track selection.

It is impossible (or very difficult) to grab the correct `trackBy` function via the row iterator structural directive. The `ngForTrackBy` directive would grab the `trackBy` function from any `*ngFor` in the datagrid, not just the row iterator.

Further, needing the item index to track selection introduces a race condition when `*ngFor` is used. We don't have the collection of items when the rows inputs are set, so `clrDgSelectable` doesn't work.

Issue Number: #157

## What is the new behavior?

The fix is to pass a "track by" function directly to the datagrid. This change introduces a `clrDgItemsTrackBy` input that will used if provided. Otherwise, the datagrid will fall back to the previous behavior of trying to grab the `trackBy` function from the row iterator structural directive.

No.

## Other information

The fallback behavior is deprecated and scheduled for removal in v17.


projects/angular/src/data/datagrid/providers/items.ts
```diff
@@ -24,8 +26,15 @@ export class Items<T = any> {
...
+   *
+   * @deprecated in v15 and scheduled for removal in v17
+   */
+  iteratorTrackBy: TrackByFunction<T> = (_index, item) => item;
+
+  /**
+   * New tracking function to identify objects. If provided, this will be used instead of `iteratorTrackBy`.
    */
-  trackBy: TrackByFunction<T> = (_index, item) => item;
+  datagridTrackBy: ClrDatagridItemsTrackByFunction<T>;
```
williamernest, line 38: Adding a trackBy getter here would reduce all the if/else logic in selection.js, and make this change easier to remove in the future. 

```
get trackBy() { 
  return this.datagridTrackBy || this.iteratorTrackBy;
}
```

projects/angular/src/data/datagrid/providers/items.ts
```diff
@@ -13,6 +13,8 @@ import { FiltersProvider } from './filters';
 import { Page } from './page';
 import { Sort } from './sort';
 
+export type ClrDatagridItemsTrackByFunction<T> = (item: T) => any;
```
williamernest, line 16: Is there a reason we need the new type here? Are we deviating from the current type so the docs get updated?

projects/angular/src/data/datagrid/providers/items.ts
kevinbuhmann: Unfortunately, that won't work because these functions do not have compatible signatures.

projects/angular/src/data/datagrid/providers/items.ts
kevinbuhmann: The Angular `TrackByFunction<T>` take the index as the first parameter. I removed that parameter because the index of the item isn't available when `clrDgSelectable` is set because `items.all` isn't set until `ClrDatagrid#ngAfterContentInit`.

projects/angular/src/data/datagrid/datagrid.spec.ts
```diff
@@ -448,13 +448,50 @@ class TabsIntegrationTest {
...
+    <clr-datagrid [clrDgItemsTrackBy]="datagridTrackById">
+      <clr-dg-column>Item</clr-dg-column>
+      <clr-dg-row *ngFor="let item of items; trackBy: iteratorTrackBy" [clrDgItem]="item">
+        <clr-dg-cell>{{ item.id }}</clr-dg-cell>
+      </clr-dg-row>
+      <clr-dg-detail *clrIfDetail></clr-dg-detail>
+    </clr-datagrid>
+  `,
+})
+class PanelDatagridAndIteratorTrackByTest {
```
williamernest, line 487: Can you add a test case here to match the use case we're addressing with this PR (iteratorTrackBy without datagridTrackBy)?

projects/angular/src/data/datagrid/providers/items.ts
williamernest: Gotcha. I still think a proxy function here would prevent us from having to add the ternary and if/else logic throughout selection.js and would simplify removing it later. 

```
  get trackBy() {
    return (_index, item) => this.datagridTrackBy ? this.datagridTrackBy(item) : this.iteratorTrackBy(_index, item);
  }
```

kevinbuhmann: That won't work. I need to be able to call `datagridTrackBy` without the index.

projects/angular/src/data/datagrid/datagrid.spec.ts
kevinbuhmann: I did. `PanelIteratorTrackByTest`.

projects/angular/src/data/datagrid/providers/items.ts
kevinbuhmann: I can try conditionally looking up the index, but the code would still have conditionally and potentially be more confusing because it would have to conditionally pass `undefined` as the index without a clear reason why at the callsite.

projects/angular/src/data/datagrid/datagrid.spec.ts
williamernest: None of these tests validate the original bug exists, or is fixed. Please add tests asserting that when configured as a multi-select datagrid and a row has ` [clrDgSelectable="false"]`, the checkboxes render in the disabled state when using datagridTrackById and they render in a non-disabled state when using just the iteratorTrackBy

projects/angular/src/data/datagrid/providers/items.ts
williamernest: I'll leave it up to you. The current instance is duplicative and can lead to confusion. There are a lot of ways to proxy this and make it easier to implement and remove later.

projects/angular/src/data/datagrid/providers/items.ts
```diff
@@ -24,8 +26,15 @@ export class Items<T = any> {
 
   /**
    * Tracking function to identify objects. Default is reference equality.
+   *
+   * @deprecated in v15 and scheduled for removal in v17
+   */
+  iteratorTrackBy: TrackByFunction<T> = (_index, item) => item;
```
williamernest, line 32: Please file a ticket to remove this function in v17 and add a reference (e.g. `TODO(CDE-[n])` in this code block.

kevinbuhmann: Good idea.

projects/angular/src/data/datagrid/providers/items.ts
kevinbuhmann: I wrote a proxy for this for this. There is a precondition that must be met before the proxy `trackBy` function is used (`items.datagridTrackBy` or `items.all` must be set), but the code isn't as confusing as I thought it would be.

projects/angular/src/data/datagrid/providers/items.ts
kevinbuhmann: Done.

projects/angular/src/data/datagrid/datagrid.spec.ts
kevinbuhmann: Done.

williamernest: Thanks for adding.

williamernest (approved):
LGTM, thanks for making the changes.

PR #543, opened 2023-03-08
fix(button): add keyboard navigation in button-group menu

dtsanevmw:
## PR Checklist

- [ x ] Tests for the changes have been added (for bug fixes / features)

## PR Type
Bugfix

- [ x ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Keyboard navigation in Button Group menu is not working
Issue Number: CDE-68

## What is the new behavior?
Keyboard navigation once the menu in a Button Group is opened it's now working.

- [ ] Yes
- [ x ] No

### DIFF
--- a/projects/angular/src/button/button-group/button-group.ts
+++ b/projects/angular/src/button/button-group/button-group.ts
@@ -4,36 +4,55 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { AfterContentInit, Component, ContentChildren, Input, QueryList } from '@angular/core';
-import { takeUntil } from 'rxjs/operators';
+import {
+  AfterContentInit,
+  AfterViewInit,
+  Component,
+  ContentChildren,
+  ElementRef,
+  Input,
+  QueryList,
+  ViewChild,
+} from '@angular/core';
+import { delay, takeUntil } from 'rxjs/operators';
 
 import { ClrDestroyService } from '../../utils/destroy/destroy.service';
+import { FOCUS_SERVICE_PROVIDER } from '../../utils/focus/focus.service';
 import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
 import { uniqueIdFactory } from '../../utils/id-generator/id-generator.service';
 import { ClrPopoverPositions } from '../../utils/popover/enums/positions.enum';
 import { ClrPopoverPosition } from '../../utils/popover/interfaces/popover-position.interface';
 import { PopoverHostDirective } from '../../utils/popover/popover-host.directive';
 import { ClrPopoverToggleService } from '../../utils/popover/providers/popover-toggle.service';
+import {
+  BUTTON_GROUP_FOCUS_HANDLER_PROVIDER,
+  ButtonGroupFocusHandler,
+} from '../providers/button-group-focus-handler.service';
+import { InitialFocus } from '../providers/button-group-focus.enum';
 import { ButtonInGroupService } from '../providers/button-in-group.service';
 import { ClrButton } from './button';
 
 @Component({
   selector: 'clr-button-group',
   templateUrl: 'button-group.html',
-  providers: [ButtonInGroupService, ClrDestroyService],
+  providers: [ButtonInGroupService, ClrDestroyService, BUTTON_GROUP_FOCUS_HANDLER_PROVIDER, FOCUS_SERVICE_PROVIDER],
   hostDirectives: [PopoverHostDirective],
   host: { '[class.btn-group]': 'true' },
 })
-export class ClrButtonGroup implements AfterContentInit {
+export class ClrButtonGroup implements AfterContentInit, AfterViewInit {
+  @ViewChild('buttonGroupToggle') private buttonGroupToggle: ElementRef;
+  @ViewChild('buttonGroupMenu') private buttonGroupMenu: ElementRef;
   @ContentChildren(ClrButton) buttons: QueryList<ClrButton>;
 
   popoverId = uniqueIdFactory();
+  InitialFocus = InitialFocus;
 
   constructor(
     public buttonGroupNewService: ButtonInGroupService,
     private toggleService: ClrPopoverToggleService,
     public commonStrings: ClrCommonStringsService,
-    private destroy$: ClrDestroyService
+    private destroy$: ClrDestroyService,
+    private focusHandler: ButtonGroupFocusHandler
   ) {}
 
   popoverPosition: ClrPopoverPosition = ClrPopoverPositions['bottom-left'];
@@ -56,6 +75,10 @@ export class ClrButtonGroup implements AfterContentInit {
     this.buttons.changes.subscribe(() => {
       this.initializeButtons();
     });
+  }
+
+  ngAfterViewInit() {
+    this.handleFocusOnMenuOpen();
   }
 
   /**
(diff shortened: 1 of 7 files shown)


projects/angular/src/button/button-group/button-group.ts
```diff
@@ -4,27 +4,42 @@
...
 @Component({
   selector: 'clr-button-group',
   templateUrl: 'button-group.html',
-  providers: [ButtonInGroupService, ClrDestroyService],
+  providers: [ButtonInGroupService, ClrDestroyService, BUTTON_GROUP_FOCUS_HANDLER_PROVIDER, FocusService],
   hostDirectives: [PopoverHostDirective],
   host: { '[class.btn-group]': 'true' },
 })
 export class ClrButtonGroup implements AfterContentInit {
+  @ViewChild('buttonGroupMenu') buttonGroupMenu: ElementRef;
```
kevinbuhmann, line 42: This should be marked `private` since it's not used outside of this class.

projects/angular/src/button/button-group/button-group.ts
```diff
@@ -111,6 +128,22 @@ export class ClrButtonGroup implements AfterContentInit {
     this.menuButtons = tempInMenuButtons;
   }
 
+  handleFocusOnMenuOpen() {
+    this.toggleService.openChange
+      .pipe(
+        first(event => event === true),
+        delay(0),
+        map(event => {
```
kevinbuhmann, line 136: I would rename these `event` parameters to `open` to make the code more understandable.

projects/angular/src/button/providers/button-group-focus-handler.service.ts
```diff
@@ -0,0 +1,61 @@
...
+import { Linkers } from '../../utils/focus/focusable-item/linkers';
+import { uniqueIdFactory } from '../../utils/id-generator/id-generator.service';
+
+@Injectable()
+export class ButtonGroupFocusHandler {
+  constructor(private focusService: FocusService) {}
+
+  // We need a Change Detector from the related component, so we can update it on Blur
+  // (which is needed because of Edge specific lifecycle mis-behavior)
+  componentCdRef: ChangeDetectorRef;
```
kevinbuhmann, line 20: Since this class is provided at the the component injector, the `ChangeDirectorRef` can be injected directly into this class.

projects/angular/src/button/providers/button-group-focus-handler.service.ts
```diff
@@ -0,0 +1,61 @@
...
+      this.initializeNodes();
+      this.focusFirstItem();
+    }
+  }
+
+  initializeNodes() {
+    if (this.buttonsContainer && this.buttonsContainer.children) {
+      const buttonNodes: HTMLElement[] = Array.from(this.buttonsContainer.children) as HTMLElement[];
+      this.buttons = buttonNodes.map(child => {
+        return {
```
kevinbuhmann, line 41: This can simplified. The `return` keyword isn't needed. Also, I think `child` should be renamed to `buttonElement`?

```typescript
this.buttons = buttonNodes.map(child => ({
  ...properties
}));
```

projects/angular/src/button/providers/button-group-focus-handler.service.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (el) {
+      this.focusService.registerContainer(el);
+      this.initializeNodes();
+      this.focusFirstItem();
+    }
+  }
+
+  initializeNodes() {
+    if (this.buttonsContainer && this.buttonsContainer.children) {
+      const buttonNodes: HTMLElement[] = Array.from(this.buttonsContainer.children) as HTMLElement[];
```
kevinbuhmann, line 39: nit: I think `buttonElements` would be a better name here.

kevinbuhmann (changes requested):
See comments. Most are minor. Two other things:

1. Please squash the commits to one commit (no merge commit).
2. The tests are incomplete.

projects/angular/src/button/providers/button-group-focus-handler.service.ts
kevinbuhmann: Also, the type annotation is redundant here. There is already a type assertion. We don't need both.

projects/angular/src/button/button-group/button-group.ts
```diff
@@ -4,27 +4,42 @@
...
+  ButtonGroupFocusHandler,
+} from '../providers/button-group-focus-handler.service';
 import { ButtonInGroupService } from '../providers/button-in-group.service';
 import { ClrButton } from './button';
 
 @Component({
   selector: 'clr-button-group',
   templateUrl: 'button-group.html',
-  providers: [ButtonInGroupService, ClrDestroyService],
+  providers: [ButtonInGroupService, ClrDestroyService, BUTTON_GROUP_FOCUS_HANDLER_PROVIDER, FocusService],
```
kevinbuhmann, line 37: There is a `FOCUS_SERVICE_PROVIDER` that should be used here.

projects/angular/src/button/providers/button-group-focus-handler.service.ts
```diff
@@ -0,0 +1,55 @@
...
+  set buttonsContainer(el: HTMLElement) {
+    this._buttonsContainer = el;
+    if (el) {
+      this.focusService.registerContainer(el);
+      this.initializeNodes();
+      this.focusFirstItem();
+    }
+  }
+
+  initializeNodes() {
```
kevinbuhmann, line 33: This method could be marked `private`.

projects/angular/src/button/providers/button-group-focus-handler.service.ts
```diff
@@ -0,0 +1,55 @@
...
+        id: uniqueIdFactory(),
+        value: buttonElement,
+        focus: () => buttonElement.focus(),
+        blur: () => buttonElement.blur(),
+      }));
+      Linkers.linkVertical(this.buttons, false);
+    }
+  }
+
+  focusFirstItem(): void {
```
kevinbuhmann, line 46: This method could be marked `private`.

kevinbuhmann (changes requested):

projects/angular/src/button/button-group/button-group.ts
```diff
@@ -111,6 +118,21 @@ export class ClrButtonGroup implements AfterContentInit {
     this.menuButtons = tempInMenuButtons;
   }
 
+  handleFocusOnMenuOpen() {
+    this.toggleService.openChange
+      .pipe(
+        first(isOpened => isOpened === true),
+        delay(0),
+        map(isOpened => {
+          if (isOpened && this.buttonGroupMenu) {
```
kevinbuhmann, line 127: I just realized that checking `isOpened` here is not necessary do to the `first` operator above. Also, I don't think we need to check `this.buttonGroupMenu` either. The `delay` ensures it exists. I don't think it should silently skip this code if there is a bug.

kevinbuhmann: Also, this should a `tap` operator, not a `map`. It's adding a side effect, not mapping the values.

kevinbuhmann (approved):
I paired with Daniel and pushed some fixes/changes.

kevinbuhmann (approved):

Issue #511, opened 2023-02-22
Form validation is triggered on page load


pourfar:
## Describe the bug

Form validations are triggered on page/component load. 

## How to reproduce

app.component.html:

```
<form clrForm>
  <clr-input-container>
    <label>Username</label>
    <input clrInput type="text" [(ngModel)]="username" name="username" required />
    <clr-control-error>Username is required</clr-control-error>
  </clr-input-container>
</form>
```

## Expected behavior

Form validation need to be triggered by form's submit button.

## Versions

**Clarity version:**

v15.0.0

**Framework version:**
Angular 15.1.6

**Device:**

- Type: PC
- OS: Windows 10
- Browser: Edge
- Version: 110.0.1587.50

gerinjacob:
@Jinnie, we are facing this issue in our template driven forms. May I know if the fix is being worked on? Would much appreciate if any tentative date can be provided.

kevinbuhmann:
:tada: This fux is included in version 15.0.3 :tada:

The release is available on:
- [GitHub release](https://github.com/vmware-clarity/ng-clarity/releases/tag/v15.0.3)
- `v15.0.3`

Your **[semantic-release](https://github.com/semantic-release/semantic-release)** bot :package::rocket:

Jinnie:
Released with:

https://github.com/vmware-clarity/ng-clarity/releases/tag/v13.15.0
https://github.com/vmware-clarity/ng-clarity/releases/tag/v15.0.3

PR #548, opened 2023-03-09
fix(forms): fix template-driven forms validation regression

Jinnie:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Forms validation regressed multiple times, after certain Angular update, after which the order of imports of Angular's FormsModule and ClarityModule started to matter for the order of receiving `touched` event and angular validation status event.
Previously, we were trying to augment our internal validity state, based on whether `touched` had arrived or not. With the race condition described above, this was causing different side effects, depending on the fix attempted, including:
- form elements only got validated on the second `blur`
- form elements got validated immediately, when page was opened
Additional problem was that `touched` changes do not have an event handler in the Angular Control object, which makes manual synchronization in the above pattern nearly impossible.

Issue Number: #511 

## What is the new behavior?

Validity is now updated only after the first `blur`  on the element.
We are no longer augmenting the internal validity state, which was making it different from the Angular validity state.
Now both Clarity and Angular validity states are in sync.
The decision. whether to display the element as valid, or not yet validated, based on the `touched` state is not handled in the control state service, but in the Form component itself, which allows us to use the standard Angular data bindings to actively monitor both validation state and `touched` state. With this approach, the elements validity classes get updated regardless of the order in which validity and `touched` state changes arrive.

- [ ] Yes
- [x] No

## Other information

Due to the complex nature of the setup in which these issues are observed, we are aiming for them to be covered by future visual regression testing.
For this purpose, two storybook pages are defined - for reactive and for template-driven setup.
So far, I am unable to create a story with OnPush change detection in storybook, so I have added two more pages to the Clarity demo app, for the OnPush scenarios.

I recommend a manual testing procedure to be run on the following pages in the Dev App:
- Template Driven
- Template Driven (OnPush)
- Reactive
- Reactive (OnPush)

The procedure would require testing that:
- No field on the pages is pre-validated. 
- The field marked as `required` gets validated on the first `blur` event on the element.
- The field marked as `required` does not need a second `blur` event to get validated.

This will need to get back ported.

### DIFF
--- a/projects/angular/src/forms/common/abstract-container.ts
+++ b/projects/angular/src/forms/common/abstract-container.ts
@@ -51,7 +51,7 @@ export abstract class ClrAbstractContainer implements DynamicWrapper, OnDestroy,
 
     return (
       /* Helper Component exist and the state of the form is NONE (not touched) */
-      (!!this.controlHelperComponent && this.state === CONTROL_STATE.NONE) ||
+      (!!this.controlHelperComponent && (!this.touched || this.state === CONTROL_STATE.NONE)) ||
       /* or there is no success component but the state of the form is VALID - show helper information */
       (!!this.controlSuccessComponent === false && this.state === CONTROL_STATE.VALID) ||
       /* or there is no error component but the state of the form is INVALID - show helper information */
@@ -60,11 +60,11 @@ export abstract class ClrAbstractContainer implements DynamicWrapper, OnDestroy,
   }
 
   get showValid(): boolean {
-    return this.state === CONTROL_STATE.VALID && !!this.controlSuccessComponent;
+    return this.touched && this.state === CONTROL_STATE.VALID && !!this.controlSuccessComponent;
   }
 
   get showInvalid(): boolean {
-    return this.state === CONTROL_STATE.INVALID && !!this.controlErrorComponent;
+    return this.touched && this.state === CONTROL_STATE.INVALID && !!this.controlErrorComponent;
   }
 
   constructor(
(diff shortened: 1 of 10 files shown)


Jinnie:
Manual testing environment:
https://stackblitz.com/edit/input-on-blur-issue-onpush-td-13-4-3?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.module.ts

projects/angular/src/forms/common/abstract-container.ts
```diff
@@ -129,4 +129,8 @@ export abstract class ClrAbstractContainer implements DynamicWrapper, OnDestroy,
       });
     }
   }
+
+  private get touched() {
+    return this.control && this.control.touched;
```
kevinbuhmann, line 134: I think this could be `this.control?.touched`.

Jinnie: Won't it return `undefined`?
Technically, it's still falsy, but I prefer the clear boolean result.
And I find `!!this.control?.touched` a bit vague.
I agree it's all about preference.
Is there any technical advantage in using one over the other?

kevinbuhmann: Both would evaluate to `undefined` if `control` is `undefined`. I'm pretty sure these expressions are exactly the same.

kevinbuhmann: And I think returning `undefined` when `control` is `undefined` is correct here. We don't know if the control is touched or not, so `false` could be misleading.

Jinnie: I'll change it as you want, because this may be considered better isolation, but just curious, how a control can get undefined after it was touched? Or how it can get touched before it's created.

kevinbuhmann: I don't think it can. This code already returns `undefined` when the touched state is unknown. I don't think that should be changed at all. I am only suggesting a code simplification, but it's not a requirement.

kevinbuhmann (approved):

williamernest (approved):

projects/angular/src/forms/common/abstract-container.ts
Jinnie: Done!

PR #552, opened 2023-03-09
chore: import scss files directly for chromatic to detect changes

williamernest:
Reviving this change to import our scss files directly. Ideally, chromatic will now detect changes to scss files and trigger a full rebuild on any/all scss changes in the library. This also removes the --externals for package-lock, as this behavior *should* be trigger by chromatic.


kevinbuhmann (approved):

kevinbuhmann (changes requested):
I think the commit message could be more accurate. This doesn't fix live reload because it's not broken. Maybe this:

```
chore: import scss files directly for chromatic to detect changes
```

kevinbuhmann (approved):

PR #562, opened 2023-03-14
fix(dropdown): manage focus correctly when a modal is opened

kevinbuhmann:
Say you have a dropdown item that opens a modal.

This is what happened before this fix:
1. The `clrDropdownItem` is clicked/activated.
2. The modal is opened and focus is moved to the modal title. (This is the custom dropdown item click event handler.)
3. The dropdown menu is closed and focus is moved to the dropdown trigger.
4. The user has to tab or click into the modal to use it.
5. When the modal is closed, focus is not moved back the dropdown trigger.

This is what happens after this fix:
1. The `clrDropdownItem` is clicked/activated.
2. Focus is moved to the dropdown trigger, but the dropdown menu remains open.
3. The modal is opened and focus is moved to the modal title. (This is the custom dropdown item click event handler.)
3. The dropdown menu is closed.
5. The user does NOT have to tab or click into the modal to use it.
6. When the modal is closed, focus is moved back to the dropdown trigger.

This could have been fixed by closing the dropdown menu after custom dropdown item click event handlers have run by removing the `setTimeout` in the `onDropdownItemClick` method; however, that would break event handlers in nested components when `*clrIfOpen` is used since the nested component would be destroyed before it has the chance to handle any events.

To work around that, I changed the code to move focus to the dropdown trigger without closing the dropdown menu. Note that it only does this if the dropdown is going to be closed. If the dropdown is configured to remain open after an item is clicked, focus is not moved.

### DIFF
--- a/.storybook/stories/dropdown/dropdown-item-that-opens-modal.stories.ts
+++ b/.storybook/stories/dropdown/dropdown-item-that-opens-modal.stories.ts
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2016-2023 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { ClrDropdownModule, ClrModalModule } from '@clr/angular';
+import { Parameters } from '@storybook/addons';
+import { Story } from '@storybook/angular';
+
+import { setupStorybook } from '../../helpers/setup-storybook.helpers';
+
+const defaultStory: Story = args => ({
+  template: `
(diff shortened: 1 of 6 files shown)


williamernest (commented):
The nested trigger "open modal" doesn't work if you click the dropdown "open modal" first.

kevinbuhmann:
> The nested trigger "open modal" doesn't work if you click the dropdown "open modal" first.

Good catch. That's fixed now. https://github.com/vmware-clarity/ng-clarity/pull/562/commits/f1326a5a28f218c786ad46c610319c600b583f7c

Jinnie (approved):
Very thorough story research.

williamernest (approved):

PR #563, opened 2023-03-15
fix(button): navigate menu via keyboard without moving focus

dtsanevmw:
## PR Checklist

## PR Type

- Changed the behavior of the keyboard navigation in menu to support all a11y requirements. 
- Keyboard navigation is now handled with having a current item in the menu while the focus is not changed and it's staying on the toggle button as required. This help us resolve all a11y issues which were not resolved with the previous functionality. 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the current behaviour ?

- Currently the service we created was working with another "FocusService" and the navigation was happening trough focusing different items. 

Issue Number: CDE-68

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/button/button-group/button-group.html
+++ b/projects/angular/src/button/button-group/button-group.html
@@ -10,16 +10,21 @@
 <ng-container *ngIf="menuButtons.length > 0">
   <div class="btn-group-overflow open" [ngClass]="menuPosition" #anchor>
     <button
+      #menuToggle
       class="btn dropdown-toggle"
       clrPopoverAnchor
       clrPopoverOpenCloseButton
+      (keydown.arrowup)="openMenu($event, InitialItem.LAST)"
+      (keydown.arrowdown)="openMenu($event, InitialItem.FIRST)"
       [attr.aria-controls]="popoverId"
       [attr.aria-expanded]="open"
       [attr.aria-label]="clrToggleButtonAriaLabel"
+      aria-haspopup="true"
     >
       <cds-icon shape="ellipsis-horizontal" [attr.title]="commonStrings.keys.more"></cds-icon>
     </button>
     <div
+      #menu
       role="menu"
       class="dropdown-menu clr-button-group-menu"
       [id]="popoverId"
--- a/projects/angular/src/button/button-group/button.ts
+++ b/projects/angular/src/button/button-group/button.ts
@@ -6,6 +6,7 @@
 
 import { Component, EventEmitter, Input, Optional, Output, SkipSelf, TemplateRef, ViewChild } from '@angular/core';
 
+import { uniqueIdFactory } from '../../utils/id-generator/id-generator.service';
 import { ClrLoadingState } from '../../utils/loading/loading';
 import { LoadingListener } from '../../utils/loading/loading-listener';
 import { ClrPopoverToggleService } from '../../utils/popover/providers/popover-toggle.service';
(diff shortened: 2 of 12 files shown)


projects/angular/src/button/button-group/button-group.html
```diff
@@ -10,21 +10,21 @@
...
       clrPopoverAnchor
       clrPopoverOpenCloseButton
-      (keydown.arrowup)="openButtonGroupMenu($event, InitialFocus.LAST_ITEM)"
-      (keydown.arrowdown)="openButtonGroupMenu($event, InitialFocus.FIRST_ITEM)"
+      (keydown.arrowup)="openMenu($event, InitialItem.LAST)"
+      (keydown.arrowdown)="openMenu($event, InitialItem.FIRST)"
       [attr.aria-controls]="popoverId"
       [attr.aria-expanded]="open"
-      [attr.aria-label]="commonStrings.keys.rowActions"
+      [attr.aria-label]="clrToggleButtonAriaLabel"
```
kevinbuhmann, line 21: This change seems unrelated to keyboard navigation. I would remove it from here and open a separate PR so that this change is called out as a separate feature in the change log. Otherwise, users won't know we added it unless the dig into the code.

kevinbuhmann (changes requested):
See comments.

After moving the custom `aria-label` feature to a separate PR, this change could be described in the the commit message as `fix(button): navigate menu via keyboard without moving focus`. That way, it's more clear in the changelog what this does.

projects/angular/src/button/providers/button-group-navigation-handler.spec.ts
```diff
@@ -92,32 +88,28 @@ export default function (): void {
       });
 
       it('first child is focused after open', function (this: TestContext) {
-        this.focusHandler.buttonsContainer = this.buttonsContainer;
-        this.focusHandler.menuOpened = true;
-        const firstChild = this.buttonsContainer.children[0];
-        expect(document.activeElement).toBe(firstChild);
+        this.navigationHandler.menuOpened = true;
+        const firstChild = this.menu.children[0];
+        expect(this.navigationHandler.isCurrent(firstChild.id)).toBe(true);
```
kevinbuhmann, line 93: I think this should check for the `clr-focus` class on the actual button element. That way, it's testing real behavior, not just the service implementation details.

projects/angular/src/button/button-group/button.ts
```diff
@@ -148,6 +167,12 @@ export class ClrButton implements LoadingListener {
     this._click.emit(true);
   }
 
+  ngOnInit() {
+    if (this.inMenu && !this.id) {
+      this.id = uniqueIdFactory();
```
kevinbuhmann, line 172: I think this can just be initialized at the `_id` property declaration. That simplifies the code and could prevent an extra DOM update.

kevinbuhmann: It should be okay to just initialize the id of all buttons, not just the ones in the in the menu.

dtsanevmw: I'll have to change some of the tests for button because they are validating if id is '' if it's not set.

kevinbuhmann: Oh. An empty `id` is invalid anyway...

PR #576, opened 2023-03-16
chore: update storybook preview styles to match core

williamernest:
Update the storybook preview styles in ng-clarity to match core (makes diffing components easier).


.storybook/preview.js
```diff
@@ -81,16 +83,19 @@ const themeDecorator = (story, { globals, parameters }) => {
   switch (currentTheme) {
     case THEMES.NG_LIGHT:
       styleElement.textContent = clrUiStyles;
+      document.body.removeAttribute('cds-text');
```
kevinbuhmann, line 86: I think we should add a `cdsTextAttribute` constant like `cdsThemeAttribute` instead of repeating this string.

.storybook/preview.js
```diff
@@ -81,16 +83,19 @@ const themeDecorator = (story, { globals, parameters }) => {
...
       break;
     case THEMES.NG_DARK:
       styleElement.textContent = clrUiDarkStyles;
+      document.body.removeAttribute('cds-text');
       document.body.removeAttribute(cdsThemeAttribute);
       document.body.style.backgroundColor = getClrUiAppBackgroundColor(currentTheme);
       break;
     default:
       styleElement.textContent = `${clrUiStyles}${cdsCoreAndShimStyles.join('')}`;
+      document.body.setAttribute('cds-text', 'body');
```
kevinbuhmann, line 98: Should we update the "@cds/core shim" docs page to include this? Currently, we say the shim only affects color, but this makes our storybook shim usage affect typography as well.

kevinbuhmann (changes requested):

.storybook/preview.js
kevinbuhmann: And our stories currently don't use `cds-text` typography where appropriate, so I am afraid that this may result in some odd typography that we don't actually want.

williamernest: I'll update the shim docs to add cds-text to the body. 

Until we get completely migrated in v17, our components won't use the cds-text everywhere they should be using them but this is a decent enough workaround since the core components inherit a lot of styling from the page.

kevinbuhmann: Okay, that sounds fine to me!

kevinbuhmann (approved):

kevinbuhmann (changes requested):
Thanks for the docs changes!

kevinbuhmann (approved):
I clicked the wrong radio button... This is approved by me.

PR #577, opened 2023-03-17
fix(button): navigate menu via keyboard with focus

dtsanevmw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Keyboard navigation in Button Group Menu is not working.

Issue Number: CDE-68

## What is the new behavior?

Keyboard navigation in Button Group Menu is working with arrows

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/button/button-group/button-group.ts
+++ b/projects/angular/src/button/button-group/button-group.ts
@@ -4,39 +4,58 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { AfterContentInit, Component, ContentChildren, Input, QueryList } from '@angular/core';
-import { takeUntil } from 'rxjs/operators';
+import {
+  AfterContentInit,
+  AfterViewInit,
+  Component,
+  ContentChildren,
+  ElementRef,
+  Input,
+  QueryList,
+  ViewChild,
+} from '@angular/core';
+import { delay, takeUntil } from 'rxjs/operators';
 
 import { ClrDestroyService } from '../../utils/destroy/destroy.service';
+import { FOCUS_SERVICE_PROVIDER } from '../../utils/focus/focus.service';
 import { ClrCommonStringsService } from '../../utils/i18n/common-strings.service';
 import { uniqueIdFactory } from '../../utils/id-generator/id-generator.service';
 import { ClrPopoverPositions } from '../../utils/popover/enums/positions.enum';
 import { ClrPopoverPosition } from '../../utils/popover/interfaces/popover-position.interface';
 import { PopoverHostDirective } from '../../utils/popover/popover-host.directive';
 import { ClrPopoverToggleService } from '../../utils/popover/providers/popover-toggle.service';
+import {
+  BUTTON_GROUP_FOCUS_HANDLER_PROVIDER,
+  ButtonGroupFocusHandler,
+} from '../providers/button-group-focus-handler.service';
+import { InitialFocus } from '../providers/button-group-focus.enum';
 import { ButtonInGroupService } from '../providers/button-in-group.service';
 import { ClrButton } from './button';
 
 @Component({
   selector: 'clr-button-group',
   templateUrl: 'button-group.html',
-  providers: [ButtonInGroupService, ClrDestroyService],
+  providers: [ButtonInGroupService, ClrDestroyService, BUTTON_GROUP_FOCUS_HANDLER_PROVIDER, FOCUS_SERVICE_PROVIDER],
   hostDirectives: [PopoverHostDirective],
   host: { '[class.btn-group]': 'true' },
 })
-export class ClrButtonGroup implements AfterContentInit {
+export class ClrButtonGroup implements AfterContentInit, AfterViewInit {
+  @ViewChild('menuToggle') menuToggle: ElementRef;
+  @ViewChild('menu') menu: ElementRef;
   @ContentChildren(ClrButton) buttons: QueryList<ClrButton>;
 
   // Aria
   @Input('clrToggleButtonAriaLabel') clrToggleButtonAriaLabel: string = this.commonStrings.keys.rowActions;
 
   popoverId = uniqueIdFactory();
+  InitialFocus = InitialFocus;
 
   constructor(
     public buttonGroupNewService: ButtonInGroupService,
     private toggleService: ClrPopoverToggleService,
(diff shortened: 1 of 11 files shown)


projects/angular/src/button/providers/button-group-focus-handler.service.ts
```diff
@@ -0,0 +1,123 @@
...
+      case InitialFocus.LAST_ITEM:
+        this.focusLastItem();
+        break;
+      default:
+        this.focusFirstItem();
+        break;
+    }
+  }
+
+  resetButtonsFocus() {
```
kevinbuhmann, line 59: This method could be marked `private`.

projects/angular/src/button/providers/button-group-focus-handler.service.ts
```diff
@@ -0,0 +1,123 @@
...
+    this.buttons.forEach(button => {
+      button.blur();
```
kevinbuhmann, line 61: Why is this needed? I'm just curious because I don't see what it does.

projects/angular/src/button/providers/button-group-focus-handler.service.ts
```diff
@@ -0,0 +1,123 @@
...
+import { FocusableItem } from '../../utils/focus/focusable-item/focusable-item';
+import { Linkers } from '../../utils/focus/focusable-item/linkers';
+import { InitialFocus } from './button-group-focus.enum';
+
+@Injectable()
+export class ButtonGroupFocusHandler {
+  private initialFocus: InitialFocus = InitialFocus.FIRST_ITEM;
+  private _unlistenFuncs: (() => void)[] = [];
+  private _menuOpened: boolean;
+  private buttons: FocusableItem[];
```
kevinbuhmann, line 20: `private` members are generally put after `public` members of the same kind. So these should go after the `public` properties.

projects/angular/src/button/providers/button-group-focus-handler.service.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  constructor(
+    private focusService: FocusService,
+    private toggleService: ClrPopoverToggleService,
+    private renderer: Renderer2
+  ) {}
+
+  setInitialFocus(initialFocus: InitialFocus) {
+    this.initialFocus = initialFocus;
+  }
```
kevinbuhmann, line 43: This could be removed in favor of just making the property `public`. If there is no logic here, we don't need a set method.

projects/angular/src/button/button-group/button-group.ts
```diff
@@ -4,39 +4,58 @@
...
   templateUrl: 'button-group.html',
-  providers: [ButtonInGroupService, ClrDestroyService],
+  providers: [ButtonInGroupService, ClrDestroyService, BUTTON_GROUP_FOCUS_HANDLER_PROVIDER, FOCUS_SERVICE_PROVIDER],
   hostDirectives: [PopoverHostDirective],
   host: { '[class.btn-group]': 'true' },
 })
-export class ClrButtonGroup implements AfterContentInit {
+export class ClrButtonGroup implements AfterContentInit, AfterViewInit {
+  @ViewChild('menuToggle') menuToggle: ElementRef;
+  @ViewChild('menu') menu: ElementRef;
```
kevinbuhmann, line 44: These properties should be marked `private`. And using `ElementRef<HTMLElement>` should be better for type safety.

projects/angular/src/button/button-group/button-group.ts
```diff
@@ -136,4 +166,14 @@ export class ClrButtonGroup implements AfterContentInit {
 
     this.popoverPosition = (ClrPopoverPositions as Record<string, any>)[this._menuPosition];
   }
+
+  private handleFocusOnMenuOpen() {
+    if (this.menuButtons.length) {
+      this.toggleService.openChange.pipe(delay(0), takeUntil(this.destroy$)).subscribe(isOpened => {
+        this.focusHandler.menuToggle = this.menuToggle?.nativeElement;
+        this.focusHandler.menu = this.menu?.nativeElement;
+        this.focusHandler.menuOpened = isOpened;
```
kevinbuhmann, line 175: This `focusHandler.menuOpened` setter has a side effect, and this code does some implicitly-required setup by setting the `menu` and `menuToggle` elements.

I think a better way would be to explicitly call the `focusHandler.initialize` method if the menu is open and pass the elements. This makes the code more explicit about what is happening and what is required.

The `focusHandler.menuOpened` property is not needed at all after this change.

```typescript
if (isOpened) {
  this.focusHandler.initialize({
    menu: this.menu.nativeElement,
    menuToggle: this.menuToggle.nativeElement,
  });
}
```

projects/angular/src/button/providers/button-group-focus.enum.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2016-2023 VMware, Inc. All Rights Reserved.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+export enum InitialFocus {
+  FIRST_ITEM = 'first', // Focus first item
+  LAST_ITEM = 'last', // Focus last item
```
kevinbuhmann, line 9: Nit: I don't think these comments are necessary. It's obvious what these enum values mean.

kevinbuhmann (changes requested):
Great work! I have a few comments on improving the code, though.

I worked through my suggested changes myself to make sure everything would work, so I went ahead and pushed my changes to a different branch. You can refer to this commit along with my comments. If you want, you can also cherry-pick my fixup commit. It's up to you!

https://github.com/vmware-clarity/ng-clarity/commits/kevin/cde-68-review

projects/angular/src/button/providers/button-group-focus-handler.service.ts
dtsanevmw: We're setting "tabindex" to "-1" on blur, so we can have only one with 0 while navigating trough items and that's just resetting all buttons to 0. Using it two times, first when we are initialising the buttons trough link function because they don't have tabindex set to something by default and after closing the menu so they can go back to 0 again.

kevinbuhmann (approved):

PR #610, opened 2023-03-31
feat(alert): add aria labels for alerts pager

dtsanevmw:
## PR Checklist

- [X] Docs have been added / updated (for bug fixes / features)
- [X] If applicable, have a visual design approval

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

There is only a title on the icon in Alerts Pager.

Issue Number: CDE-132

## What is the new behavior?

Adds an aria-labels to both buttons in the pager which are more specific and uses the current alert page as details to the aria-label. 

- [ ] Yes
- [X] No


projects/angular/src/emphasis/alert/alerts-pager.ts
```diff
@@ -45,6 +45,12 @@ export class ClrAlertsPager implements OnInit, OnDestroy {
     return this.multiAlertService.current;
   }
 
+  get alertsCurrentPageAriaLabel() {
+    return `${this.commonStrings.keys.alertCurrentAlertAriaLabel}, ${this.multiAlertService.current + 1} of ${
+      this.multiAlertService.count
+    }`;
```
kevinbuhmann, line 51: The `alertCurrentAlertAriaLabel` would need to be a template. The "X of Y" needs to be translated properly.

projects/angular/src/emphasis/alert/alerts-pager.html
```diff
@@ -6,14 +6,28 @@
...
 <div class="alerts-pager-control">
   <div class="alerts-page-down">
-    <button class="alerts-pager-button" (click)="pageDown()">
-      <cds-icon shape="angle" direction="left" [attr.title]="commonStrings.keys.previous"></cds-icon>
+    <button
+      class="alerts-pager-button"
+      type="button"
+      (click)="pageDown()"
+      [attr.aria-label]="commonStrings.keys.alertPreviousAlertAriaLabel"
+      aria-describedby="current-alert"
```
kevinbuhmann, line 14: I don't see any element with `current-alert` as its id.

projects/angular/src/emphasis/alert/alerts-pager.html
```diff
@@ -6,14 +6,28 @@
...
+    >
+      <cds-icon shape="angle" direction="left"></cds-icon>
     </button>
   </div>
-  <div class="alerts-pager-text">{{this.multiAlertService.current+1}} / {{this.multiAlertService.count}}</div>
+  <div class="alerts-pager-text" id="alerts-pager" [attr.aria-label]="alertsCurrentPageAriaLabel" aria-current="true">
```
kevinbuhmann, line 19: I don't think `aria-current="true"` is correct here.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current

projects/angular/src/emphasis/alert/alerts-pager.ts
kevinbuhmann: The `X/Y` text is visible on the page, so I am not sure if it's needed in the `aria-label` anyway. cc: @AmyLiNow

dtsanevmw: I added it to be as additional to the buttons. So once one of the two arrow is focused then the screen reader would say "Next Alert, button" then "Current alert, 1 of 4". Otherwise if this element is used without the additional label it would say "1 slash 4" which would not be exactly clear for someone who is using it.

kevinbuhmann: Okay, we should make sure to get a review from Amy on this.

projects/angular/src/emphasis/alert/alerts-pager.html
```diff
@@ -6,14 +6,28 @@
...
+  </div>
   <div class="alerts-page-up">
-    <button class="alerts-pager-button" (click)="pageUp()">
-      <cds-icon shape="angle" direction="right" [attr.title]="commonStrings.keys.next"></cds-icon>
+    <button
+      class="alerts-pager-button"
+      type="button"
+      (click)="pageUp()"
+      [attr.aria-label]="commonStrings.keys.alertNextAlertAriaLabel"
+      aria-describedby="alerts-pager"
```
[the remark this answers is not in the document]
kevinbuhmann: I think this should also be reviewed by Amy. I am not sure if this `aria-describedby` attribute is correct or not.

projects/angular/src/emphasis/alert/alerts-pager.ts
AmyLiNow: Technically there is no WCAG violation but you can argue that it can be interpreted as math because of the slash, so Daniel is right, it is confusing and unclear for those who use screen readers. 

I see this currently `<div id="alerts-pager" class="alerts-pager-text" aria-label="Current alert, 1 of 4"> 1 / 4 </div>`
`aria-label ` only works(some screen readers are starting to be more forgiving and still read aria-labels with not dependent role) when the HTML element/tag has an identity that can be named (<button>, <a>, <header>, <nav>, etc) or an ARIA `role` ( button, link, navigation, region etc). 
In this is just a `div`. 

Recommendations (#2 is ideal):
1. At a minimum, I recommend changing "1/4" to "1 of 4" as it could be read as math division. No aria would be needed here.
2. If possible I recommend a combination of visually hidden text "Current alert, " and visually text "1 of 4". so it would look like this: `<span class="screen-reader-only">Current Alert, </span><span> 1 of 4</span>`. This also will not require aria.

projects/angular/src/emphasis/alert/alerts-pager.html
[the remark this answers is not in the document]
AmyLiNow: For this I recommend aria-label as you are naming the button, not giving additional hints or information. aria-describedby is meant for additional info and also only works on the tab key for some screen readers.

One option is the give the aria-label a more descriptive label, for example if the current alert was 2 of 4, you can do aria-label="Previous alert message, 1 of 4" and aria-label="Next alert message, 3 of 4"

dtsanevmw:
I added more descriptive labels to the buttons as Amy proposed and that removes the need to have the label on "1/4" div.  That div is not focusable and it's just for visual representation. Now the buttons would have enough information for the current alert being: "Next alert message, 1 of 4" and "Previous alert message, 1 of 4"

projects/angular/src/emphasis/alert/alerts-pager.ts
```diff
@@ -45,6 +45,20 @@ export class ClrAlertsPager implements OnInit, OnDestroy {
     return this.multiAlertService.current;
   }
 
+  get previousAlertAriaLabel() {
+    return this.commonStrings.parse(this.commonStrings.keys.alertPreviousAlertAriaLabel, {
+      CURRENT: (this.multiAlertService.current + 1).toString(),
```
kevinbuhmann, line 50: I think you meant to subtract 1 here?

kevinbuhmann (changes requested):

kevinbuhmann (changes requested):
Please add unit tests.

projects/angular/src/emphasis/alert/alerts-pager.spec.ts
```diff
@@ -99,6 +100,55 @@ export default function () {
         expect(this.fixture.componentInstance.currentAlert).toEqual(this.alert);
       });
     });
+
+    describe('View Basics', function () {
+      let fixture: ComponentFixture<TestBasics>;
+      let pagerInstance: ClrAlertsPager;
+      let commonStrings: ClrCommonStringsService;
+      const testNextAriaLabel = (current: number, count: number) => {
```
kevinbuhmann, line 108: A couple things here:

1. This function does test anything, so I would rename to `get...` instead of `test...`.
2. I think this should be a function declaration instead of a variable declaration with a function value. I think a first-class function declared after the specs (out of the way) would make the tests a bit easier to read.

kevinbuhmann (changes requested):

projects/angular/src/emphasis/alert/alerts-pager.spec.ts
```diff
@@ -99,6 +100,55 @@ export default function () {
...
+        pagerInstance = fixture.componentInstance.pagerInstance;
+        commonStrings = pagerInstance.commonStrings;
+      });
+
+      afterEach(function () {
+        fixture.destroy();
+      });
+
+      it('next alert aria-label is correct after page change', function () {
+        expect(pagerInstance.nextAlertAriaLabel).toBe(testNextAriaLabel(2, 4));
```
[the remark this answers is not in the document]
kevinbuhmann: I think the other tests that use common strings instead of string literals for assertions should be changed too. I can do that in a separate PR.

kevinbuhmann, line 142: This is just testing the component properly value, but really we should query the DOM and test the actual `aria-label` attribute.

projects/angular/src/emphasis/alert/alerts-pager.ts
```diff
@@ -45,6 +45,22 @@ export class ClrAlertsPager implements OnInit, OnDestroy {
     return this.multiAlertService.current;
   }
 
+  get previousAlertAriaLabel() {
```
kevinbuhmann, line 48: This should be marked `protected` since it's just for the template.

projects/angular/src/emphasis/alert/alerts-pager.ts
```diff
@@ -45,6 +45,22 @@ export class ClrAlertsPager implements OnInit, OnDestroy {
...
+    const CURRENT = this.currentAlertIndex + 1;
+    return this.commonStrings.parse(this.commonStrings.keys.alertPreviousAlertAriaLabel, {
+      CURRENT: (CURRENT === 1 ? this.multiAlertService.count : CURRENT - 1).toString(),
+      COUNT: this.multiAlertService.count.toString(),
+    });
+  }
+
+  get nextAlertAriaLabel() {
```
kevinbuhmann, line 56: This should be marked `protected` since it's just for the template.

projects/angular/src/emphasis/alert/alerts-pager.spec.ts
```diff
@@ -117,3 +167,9 @@ export class TestInstance {
   @ViewChild(ClrAlertsPager) pagerInstance: ClrAlertsPager;
   currentAlert: ClrAlert;
 }
+@Component({
```
kevinbuhmann, line 170: nit: missing blank line here.

kevinbuhmann (approved):

PR #611, opened 2023-04-03
fix(tree): removes static width and height for tree nodes with checkbox

dtsanevmw:
Original issue: https://github.com/vmware-clarity/ng-clarity/issues/601

## PR Checklist

Please check if your PR fulfills the following requirements:

- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)
- [X] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

The checkbox-wrapper in tree-view have static height and width set which is cutting of the nodes.

Issue Number: CDE-136

## What is the new behavior?

Removes the static height and width so the checkbox-wrapper can get enough space for the checkbox and it's label.

## Does this PR introduce a breaking change?

- [ ] Yes
- [X] No

PR #617, opened 2023-04-03
fix(tree): removes static width and height for tree nodes with checkbox (backport to 13.x)

dtsanevmw:
backport of #611


kevinbuhmann (changes requested):
Please move the `storyApi.add('Variants', ...)` statement into an `if (variants) { ... }` block.

kevinbuhmann (approved):

PR #626, opened 2023-04-04
fix(datagrid): change label around clickable row

williamernest:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

The row selection mode (`clrDgRowSelection`) feature uses a top-level label to wrap around the entire row in order to proxy clicks to the appropriate checkbox. This feature is deprecated, but cannot be removed for legacy reasons. 

Issue Number: Resolves #378 . CDE-5

## What is the new behavior?

The label wrapping the row has been replaced with a div that proxies click events to trigger the appropriate row selection (single vs multi). 

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

PR #629, opened 2023-04-05
fix(datagrid): change label around clickable row (backport 13.x) (#626)

williamernest:
Backport of #626


projects/angular/src/data/datagrid/datagrid-row.html
```diff
@@ -2,12 +2,17 @@
   We need to wrap the #rowContent in label element if we are in rowSelectionMode.
```
jerone, line 2: Looks like these comments are not valid anymore.

PR #631, opened 2023-04-05
chore: add accordion api markdown example

williamernest:
API markdown file example for documenting css variables and class names used throughout components.


kevinbuhmann (approved):
Approved, but we might want to talk about how detailed the CSS class names should be.

PR #640, opened 2023-04-18
feat(wizard): support custom heading levels for titles

dtsanevmw:
## PR Checklist

- [X] Docs have been added / updated (for bug fixes / features)
- [X] If applicable, have a visual design approval

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

The main title was wrapped up around h2 element and even adding a custom one it was still inside h2.

Issue Number: CDE-128

## What is the new behavior?

Added support for custom templates. 

- [] Yes
- [X] No


projects/angular/src/wizard/wizard.html
```diff
@@ -14,15 +14,16 @@
   [clrModalLabelledById]="wizardId"
 >
   <nav class="modal-nav clr-wizard-stepnav-wrapper">
-    <h1 class="clr-wizard-title" id="{{wizardId}}"><ng-content select="clr-wizard-title"></ng-content></h1>
+    <h1 *ngIf="title" class="clr-wizard-title" id="{{wizardId}}">{{title}}</h1>
+    <ng-content select="clr-wizard-title"></ng-content>
```
kevinbuhmann, line 18: I believe this is a breaking change. If a consumer doesn't change their application code, the heading element will be lost.

kevinbuhmann (changes requested):

projects/angular/src/wizard/wizard.html
kevinbuhmann, line 18: I believe this is a breaking change. If a consumer doesn't change their application code, the heading element will be lost.

kevinbuhmann (changes requested):

projects/angular/src/wizard/wizard-page-title.ts
```diff
@@ -4,11 +4,26 @@
...
   selector: '[clrPageTitle]',
 })
 export class ClrWizardPageTitle {
-  constructor(public pageTitleTemplateRef: TemplateRef<any>) {}
+  /**
+   * In order to be a11y compliant [clrPageTitle] should support custom headings
+   * different from the default <h2></h2>. To accomplish that we project the content in
+   * the <h2></h2> only if no other headings are provided.
+   */
+  isIncludingHeadingElements = false;
```
kevinbuhmann, line 18: I would rename this to `includesHeadingElements`.

projects/angular/src/wizard/wizard.html
```diff
@@ -14,15 +14,22 @@
   [clrModalLabelledById]="wizardId"
 >
   <nav class="modal-nav clr-wizard-stepnav-wrapper">
-    <h1 class="clr-wizard-title" id="{{wizardId}}"><ng-content select="clr-wizard-title"></ng-content></h1>
+    <ng-content select="clr-wizard-title"></ng-content>
```
kevinbuhmann, line 17: Can we move the `clr-wizard-title` CSS class to a wrapper span/div so that users don't have to include that class on custom heading elements?

kevinbuhmann: Same thing for the `id`.

kevinbuhmann (changes requested):
If an `h2` (the default) is projected for `clrPageTitle`, this happens:

image

The stepnav shouldn't have large text like this. Also, there are some heading alignment changes in the visual snapshots.

  I took a stab at this, and I think I found a simpler solution that doesn’t introduce any visual changes. Please take a look at 40cc8e8091fedd60f3b0ca33cd017d13bbe9e04a. If you like my solution, I think we should take it instead.

projects/angular/src/wizard/wizard.html
```diff
@@ -14,15 +14,24 @@
   [clrModalLabelledById]="wizardId"
 >
   <nav class="modal-nav clr-wizard-stepnav-wrapper">
-    <h1 class="clr-wizard-title" id="{{wizardId}}"><ng-content select="clr-wizard-title"></ng-content></h1>
+    <div class="clr-wizard-title-wrapper">
+      <ng-content select="clr-wizard-title"></ng-content>
+    </div>
```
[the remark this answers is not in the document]
dtsanevmw: I started with the idea to leave the styling to whoever uses these custom styles. But I like the idea to have consistent styles for all headings and to be used just for a11y. I added couple of css rules that should handle these cases.

projects/angular/src/wizard/wizard.html
```diff
@@ -14,15 +14,18 @@
...
     <clr-wizard-stepnav></clr-wizard-stepnav>
   </nav>
 
-  <h2 class="modal-title">
-    <span tabindex="-1" #wizardTitle class="modal-title-text">
-      <ng-template [ngTemplateOutlet]="navService.currentPageTitle"></ng-template>
-    </span>
-
+  <div class="modal-title">
```
kevinbuhmann, line 23: I think this change introduces visual changes. I would keep the heading element here for this change. There should not be any visual changes here. This PR should be scoped to just the custom heading levels. I have WIP on the wizard redesign that moves that header actions.

projects/angular/src/wizard/wizard.html
```diff
@@ -14,15 +14,18 @@
   [clrModalLabelledById]="wizardId"
 >
   <nav class="modal-nav clr-wizard-stepnav-wrapper">
-    <h1 class="clr-wizard-title" id="{{wizardId}}"><ng-content select="clr-wizard-title"></ng-content></h1>
+    <div class="clr-wizard-title" role="heading" [attr.aria-level]="wizardTitle.headingLevel" id="{{wizardId}}">
```
kevinbuhmann, line 17: This causes an incorrect visual change as well.

projects/angular/src/wizard/wizard-title.ts
```diff
@@ -4,12 +4,11 @@
...
 
 @Directive({
-  selector: 'clr-wizard-title, clr-wizard-pagetitle',
+  selector: 'clr-wizard-title',
 })
-export class ClrWizardCustomTags {
-  // No behavior
-  // The only purpose is to "declare" the tag in Angular
+export class ClrWizardTitle {
+  @Input('clrHeadingLevel') headingLevel: '1' | '2' | '3' | '4' | '5' | '6' = '1';
```
kevinbuhmann, line 13: This should support numbers as well. I would add a `heading-level.ts` file with the following:

```typescript
export type HeadingLevel = 1 | 2 | 3 | 4 | 5 | 6 | '1' | '2' | '3' | '4' | '5' | '6';
```

kevinbuhmann (changes requested):
See comments. This PR should not have any visual changes. Also, the storybook should be updated and unit tests are needed.

See ea7d71450b8dc9988cfd2d894cb0fec6be6c8d37 for a working implementation with these changes and without visual changes.

projects/angular/src/wizard/wizard.html
```diff
@@ -14,15 +14,16 @@
...
+    </div>
     <clr-wizard-stepnav></clr-wizard-stepnav>
   </nav>
 
-  <h2 class="modal-title">
+  <div class="modal-title" role="heading" [attr.aria-level]="navService.currentPage.pageTitle.headingLevel || 2">
     <span tabindex="-1" #pageTitle class="modal-title-text">
       <ng-template [ngTemplateOutlet]="navService.currentPageTitle"></ng-template>
     </span>
-
```
kevinbuhmann, line 25: Please put this blank line back.

projects/angular/src/wizard/_wizard.clarity.scss
```diff
@@ -366,6 +367,16 @@
       display: block;
     }
 
+    .clr-wizard-page-title {
```
kevinbuhmann, line 370: These styles are not used or needed.

kevinbuhmann (changes requested):

projects/angular/src/wizard/index.ts
```diff
@@ -10,9 +10,11 @@ export * from './wizard-stepnav';
 export * from './wizard-stepnav-item';
 export * from './wizard-button';
 export * from './wizard-header-action';
-export * from './wizard-custom-tags';
```
kevinbuhmann, line 13: This file should be deleted.

kevinbuhmann (approved):

PR #645, opened 2023-04-20
fix(stack-view): remove static height in modals

mindarelus:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, there is static height 55vh set to the stack-view when in modal.

Issue Number: CDE-10

## What is the new behavior?
Now there is no static height set to the stack-view when in modal.

- [ ] Yes
- [x] No

If some application relies on the 55vh height set to the stack-view in modal, it can be set directly on the modal.

### DIFF
--- a/projects/angular/src/data/stack-view/_stack-view.clarity.scss
+++ b/projects/angular/src/data/stack-view/_stack-view.clarity.scss
@@ -302,7 +302,6 @@
       Styles specific to stack views used in a modal
     */
     .modal & {
-      height: 55vh;
       margin-bottom: 0;
     }
   }


kevinbuhmann (commented):
I don't think this should be considered a breaking change. I mean, you're right that it technically could be if we use a very broad definition of breaking change. But with that kind of definition, almost any bug fix could be considered a breaking change. I don't think this needs to be held back until v16.

I think it's unlikely that any application would rely on this anyway.

kevinbuhmann (approved):

mindarelus:
Removed the height: 55vh; from the stack-view's modal style rules, because that is not the desired behavior. If an application happens to rely on this value, it can be set directly on the modal.

PR #673, opened 2023-05-10
feat(wizard): add custom aria-label for stepnav section

dtsanevmw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently the stepnav wrapper is using <nav> element and there is no aria-label or possibility to set one for it.

Issue Number: CDE-153

## What is the new behavior?

The stepnav wrapper is now using a <div> element with role="region" and an aria-label. Which can also be set trough the clrWizardStepnavAriaLabel input.

- [ ] Yes
- [X] No

### DIFF
--- a/.storybook/stories/wizard/wizard.stories.ts
+++ b/.storybook/stories/wizard/wizard.stories.ts
@@ -4,7 +4,7 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-import { ClrWizard, ClrWizardModule } from '@clr/angular';
+import { ClrWizard, ClrWizardModule, commonStringsDefault } from '@clr/angular';
 import { action } from '@storybook/addon-actions';
 import { Parameters } from '@storybook/addons';
 import { Story } from '@storybook/angular';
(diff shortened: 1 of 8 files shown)


projects/angular/src/wizard/wizard.ts
```diff
@@ -47,6 +48,11 @@ import { ClrWizardTitle } from './wizard-title';
   },
 })
 export class ClrWizard implements OnDestroy, AfterContentInit, DoCheck {
+  /**
+   * Set the aria-label for the stepnav section of the wizard. Set using `[clrWizardStepnavAriaLabel]` input.
+   */
+  @Input('clrWizardStepnavAriaLabel') stepnavAriaLabel = commonStringsDefault.wizardStepnavAriaLabel;
```
kevinbuhmann, line 54: We need to inject the common strings service and the get the value from there so that it is translated if needed.

.storybook/stories/wizard/wizard.stories.ts
```diff
@@ -61,6 +62,7 @@ const defaultParameters: Parameters = {
     clrWizardPreventDefaultNext: { defaultValue: false },
     clrWizardPreventDefaultCancel: { defaultValue: false },
     clrWizardPreventModalAnimation: { defaultValue: false },
+    clrWizardStepnavAriaLabel: { defaultValue: 'Step navigation', control: { type: 'text' } },
```
kevinbuhmann, line 65: 1. I think `control: { type: 'text' }` can be removed since it can be inferred.
2. I think we should use `commonStringsDefault.wizardStepnavAriaLabel` here like in other stories so it's updated for us if the default value is changed.

kevinbuhmann (changes requested):

kevinbuhmann (approved):

PR #679, opened 2023-05-12
fix(combobox): prevent flashing on option selected

dtsanevmw:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The combobox popover is flashing on option selection due to opacity change of the popover content.  Opacity was set to 0 when the "realign" is called which was not needed in this case. It is needed only on initial creation of the popover when the position of it is top: 0, left:0 and then when it's in the right position it would appear.

Issue Number: CDE-11

## What is the new behavior?

No longer flashing.

- [ ] Yes
- [X] No


Jinnie (approved):
Thanks for checking this case and providing the fix.

PR #704, opened 2023-05-23
chore: sort class members

kevinbuhmann:


williamernest (approved):
Thanks for driving this. 
giphy

PR #707, opened 2023-05-24
chore: add button css docs

kevinbuhmann:
CDE-256


projects/angular/src/button/API.md
```diff
@@ -0,0 +1,160 @@
...
+| --clr-btn-inverse-checked-bg-color              | Changes the checked checkbox inverse button background color.                           |
+| --clr-btn-icon-disabled-color                   | Changes the icon button `disabled` text color.                                          |
+| --clr-btn-group-focus-outline                   | Changes the `:focus` outline color of a checkbox or radio inside a button group button. |
+
+## Class names
+
+| Class                    | Description               |
+| ------------------------ | ------------------------- |
+| .btn                     | button                    |
+| .btn.btn-info            | info button               |
```
williamernest, line 146: Can you change these to be class names instead of selectors? 

The description should be in a similar format to the above table.

williamernest (changes requested):

projects/angular/src/button/API.md
kevinbuhmann: Just the class name is not good documentation. The classes must be used in a specific structure.

williamernest: It's important that these remain consistent with other components, and listing the selectors in this way will lead to longer and longer class names which will be complicated for users to parse/read and even more difficult to maintain.

williamernest (approved):

Jinnie (changes requested):
Looks great, just a few wording details for consistency.
I'll post more info in the dev channel.

Jinnie (approved):

PR #712, opened 2023-05-26
fix(button): use proper warning colors

valentin-mladenov:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Warning button style is set to danger (red) instead of warning (yellow) style.

## What is the new behavior?

Warning button style is set to warning (yellow) style.

- [ ] Yes
- [ ] No

### DIFF
--- a/projects/angular/src/button/_properties.buttons.scss
+++ b/projects/angular/src/button/_properties.buttons.scss
@@ -137,28 +137,28 @@
 
       // Warning button colors
       --clr-btn-warning-color: var(--clr-color-neutral-0);
-      --clr-btn-warning-bg-color: var(--clr-color-danger-700);
-      --clr-btn-warning-border-color: var(--clr-btn-danger-bg-color);
-      --clr-btn-warning-hover-bg-color: var(--clr-color-danger-800);
-      --clr-btn-warning-hover-color: var(--clr-btn-danger-color);
-      --clr-btn-warning-box-shadow-color: var(--clr-color-danger-900);
+      --clr-btn-warning-bg-color: var(--clr-color-warning-700);
+      --clr-btn-warning-border-color: var(--clr-btn-warning-bg-color);
+      --clr-btn-warning-hover-bg-color: var(--clr-color-warning-800);
+      --clr-btn-warning-hover-color: var(--clr-btn-warning-color);
+      --clr-btn-warning-box-shadow-color: var(--clr-color-warning-900);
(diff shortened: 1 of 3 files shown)


projects/angular/src/button/_properties.buttons.scss
```diff
@@ -137,28 +137,28 @@
 
       // Warning button colors
       --clr-btn-warning-color: var(--clr-color-neutral-0);
```
dtsanevmw, line 139:
```suggestion
      --clr-btn-warning-color: var(--clr-btn-warning-color);
```

dtsanevmw (changes requested):
I added one change for text color there. Also can you please compare the dark theme colors to the one we have in core - https://storybook.core.clarity.design/?path=/story/components-button--page&globals=theme:dark it seems that there are different from these here.

projects/angular/src/button/_properties.buttons.scss
valentin-mladenov: @dtsanevmw I don't think `--clr-btn-warning-color` referencing itself would help. 
Also all of the button colors are using `--clr-color-neutral-0` for that variable.
What is your logic behind the suggestion?

dtsanevmw: @valentin-mladenov Sorry I reviewed in the demo app and there it was overridden by the shim.cds-core. However all the button seems to use neutral-0 except the warning button, that's what I see in core. I suspect it's because of the contrast ratio for a11y. The right one seems to be that `--clr-btn-warning-color: var(--cds-global-color-construction-900);` which is equivalent to `var(--clr-color-neutral-900)`. 
Screenshot 2023-05-29 at 10 46 39

valentin-mladenov: Changed to use neutral 900 in commit e1c6bd5

valentin-mladenov: After some more investigation the color `--cds-global-color-construction-900` used for warning in core, doesn't look correct. When switching to high contrast theme the text blends with the background of the button.
https://storybook.core.clarity.design/?path=/docs/components-button--page&globals=theme:high-contrast

@colinreedmiller which color shall we use `--clr-color-neutral-0` or `--clr-color-neutral-900`?

valentin-mladenov:
Corrected in token migration PR #829

PR #726, opened 2023-05-31
chore: add forms css docs

dtsanevmw:
CDE-265


projects/angular/src/forms/STYLES.MD
```diff
@@ -0,0 +1,58 @@
...
+| --clr-forms-radio-disabled-shadow               | Changes the shadow of disabled radio.                              |
+| --clr-forms-radio-focused-shadow                | Changes the shadow of focused radio.                               |
+| --clr-forms-range-progress-fill-color           | Changes the fill color of range progress.                          |
+| --clr-forms-range-track-color                   | Changes the fill color of range track.                             |
+
+## Class names
+
+| Class name        | Description                                             |
+| ----------------- | ------------------------------------------------------- |
+| clr-form          | Styles an element.                                      |
```
Jinnie, line 56: Main form element.

projects/angular/src/forms/STYLES.MD
```diff
@@ -0,0 +1,58 @@
...
+| clr-form-group    | Styles an element with padding.                         |
```
Jinnie, line 57: Form group identifier.

projects/angular/src/forms/STYLES.MD
```diff
@@ -0,0 +1,58 @@
+# FORMS - Style API
+
+## CSS Variables
+
+| CSS Variable Name                               | Description                                                        |
+| ----------------------------------------------- | ------------------------------------------------------------------ |
+| --clr-form-disabled-background-color            | Changes the default background color for disabled forms.           |
```
Jinnie, line 7: As we agreed, let's remove the "Changes the..." boilerplate prefix.
Please, apply everywhere.

Jinnie (changes requested):
Some comments on the agreed wording - let's try to keep it simple and short.
Also, for the class names, let us make them describing the role/function of the element, not so much the behavior. As the behavior may change. With the exception, of course, to the classes that especially identify a behavior, like "disabled" etc.

Jinnie (approved):
Great! Thanks

PR #727, opened 2023-05-31
chore: add modal css docs

dtsanevmw:
CDE-272


projects/angular/src/modal/STYLES.md
```diff
@@ -0,0 +1,33 @@
+# MODAL - Style API
+
+## CSS Custom Properties
+
+| CSS Variable Name                    | Description                    |
```
Jinnie, line 5: Maybe we should change it here too?

Jinnie (approved):
Awesome, thanks!!

PR #760, opened 2023-06-08
fix(accordion): style disabled panel button correctly

kevinbuhmann:
CD-247


dtsanevmw:
@kevinbuhmann Shouldn't we add the same color to the caret icon in disabled state?

kevinbuhmann:
> Shouldn't we add the same color to the caret icon in disabled state?

The caret icon color isn't changed in Clarity Core for disabled accordion panels, so I don't think so.

colinreedmiller:
> @kevinbuhmann Shouldn't we add the same color to the caret icon in disabled state?

You can get a decision on such questions quickly by asking design to specify the correct color. Please add a member of the design team as a reviewer on this ticket.

Jinnie (approved):

williamernest:
Matching the Core styles is sufficient for this task and keeps the scope aligned with the intent. If we decide that the disabled styling should also apply to the icon, we can file a followup ticket to address that in both libraries simultaneously.

kevinbuhmann:
I still confirmation from design on my text color choices for ng-light and ng-dark.

Jinnie:
As Will said the parity with Core should be enough. I don't want this task to creep into the next iteration. If design has some comments, we can add them iteratively.
I will merge this so we can finalize the iteration.

PR #775, opened 2023-06-17
feat(popover): use scroll position in position calculation and visible event added

dtsanevmw:
## PR Checklist

## PR Type

- [] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently if you open a popover it's being rendered at the top left corner with opacity 0 which just makes it invisible but it's available for interactions. So when you click the action button in datagrid and you have available scroll the page gets scrolled to the top due to focusing of the first item in that popover which is positioned in the top corner.

Issue Number: CDE-150

## What is the new behavior?

We are using the scrolled value in the calculations because while the content was scrolled to the top the wrong positioned popover was not noticed. Using an event which emits once the popover is visible fixes a lot of possible and current issues because of most components are awaiting popover open event which is not when the popover is really ready for interactions.  
Initially I fixed the issue with using 'visibility: hidden' instead of opacity to hide the popover but that creates a lot of other issues both in tests and other places which in perfect scenario should be fixed but in the current moment the task gets out of scope a lot. 

- [ ] Yes
- [X] No

### DIFF
--- a/projects/angular/src/data/datagrid/datagrid-action-overflow.ts
+++ b/projects/angular/src/data/datagrid/datagrid-action-overflow.ts
@@ -94,7 +94,9 @@ export class ClrDatagridActionOverflow implements OnDestroy {
     this.subscriptions.push(
       this.smartToggleService.openChange.subscribe(openState => {
         this.open = openState;
-        if (openState) {
+      }),
+      this.smartToggleService.popoverVisible.subscribe(visible => {
+        if (visible) {
           this.initializeFocus();
         }
       })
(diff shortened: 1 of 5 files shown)


mindarelus (approved):

projects/angular/src/data/datagrid/datagrid-action-overflow.ts
```diff
@@ -128,16 +130,14 @@ export class ClrDatagridActionOverflow implements OnDestroy {
   private initializeFocus(): void {
     if (isPlatformBrowser(this.platformId)) {
       this.zone.runOutsideAngular(() => {
-        setTimeout(() => {
-          const buttons = Array.from(document.querySelectorAll<HTMLButtonElement>('button.action-item'));
```
Jinnie, line 132: Love to see this go away

Jinnie: Now we should be able to remove the wrapping `runOutsideAngular` too

Jinnie (approved):

Issue #355, opened 2022-10-20
Datagrid filter overlay disappear while typing.


Qumhieh:
![Datagrid filter overlay](https://user-images.githubusercontent.com/46529617/196920773-c31cd9ce-3340-4130-912b-95a7ffcf90d4.gif)
**This is a bug report for the `@clr` Angular or UI versions of the design system.
For the web-component implementation of Clarity (`@cds`), visit [vmware-clarity/core](https://github.com/vmware-clarity/core).**

## Describe the bug
Datagrid filter overlay keeps disappearing while typing, see the attached gif. I noticed that it's only happening when the datagrid located at the bottom of the page, and the screen height decreasing while typing. (which is something related to page, or container scrolling event)

## How to reproduce

It can be reproduced in clarity datagrid documentation, built-in filter section https://clarity.design/documentation/datagrid/built-in-filters

1.  Visit the link.
2.  scroll all the way down, to the datagrid and the page navigation buttons "custom filtering", and "pagination". (Should fully scroll down)
3.  click on the filter icon for Pokemon column, and start typing.
4.  notice that the overlay disappear with each typing.

## Expected behavior

Datagrid filter should not disappear while typing.

## Versions

**Clarity version:**

- [x] v12.x
- [x] v13.x
- [ ] Other:

**Framework:**

- [x] Angular
- [ ] React
- [ ] Vue
- [ ] Other:

**Framework version:**
_ie: Angular 11_

**Device:**

- Type: MacBook
- OS: iOS
- Browser: chrome.
- Version: 106.0.5249.119

## Additional notes

From the source code i found that the filter overlay should disappear on datagrid scrolling, but not window, or parent container. I investigated the issue and i'm pretty sure it's related to the scrolling event somehow.

Jinnie:
We did some experiments (thanks @bbogdanov) with this.
So far, it seems we don't have satisfactory solution or workaround that does not involve more serious architectural changes to the popovers component/s in Clarity.

The closest we got was to prevent closing the overlay when scroll is caused by screen resize, but then the popover gets displaced from the component that opened it.

PR #782, opened 2023-06-22
fix(datagrid): keep filter overlay open while typing

dtsanevmw:
Removing the scroll listener right before filter change resolves the issue. Also realigning it with the correct position when the height is changed.

## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When typing in a filter overlay and the items in a datagrid get filtered in lower number, then the height of datagrid changes and depending of the position in the page it triggers a scroll event which closes the filter.

Issue Number: CDE-8, #355 

## What is the new behavior?

No longer closes on typing.

- [ ] Yes
- [X] No


mindarelus (approved):

projects/angular/src/data/datagrid/datagrid-column.ts
```diff
@@ -307,6 +314,10 @@ export class ClrDatagridColumn<T = any>
     return this.wrappedInjector.get(WrappedColumn, this.vcr).columnView;
   }
 
+  get datagridElementRef() {
+    return this.elementRef?.nativeElement?.closest('clr-datagrid');
+  }
```
kevinbuhmann, line 319: 1. This returns the element not an `ElementRef`, so please drop `Ref` from the property name.
2. This will run the query every time the property is called. To prevent it from querying every time a filter is changed, you can cache the value in the `listenForFilterChanges`. And since it's only used there, I would just inline the query.
3. Further, it would still run this query for every column. That is unnecessary. I would instead inject the `ClrDatagrid` component, make the `ClrDatagrid#el` property `public`, and use that instead.

kevinbuhmann (changes requested):

kevinbuhmann (changes requested):

kevinbuhmann (commented):
I think `fix(datagrid): keep filter overlay open while typing` would be a better PR title/commit message.

projects/angular/src/data/datagrid/datagrid-column.ts
dtsanevmw: I added a reference in the columns service as injecting it creates a lot of dependency issues on tests and other places.  Maybe it's better to have it in separate datagrid service, tell me what you think.

kevinbuhmann (changes requested):
I looked into this a bit more. I think this solution works, but it's "hacky" and likely to be difficult to maintain.

The root issue that the height of the datagrid is sometimes changed when filtering. It's inconsistent because it happens on the page linked in #355 but not in our Storybook. We should fix that bug instead so that the scroll doesn't happen in the first place instead of working around one bug to fix another.

projects/angular/src/data/datagrid/all.spec.ts
```diff
@@ -72,6 +73,7 @@ describe('Datagrid', function () {
     DisplayModeServiceSpecs();
     TableSizeServiceSpec();
     ColumnResizerServiceSpecs();
+    ColumnsServiceSpec();
```
kevinbuhmann, line 76: Please pull this out to its own `chore` PR so we can get it merged now while we work on fixing the underlying scroll issue.

projects/angular/src/data/datagrid/datagrid-column.ts
kevinbuhmann: Yeah, circular dependencies are annoying.

dtsanevmw:
I got back to that PR and basically we have 2 states of the datagrid. If we have pagination and page size ( items per page ) then we automatically set a height for the whole datagrid and that issue is not appearing. Although when there is no pagination and page size then we don't set that and basically the height of the datagrid can change based on the items inside. So when we don't have a page size we listen for filter changes and stop listening for scroll events while aligning the popover to it's new position.

dtsanevmw:
Will wait the popover refactor https://github.com/vmware-clarity/ng-clarity/pull/1435 to be merged

valentin-mladenov:
Is done with #2062

PR #808, opened 2023-07-11
feat: ship `STYLES.md` content in `@clr/ui` package

kevinbuhmann:
This will allow the website to read and display the style docs.


dtsanevmw (approved):

scripts/compile-style-docs.js
```diff
@@ -0,0 +1,19 @@
...
+const styleDocs = glob
+  .sync('./projects/**/STYLES.md')
+  .map(filePath => fs.readFileSync(filePath).toString())
+  .join('\n')
+  .split(/^(?=# )/m) // split on level-one headings without capturing the leading hash symbol
+  .sort() // sort the components alphabetically
+  .map(content => content.trim())
+  .join('\n\n');
+
+fs.writeFileSync('./dist/clr-ui/STYLES.md', `${styleDocs}\n`);
```
williamernest, line 19: What's the value of concatenating all the docs together rather than creating a docs/ folder and copying over current paths (or using a defined path per component (`./dist/clr-ui/docs/accordon/SYLES.md`) ?

kevinbuhmann: This is really easy to process on the website side. Some of the individual `STYLE.md` have one component, some have multiple components. I don't really want to worry about splitting it up here.

kevinbuhmann: My idea was to do as little as possible on `ng-clarity` side just to get the content in the package. I did do a bit more more than absolutely required by sorting the components, but I did that so the resulting file was well-ordered and more likely to be deterministic.

kevinbuhmann: I think it would be worse to have extra logic here to split things up AND extra logic on the website side to process multiple files.

williamernest: Discussed offline. Going to move forward with this approach.

PR #811, opened 2023-07-13
feat(card): add tokens and update shim

dtsanevmw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-660

- [ ] Yes
- [ ] No


projects/angular/src/layout/_card.clarity.scss
```diff
@@ -169,27 +188,72 @@
...
-      card-variables.$clr-card-title-color,
+      padding,
+      clr-card-footer-padding,
+      card-variables.$card-children-padding-vertical card-variables.$card-children-padding-horizontal,
       variables.$clr-use-custom-properties
     );
+  }
+
+  h3 {
+    &.card-header {
```
kevinbuhmann, line 210: This could be simplified to just `h3.card-header` since there isn't other nested rules. Same for the `h4` one.

projects/angular/src/layout/_card.clarity.scss
```diff
@@ -169,27 +188,72 @@
...
+
+  .card-header,
+  .card-title {
     @include mixins.css-var(
       font-weight,
       clr-card-title-font-weight,
       card-variables.$clr-card-title-font-weight,
       variables.$clr-use-custom-properties
     );
-    font-size: card-variables.$clr-card-title-font-size;
```
kevinbuhmann, line 192: Should be a default `font-size` in case the `.card-title` is not an `h3` or `h4`? The code examples on the website all have the `.card-title` as an `h3` or `h4`, but there are some non-examples that use a `div` instead. I don't know if the card title should alway be a heading element or not, but this seems like a breaking change when the shim is not applied that we should not ship in a non-major release.

Partick (changes requested):
So far the type and border colors in light theme appear slightly off and the colors in dark theme appear off significantly. 

All text should be set to cds-alias-typography-color-400, which appears dark grey in the light theme (construction 900) and light grey (construction 100) in the dark theme. 

The border should be set to cds-alias-object-border which maps to Construction 500 in dark theme and Construction 200 in light theme. Neither appear to be right. Please ensure you are mapping the color attributes to the correct aliases. 
image
image

projects/angular/src/layout/_card.clarity.scss
dtsanevmw: Yes, you are right. We don't want to block them from using different elements.

projects/angular/src/layout/_card.clarity.scss
```diff
@@ -169,27 +188,69 @@
...
+      variables.$clr-use-custom-properties
+    );
+    @include mixins.css-var(
+      color,
+      clr-card-header-title-color,
+      card-variables.$clr-card-title-color,
+      variables.$clr-use-custom-properties
+    );
+  }
+  h4.card-title,
```
kevinbuhmann, line 230: nit: a blank line here would make this more readable

projects/angular/src/layout/_card.clarity.scss
```diff
@@ -169,27 +188,69 @@
...
+      padding,
+      clr-card-footer-padding,
+      card-variables.$card-children-padding-vertical card-variables.$card-children-padding-horizontal,
+      variables.$clr-use-custom-properties
+    );
   }
 
-  .card-header,
+  h3.card-header,
+  .card-header {
```
kevinbuhmann, line 210: Wouldn't just `.card-header` do exactly the same thing?

kevinbuhmann: Same thing for `.card-title`.

dtsanevmw: If it's with `.card-header` only and in a h3, we don't select it explicitly and it being overwritten due to the lower priority.

kevinbuhmann: Overwritten by the `h3` typography styles? Okay, that makes sense.

projects/angular/src/emphasis/alert/_alert.clarity.scss
```diff
@@ -489,9 +489,18 @@
...
+        clr-card-alert-margin-top,
+        variables.$clr_baselineRem_0_25,
+        variables.$clr-use-custom-properties
+      );
+      @include mixins.css-var(
+        margin-bottom,
+        clr-card-alert-margin-bottom,
+        variables.$clr_baselineRem_0_25,
+        variables.$clr-use-custom-properties
+      );
```
Jinnie, line 503: Where did the horizontal margin go?
Now we have separate top and bottom and no left-right

tests/snapshots/card/card--variants-core-dark.png
Jinnie, line 1: We have black text on dark grey background. It does not look right.

projects/angular/src/layout/_card.clarity.scss
```diff
@@ -14,10 +14,12 @@
 @include mixins.exports('card.clarity') {
   %card-border-appearance {
     // IE/OldEdge
```
Jinnie, line 16: Maybe it's not the right time now, but all these are marked IE/OldEdge and probably we can just remove them.

projects/angular/src/progress/progress-bars/_progress-bars.clarity.scss
```diff
@@ -561,6 +561,10 @@
       // by default, positioned at the topmost of its card-block container
       margin: 0;
       margin-top: -1 * card-variables.$card-children-padding-vertical;
+      @if variables.$clr-use-custom-properties == true {
```
Jinnie, line 564: I think I am also doing something similar in the combobox. I was thinking what will happen if we remove the "exports" mixin, as we talked that we may after moving to @use. Won't all these @ifs we put stop working?

Jinnie (commented):
Just a note - the "lists" PR may introduce changes to Cards screenshots, because of the "cards with lists" stories. 
Maybe this PR should go after the lists, to include the above story.

projects/angular/src/progress/progress-bars/_progress-bars.clarity.scss
Jinnie: Pls ignore for the time being.

projects/angular/src/layout/_card.clarity.scss
Jinnie: Yes, it's not the right time for this. Resolving

projects/angular/src/layout/_card.clarity.scss
```diff
@@ -170,27 +187,78 @@
...
   .card-footer {
-    padding: card-variables.$card-children-padding-vertical card-variables.$card-children-padding-horizontal;
+    display: flex;
+    flex-direction: row;
+    @include mixins.css-var(
+      gap,
+      clr-card-footer-gap,
+      variables.$clr_baselineRem_0_5,
+      variables.$clr-use-custom-properties
+    );
```
valentin-mladenov, line 207: Wouldn't this change affect ng theme? Isn't it better to use `button-variables.$clr-btn-horizontal-margin` for fallback?
```suggestion
    @include mixins.css-var(
      gap,
      clr-card-footer-gap,
      button-variables.$clr-btn-horizontal-margin,
      variables.$clr-use-custom-properties
    );
```

valentin-mladenov (changes requested):
Check my comments.

projects/angular/src/layout/_card.clarity.scss
dtsanevmw: I copied the value which `clr-btn-horizontal-margin` is using so it basically the same and we don't have to import the button-variables for this specific value. The footer is not specifically made for buttons only so It's confusing to use the buttons margin.

valentin-mladenov (approved):
LGTM

PR #814, opened 2023-07-14
fix(tooltip): allow user to interact with tooltip content using mouse

elesueur:
The previous change to introduce a fix for VPAT-617 meant the user was unable to copy text out of a tooltip because the ::after content overlays the tooltip content.

This change adds a -1 z-index to the ::after so that it is underneath which exposes the tooltip content div to allow text to be selected and copied.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

User hovers over tooltip trigger, tooltip displays, user then mouses over the tooltip content, but cannot select any text due to the overlaid ::after content.

Issue Number: N/A

## What is the new behavior?

User can now select text from tooltip and copy it to their clipboard.

- [ ] Yes
- [x] No

before:

before

after:

after

### DIFF
--- a/projects/angular/src/popover/tooltip/_tooltips.clarity.scss
+++ b/projects/angular/src/popover/tooltip/_tooltips.clarity.scss
@@ -136,6 +136,7 @@
       white-space: normal;
       z-index: map.get(variables.$clr-layers, tooltips);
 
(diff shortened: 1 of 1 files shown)


kevinbuhmann (approved):
Thank you!

PR #826, opened 2023-07-20
chore: deploy pr preview whether or not build succeeds if docs generated

kevinbuhmann:
This change is helpful in cases where only the visual regression tests fail. It may be helpful to have a preview of the PR to inspect.


valentin-mladenov (approved):
LGTM

PR #839, opened 2023-07-27
feat: use design tokens by default, add legacy opt-out

williamernest:
Bumps the dependency of packages to depend on the latest Core release the contains the updated tokens. 

BREAKING CHANGE


projects/angular/package.json
```diff
@@ -21,6 +21,6 @@
     "@angular/cdk": "15 || 16",
     "@angular/common": "15 || 16",
     "@angular/core": "15 || 16",
-    "@cds/core": "^5.6.0 || ^6.0.0"
+    "@cds/core": ">= 6.5.0"
```
williamernest, line 24:
```suggestion
    "@cds/core": ">= 6.6.0"
```

projects/ui/package.json
```diff
@@ -10,5 +10,8 @@
   "license": "MIT",
   "bugs": {
     "url": "https://github.com/vmware-clarity/ng-clarity/issues"
+  },
+  "devDependencies": {
+    "@cds/core": "^6.5.0"
```
williamernest, line 15:
```suggestion
    "@cds/core": "^6.6.0"
```

kevinbuhmann, line 15: Shouldn't this be a peer dependency?

kevinbuhmann (changes requested):

projects/ui/src/clr-ui-dark.scss
```diff
@@ -4,5 +4,4 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
-// Import from clr-angular to preserve file structure until sass theming is removed
-@import '../../angular/src/dark-theme';
+@use '@cds/core/styles/theme.dark.min.css';
```
kevinbuhmann, line 7: This file used to be its own entry-point that replaces `clr-ui.min.css`. I think we should either delete it (preferable to me) or make it work on its own.

williamernest: Good catch. In that case, I would prefer incorporating the dark theme vars in the single clr-ui.min.css file and tell developers it's built-in (just toggle the cds-theme attribute)

kevinbuhmann: Agree.

kevinbuhmann (approved):
I think this looks good to me now!

PR #847, opened 2023-07-31
chore(vrt): expand parallel workers for tests

williamernest:
Reduces VRT duration from ~11m to ~6-7m. Further improvements can be made using a matrix and sharding across different jobs but the current structure of handling the unused screenshots would cause problems. We can investigate those later.

### DIFF
--- a/tests/visual-snapshots.spec.ts
+++ b/tests/visual-snapshots.spec.ts
@@ -39,6 +39,10 @@ for (const { storyId, component } of stories) {
 
       await page.goto(`http://localhost:8080/iframe.html?${storyParams}`);
 
+      console.log(
+        `Executing ${screenshotPath}:${storyId}:${theme} story. Worker index: ${process.env.TEST_WORKER_INDEX}. Parallel Index: ${process.env.TEST_PARALLEL_INDEX}`
+      );
+
       await expect(page).toHaveScreenshot(screenshotPath.split(path.sep), {
         animations: 'disabled',
         caret: 'hide',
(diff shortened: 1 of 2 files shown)


tests/visual-snapshots.spec.ts
```diff
@@ -49,4 +49,6 @@ for (const { storyId, component } of stories) {
   }
 }
 
-handleUnusedScreenshots(usedScreenshotPaths);
+test.afterAll(() => {
+  handleUnusedScreenshots(usedScreenshotPaths);
+});
```
kevinbuhmann, line 54: I am pretty sure running this in `afterAll` isn't necessary. The `usedScreenshotPaths` array is populated before any tests are executed, so I don't think we have to wait until `afterAll` to handle unused screenshots.

williamernest: Good catch. Removed it.

kevinbuhmann: Actually, this breaks it. The `handleUnusedScreenshots` needs to know if it should update the snapshots or report a failure. It checks `process.argv.includes('--update-snapshots')` for this, but that doesn't work if this it is called in `afterAll`.

kevinbuhmann: I commented again without refreshing and did not see your comment. Sorry about that. More info, though.

kevinbuhmann: Thanks!

kevinbuhmann (approved):

PR #849, opened 2023-08-02
fix(accordion): set caret color to match text in disabled state

williamernest:
Sets the accordion caret color to match the text color when the header button is disabled.


projects/angular/src/accordion/_accordion.clarity.scss
```diff
@@ -118,6 +118,15 @@
     &:disabled {
       cursor: not-allowed;
 
+      .clr-accordion-status {
```
kevinbuhmann, line 121: We can put `.clr-accordion-status, .clr-accordion-title` here and de-duplicate the `color` declaration.

kevinbuhmann (changes requested):

kevinbuhmann (approved):

PR #853, opened 2023-08-03
fix(forms): use correct border color token in shim (typo fix)

williamernest:
Fixes incorrect token name


kevinbuhmann (approved):
For a better changelog entry, I would change the PR title/commit message to "fix(forms): use correct color border color token in shim" or similar so people know at-a-glance what's actually fixed. Either way, approved.

kevinbuhmann:
Just saw your message in Slack that the approver can merge this.

PR #864, opened 2023-08-07
fix(header): update nav toggle label strings to include open/close

kevinbuhmann:
The previous labels did not adequately describe what the buttons will do.


projects/angular/src/utils/i18n/common-strings.default.ts
```diff
@@ -73,10 +73,10 @@ export const commonStringsDefault: ClrCommonStrings = {
   // Stack View
   stackViewChanged: 'Value changed.',
   // Responsive Nav
-  responsiveNavToggleOpen: 'Navigation menu',
-  responsiveNavToggleClose: 'Navigation menu',
-  responsiveNavOverflowOpen: 'Navigation overflow menu',
-  responsiveNavOverflowClose: 'Navigation overflow menu',
+  responsiveNavToggleOpen: 'Open navigation menu',
+  responsiveNavToggleClose: 'Open navigation menu',
```
dtsanevmw, line 77: Shouldn't this be like that or?
```suggestion
  responsiveNavToggleClose: 'Close navigation menu',
```

projects/angular/src/utils/i18n/common-strings.default.ts
```diff
@@ -73,10 +73,10 @@ export const commonStringsDefault: ClrCommonStrings = {
...
+  responsiveNavOverflowOpen: 'Open navigation overflow menu',
+  responsiveNavOverflowClose: 'Open navigation overflow menu',
```
dtsanevmw, line 79: Shouldn't this be like that or?
```suggestion
  responsiveNavOverflowClose: 'Close navigation overflow menu',
```

dtsanevmw (changes requested):
Two suggestions about the close labels

projects/angular/src/utils/i18n/common-strings.default.ts
kevinbuhmann: Thanks for catching this.

projects/angular/src/utils/i18n/common-strings.default.ts
kevinbuhmann: Thanks for catching this!

dtsanevmw (approved):
LGTM!

PR #867, opened 2023-08-09
feat(range): make use of design tokens

mivaylo:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


mivaylo:
CDE-694

projects/angular/src/forms/styles/_range.clarity.scss
```diff
@@ -32,61 +42,117 @@
         forms-variables.$clr-forms-range-progress-fill-color,
         variables.$clr-use-custom-properties
       );
+      @include forms-mixins.range-border-style(
```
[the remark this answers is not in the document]
mivaylo: Well that would simplify things, but I don't understand the figma this way. I understand we want the flexibility of having 3 different colors (background, progress and border) two of which are equal by default, but not always:

image

Maybe I am wrong though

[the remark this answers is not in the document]
Jinnie: I am just talking about the border of the progress. Which is white in the current implementation and gives the strange overlap effect on the dot.

projects/angular/src/forms/range/range-container.ts
dtsanevmw, line 1: Do we really need a functional changes  just for the token migration?

dtsanevmw (commented):
I agree that we have a "border" around the non-filled part of the range but I see that in core that's not available. So maybe it will be better to stick with migrating what core have and leave the new visual things which are more of a rework for the future release.

projects/angular/src/forms/range/range-container.ts
mivaylo: I am not sure this change qualifies as a functional one, but it is required from the new design (we now want border-radius)

This is the behavior of the .fill-input blue part without the fix:

https://github.com/vmware-clarity/ng-clarity/assets/132376042/0535ddba-8ddb-421d-b44c-1f1d5c255e65

and this is the behavior with the fix:

https://github.com/vmware-clarity/ng-clarity/assets/132376042/5808876a-a9a8-4e7a-b07c-9aa3909e17d0


so I don't know whether we need it but I think we better have it

mivaylo:
Okay, I will revert the logic in the range-container file

projects/angular/src/forms/styles/_mixins.forms.scss
```diff
@@ -205,17 +205,33 @@
...
   -webkit-appearance: none;
   left: 0;
-  height: $height;
 
+  @include mixins.css-var(height, $height-prop, $height, $use-custom-props);
   @include mixins.css-var(background-color, $background-color-prop, $background-color, $use-custom-props);
 }
 
+@mixin range-border-style(
+  $border-color,
```
mindarelus, line 221: maybe rename to border-color-fallback, as well as the other fallbacks, as it's misleading

Jinnie (approved):
LGTM

PR #873, opened 2023-08-10
feat(stepper): add tokens/shim support

williamernest:


projects/ui/src/shim.cds-core.scss
```diff
@@ -1143,6 +1151,11 @@
       border: 1px solid var(--clr-alert-danger-border-color);
     }
   }
+
+  // TODO(CDE-884): Clean up accordion styles and properly separate accordion/stepper styles
```
williamernest, line 1155: Added this instead of adding 2 variables due to the media query. I filed a followup ticket to address the underlying styling issues.

dtsanevmw (changes requested):
- Because of the approach of using a box shadow for the left "border", the `global-space-9` padding is not that accurate because it's starting before the shadow

- There should be the same `global-space-13` space between the title and page number / icon.

dtsanevmw (approved):

PR #880, opened 2023-08-17
feat(wizard): add input for content role

kevinbuhmann:
The wizard content role is defaulted to `main`, but consumers can change it if needed.

This change will allow us to remove nested `main` landmarks in the documentation website.


dtsanevmw (dismissed):
Lgtm!

dtsanevmw (approved):

PR #903, opened 2023-08-22
Revert "chore: rename `beta` prerelease channel to `next`"

kevinbuhmann:
Reverts vmware-clarity/ng-clarity#777


dtsanevmw (approved):

valentin-mladenov (approved):
LGTM

williamernest:
Will this auto deploy a beta when we run our release process Of there are changes in the beta branch? If so, we should delay landing this until after we cut another v15 release today

kevinbuhmann:
> Will this auto deploy a beta when we run our release process if there are changes in the beta branch? If so, we should delay landing this until after we cut another v15 release today

No. You would have have to select the `beta` branch when running the release workflow.

PR #945, opened 2023-09-08
fix(stepper): don't open first panel if clrInitialStep is set

dtsanevmw:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently the stepper is always opening the first panel no matter what. However if the initialStep is different from the first panel it still emits change events for the first panel.

Issue Number: N/A

## What is the new behavior?

No longer opens the first panel if the initial panel is different.

- [ ] Yes
- [X] No

### DIFF
--- a/projects/angular/src/accordion/stepper/models/stepper.model.ts
+++ b/projects/angular/src/accordion/stepper/models/stepper.model.ts
@@ -9,9 +9,14 @@ import { AccordionModel } from '../../models/accordion.model';
 
 export class StepperModel extends AccordionModel {
   private stepperModelInitialize = false;
+  private initialPanel: string;
 
   get allPanelsCompleted(): boolean {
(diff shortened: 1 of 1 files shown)


projects/angular/src/accordion/stepper/models/stepper.model.ts
```diff
@@ -81,6 +87,9 @@ export class StepperModel extends AccordionModel {
   }
 
   private openFirstPanel() {
```
mivaylo, line 89: Do you think it would make more sense if we rename this method to `openInitialPanel()`  and adjust the logic if needed - open initial panel, and if that is not set - open first panel?

mivaylo (approved):

PR #949, opened 2023-09-14
fix(datagrid): read only column title on focus

dtsanevmw:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When focusing a column header currently Screen reader reads the whole content. For example: 'Column name, Filter button, Column separator'.
We used aria-hidden for the children inside so they are hidden when the parent element is being read.

Issue Number: VPAT-2938

## What is the new behavior?

Screen reader now would read just the column name and when you click TAB to focus the individual items inside then it would read their names.

- [ ] Yes
- [X] No


Jinnie (commented):
I am not sure that this needs to be fixed.
If we implement this solution there will be no way for the screenreader user to know if a column has filter.
The problem has 2 aspects... one is traversing the header cells themselves and the second is browsing the table content. As table cells are associated with their header names, we definitely do Not want to have all filter, resize etc. data announced.
But when header cells are directly browsed, then we may want to keep this information.
I suggest strictly checking these details with A11Y team before merging this PR.

PR #976, opened 2023-10-02
fix(alert): remove hostBinding and use renderer for alert-hidden class

valentin-mladenov:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Using HostBinding for `alert-hidden` class Angular throws expression changed error which creates problems for unit tests in user projects

Issue Number: CDE-1249

## What is the new behavior?
Using `renderer` to set `alert-hidden` class on the host element manually which doesn’t have issues with change detection.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/emphasis/alert/alert.ts
+++ b/projects/angular/src/emphasis/alert/alert.ts
@@ -7,13 +7,14 @@
 import {
   ChangeDetectorRef,
   Component,
+  ElementRef,
   EventEmitter,
-  HostBinding,
   Input,
   OnDestroy,
   OnInit,
   Optional,
   Output,
+  Renderer2,
 } from '@angular/core';
 import { Subscription } from 'rxjs';
 
@@ -44,7 +45,9 @@ export class ClrAlert implements OnInit, OnDestroy {
     private iconService: AlertIconAndTypesService,
     private cdr: ChangeDetectorRef,
     @Optional() private multiAlertService: MultiAlertService,
-    private commonStrings: ClrCommonStringsService
+    private commonStrings: ClrCommonStringsService,
+    private renderer: Renderer2,
+    private hostElement: ElementRef
   ) {}
 
   @Input('clrAlertType')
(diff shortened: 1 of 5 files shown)


Jinnie (approved):
Nice!

projects/angular/src/emphasis/alert/alert.ts
```diff
@@ -73,13 +76,19 @@ export class ClrAlert implements OnInit, OnDestroy {
     return this.iconService.iconInfoFromType(this.iconService.alertType).cssClass;
   }
 
-  @HostBinding('class.alert-hidden')
   get hidden() {
     return this._hidden;
```
williamernest, line 80: Does initializing `this._hidden = false` resolve this issue without the additional logic to manually add/remove the class?

projects/demo/src/app/alert/angular/alert-angular-app-level-alerts.demo.html
```diff
@@ -51,6 +51,15 @@
         <span class="title">Header</span>
       </div>
     </header>
+
+    <clr-alerts>
```
williamernest, line 55: This is great for having in the demo application. Adding the specific CDE in a comment could help us understand why this particular case is necessary. 

I'd also like a test to assert that an error is not logged to prevent future regressions.

williamernest (changes requested):

projects/angular/src/emphasis/alert/alert.ts
valentin-mladenov: The error happens when a change is detected `undefined ` to `false` or `false` to `true` it doesn't matter what the default value is, it will happen with default value `true` to `false` as well.
The error lies in the way Angular makes additional check detection in DEV mode. More here: https://angular.io/errors/NG0100

projects/demo/src/app/alert/angular/alert-angular-app-level-alerts.demo.html
valentin-mladenov: Updated test in commit 9b9b33f

valentin-mladenov:
Related to PR #982

PR #985, opened 2023-10-05
feat(buttons): remove fallbacks and shim relations

valentin-mladenov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-1214

- [ ] Yes
- [ ] No


projects/angular/src/button/_buttons.clarity.scss
```diff
@@ -160,14 +108,13 @@
 }
 
 @mixin btn-min-max-widths() {
-  min-width: variables.$clr_baselineRem_3;
-  max-width: variables.$clr_baselineRem_15;
+  min-width: #{tokens.$cds-global-space-14};
```
williamernest, line 111: SCSS interpolation isn't necessary on property declarations.

projects/angular/src/button/_buttons.clarity.scss
```diff
@@ -5,151 +5,99 @@
...
+    cds-icon,
+    clr-icon {
+      @include button-css-var($button-type, color, disabled-color);
+    }
   }
 }
 
 @mixin btn-checked-styles($button-type: default, $type: checkbox) {
   // add semantic prop: group-checked-font-color
-  input[type='#{$type}']:checked + label,
```
williamernest, line 147: Can you elaborate on the differences here? Is this also refactoring the state declarations from other places in the file to live here?

projects/angular/src/button/_properties.buttons.scss
```diff
@@ -4,333 +4,327 @@
...
-      --clr-btn-appearance-standard-font-weight: #{button-variables.$clr-btn-appearance-standard-font-weight};
-      --clr-btn-appearance-standard-height: #{button-variables.$clr-btn-appearance-standard-height};
-      --clr-btn-appearance-standard-padding: #{button-variables.$clr-btn-appearance-standard-padding};
-      --clr-btn-appearance-standard-icon-size: #{variables.$clr_baselineRem_0_5};
+      --clr-btn-appearance-standard-height: var(--clr-btn-height-sm);
+      --clr-btn-appearance-standard-padding: 0 #{tokens.$cds-global-space-6};
+      --clr-btn-appearance-standard-icon-size: #{tokens.$cds-global-space-7};
 
       // form-style button appearance styles - overrideable but not advertised!
-      --clr-btn-appearance-form-line-height: #{button-variables.$clr-btn-appearance-form-line-height};
```
williamernest, line 50: Same comment as before, we should preserve the scss variables so we can get compile-time checking since these properties can be used in so many places.

projects/angular/src/button/button-group/_button-group.clarity.scss
```diff
@@ -26,41 +25,29 @@
...
         height: 100%;
       }
 
-      input[type='radio']:focus,
-      input[type='checkbox']:focus {
-        $outline-shadow: 0 0 variables.$clr_baselineRem_5px
-          var(--clr-btn-group-focus-outline, button-variables.$clr-btn-group-focus-outline);
-        $ie-outline-shadow: 0 0 variables.$clr_baselineRem_5px button-variables.$clr-btn-group-focus-outline;
+      // ATM after token migration (alpha or beta v17)
+      // remove when detailed radio/checkbox groups are implemented
```
williamernest, line 43: Can you file a separate JIRA ticket to track this removal?

williamernest (changes requested):
Overall, it looks good. A lot of moving pieces with this one, but it makes sense with how complex buttons have become. I have a few points to address, let me know if you want to discuss them or when you want me to re-review them.

projects/angular/src/button/button-group/_button-group.clarity.scss
valentin-mladenov: JIRA ticket was created CDE-1310.

projects/angular/src/button/_buttons.clarity.scss
```diff
@@ -5,151 +5,99 @@
  */
 @use 'sass:meta';
 @use '../utils/mixins';
-@use '../utils/variables/variables';
-@use 'variables.buttons' as buttons-variables;
```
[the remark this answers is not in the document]
valentin-mladenov: I don't understand. All CSS vars (eg `--clr-btn-appearance-standard-height`) are tied to CDS tokens (eg `#{tokens.$cds-global-space-9}`). Old CLR variables are not used any more and have no references in the files.

Is your suggestion to tie CLR varables to tokens (eg: `$clr-btn-appearance-standard-height: {tokens.$cds-global-space-9} !default;`) and then reference them as fallbacks?

projects/angular/src/button/_buttons.clarity.scss
valentin-mladenov: This change is tied to previously missing design specification for checkboxes styled as group buttons. New spec is added in figma.
Old behavior can be found [button group](https://clarity.design/documentation/button-group/code) checkbox section.
The full implementation will be resolved with JIRA CDE-1310.

projects/angular/src/button/_properties.buttons.scss
valentin-mladenov: Property in  question `clr-btn-appearance-form-line-height` was never exposed as API and should not be overridden, also all font related CSS now is handled by `generate-typography-token` mixin which is CDS token based.

projects/angular/src/button/_buttons.clarity.scss
valentin-mladenov: corrected in commit fabccb1

projects/angular/src/button/_buttons.clarity.scss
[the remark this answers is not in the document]
williamernest: The problem is developers can still misstype `--clr-btn` properties. In the same way that core provides the token SCSS variables that map to CSS vars in the tokens file we're importing, we should follow the same pattern. This would keep the properties file, where we should map SCSS variables to the appropriate --clr or --cds token, and this file would only use the scss variable. I'm not asking you to do it for all of the variables, but that file should be kept and CSS vars that we're using in this file should be mapped via a SCSS variable from the properties file. 

CDE-864 is the related ticket to provide SCSS variables for all --clr variables in order to get compile time checking.

projects/angular/src/button/_properties.buttons.scss
williamernest: This particular property wasn't the one in question, the previous comment has details.

projects/angular/src/button/_buttons.clarity.scss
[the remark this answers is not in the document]
valentin-mladenov: If I understood correctly the flow will be to reference CLR variable that is mapped to CLR property which is mapped to CDS variable which i mapped to CDS property.

Example
`
$clr-btn-appearance-standard-height: var(--clr-btn-appearance-standard-height);` -> `--clr-btn-appearance-standard-height: {tokens.$cds-global-space-9};` -> ` $cds-global-space-9: var(--cds-global-space-9)` -> `var(--cds-global-space-9): 16px`.
And in buttons `.btn { height: button-vars.$clr-btn-appearance-standard-height; }`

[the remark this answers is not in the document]
williamernest: Yes. That ensures that every instance of us using the $clr-btn- scss variable will be checked at compile time. We only need to thoroughly validate it is correctly spelled in one place despite potentially using it in multiple places. It also gets us closer to our secondary goal of exporting a file of SCSS variables so our product teams can get compile-time checking as well.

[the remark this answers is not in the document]
valentin-mladenov: We are following 2 approaches.

1.  One is consistent with static variable property assignment ` $clr-btn-appearance-standard-height: var(--clr-btn-appearance-standard-height);` . With `if {} else {}` strategy. [Line 54 through 117](https://github.com/vmware-clarity/ng-clarity/blob/buttons-remove-shim_cde-1214/projects/angular/src/button/_variables.buttons.scss#L54).
2. The other is using map `$clr-btn-primary: ( color: var(--clr-btn-primary-color), ... ) !default;`. With `map-get` strategy.[line 252 through 269](https://github.com/vmware-clarity/ng-clarity/blob/buttons-remove-shim_cde-1214/projects/angular/src/button/_variables.buttons.scss#L252).

Question is for such a variable bundle which will be the better option. The second defiantly will NOT follow the requirement.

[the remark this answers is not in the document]
williamernest: We can address the second strategy later as part of CDE-864.

valentin-mladenov:
@williamernest you can re-review this PR.

williamernest (approved):

PR #1018, opened 2023-10-26
fix(vertical-nav): fix inconsistent padding and font-weight of nav item groups

williamernest:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

In the default v16 theme, a top-level anchor tag category item will render indented and not visually aligned with the other top-level list items. The font-weight of the top-level anchor tag items is also incorrect, as all top-level groups should use the 600 font-weight.

Issue Number: CDE-1349

## What is the new behavior?

The top-level category items are rendered in the appropriate weight (both buttons and links). An additional ticket has been created to investigate if the vertical nav should support both buttons and links (angular router works with both, and links have problematic a11y issues associated with them in this context). This change also fixes the misalignment between links and buttons as top-level category items. 

For the "old" theme, this change changes the top-level nav group link items to use the same heavier font-weight as buttons in the same context. Since this change updates the anchor tag top-level nav group items to match the buttons used in the same context, I consider it to be an acceptable compromise for the old theme since that theme will be removed in the upcoming v17. 

- [ ] Yes
- [x] No


projects/ui/src/shim.cds-core.scss
```diff
@@ -1826,6 +1826,9 @@
   .clr-vertical-nav.has-icons .nav-group-children .nav-link {
     padding: var(--clr-vertical-nav-item-child-padding);
   }
+  .clr-vertical-nav.has-nav-groups .nav-group-content > .nav-link {
+    font-weight: var(--clr-vertical-nav-item-top-level-font-weight);
+  }
```
dtsanevmw, line 1831: Is this needed when we have it added inside [_vertical-nav.clarity.scss](https://github.com/vmware-clarity/ng-clarity/pull/1018/files#diff-241bd325c28d64d8afd38964345109e89ed416488824ee6bc9d6625d0a758938R513) 
 and it's pointing to the same property?

dtsanevmw (changes requested):
Just one small comment for duplicated rule.

projects/ui/src/shim.cds-core.scss
williamernest: Good catch!

dtsanevmw (approved):

PR #1019, opened 2023-10-27
chore(vertical-nav): update storybook routing to include all top level items

williamernest:
Updates the routing vertical-nav story to include all top-level item types (links without a group, links with a group, buttons with a group).


dtsanevmw (approved):
Some indentations in the template only.

PR #878, opened 2023-08-15
feat(datepicker): token migration and update shim

valentin-mladenov:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #1034, opened 2023-11-06
feat(datepicker): shim relation and fallback removal 

valentin-mladenov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-1225

- [ ] Yes
- [x] No


projects/angular/src/forms/datepicker/_datepicker.clarity.scss
```diff
@@ -8,40 +8,32 @@
...
   //Other Props
   color: inherit;
   border: none;
-  @include mixins.css-var(
-    border-radius,
-    clr-calendar-cell-border-radius,
-    variables.$clr-global-borderradius,
-    variables.$clr-use-custom-properties
-  );
+  border-radius: tokens.$cds-alias-object-border-radius-100;
```
Jinnie, line 27: Ok, why do we decide to directly bring a CDS variable here and not keep a configurable `clr-calendar-cell-border-radius` property and variable that point to the CDS var?

What's the difference between this case and the others where we do the latter?

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
```diff
@@ -8,40 +8,32 @@
 @use '../../utils/mixins';
 @use '../../utils/variables/variables.global';
```
Jinnie, line 9: We should be able to remove this `@use` now that all globals are replaced with local vars or tokens.

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
```diff
@@ -445,81 +276,26 @@
...
-        datepicker-variables.$clr-calendar-active-cell-background-color,
-        variables.$clr-use-custom-properties
-      );
-      @include mixins.css-var(
-        color,
-        clr-calendar-active-cell-color,
-        datepicker-variables.$clr-calendar-active-cell-color,
-        variables.$clr-use-custom-properties
-      );
+      background: datepicker-variables.$clr-calendar-active-cell-background-color;
```
Jinnie, line 293: I think we should be switching to `background-color` when we create/update content.
No need to mass replace, though it won't be too bad.

projects/angular/src/forms/datepicker/_variables.datepicker.scss
```diff
@@ -3,41 +3,55 @@
  * This software is released under MIT license.
  * The full license information can be found in LICENSE in the root directory of this project.
  */
-@use '../../utils/variables/variables';
+@use '@cds/core/tokens/tokens.scss';
+
+$clr-calendar-click-target: #{tokens.$cds-global-space-11};
```
Jinnie, line 8: Why do we hardcode some values here, while others need a property? 
Wasn't the idea to use this file as compile time checking only? Now it still behaves like a secondary theme layer (SCSS based).
If we don't need a property and want to hardcode something, we can easily do it in the component scss file itself.

Jinnie (commented):
A few questions only, minor things and convention questions/comments.

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
valentin-mladenov: `clr-calendar-cell-border-radius` is introduced with TM PR #878. The property is not needed any more and using CDS token directly is cleaner.

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
valentin-mladenov: Correct, but global `variables` file have map `clr-layers` that stores `z-index` level and istoo hard to use without it. It's used only here: 
`z-index: map.get(variables.$clr-layers, dropdown-menu);`

At later stage if there is CDS tokens about those levels we can use them.

projects/angular/src/forms/datepicker/_variables.datepicker.scss
valentin-mladenov: In this example `$clr-calendar-click-target` don't have property to be associated with (`--clr-calendar-click-target`), but is old (before TM) variable that might be used from the client applications.

The idea is to keep it and replace it's `CLR value` with `CDS token`.

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
Jinnie: true, it remains used.

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
Jinnie: I still think this can be exposed, but I can agree with the "when in doubt, take the conservative approach". We can add it back if needed.

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
Jinnie: Ok, if our property is called "color" then it would better be color.
But for backward compatibility sake, we'd better leave it open for any possible overrides with img backgrounds.
So, I agree to keep this as it is.

projects/angular/src/forms/datepicker/_variables.datepicker.scss
Jinnie: I think we can bring this into the component file.
Click target areas should be system wide usability concerns. If it does not meet a11y criteria, we'd better fix it than expose it to the user. It's a too deep technical detail for me.
If you prefer it to stay this way, I am not strongly against it. You can dismiss this comment and merge it.

Jinnie (approved):

PR #1045, opened 2023-11-08
chore: prevent installing Angular 17 instead of Angular 16

kevinbuhmann:
The previous command updated all unlisted Angular packages to v17.


valentin-mladenov (approved):
LGTM

PR #1067, opened 2023-11-16
feat(wizard): add back inline wizards in 15.x

kevinbuhmann:
This reverts commit 7470769480b843dc332adbd25f67624ce4979e56 (#419), removes the deprecated comments, and adds a comment explaining when inline wizards are used.

Inline wizards are used by vSphere plugins to display a wizard inside an iframe that is inside a modal.

CDE-154


williamernest (approved):
LGTM

PR #1074, opened 2023-11-23
feat(forms): support for dynamic control change

dtsanevmw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

With dynamic controls which are not available initially but when they appear available combined with a value, we don't track that and status validation is not triggered and it stays in NONE state.

Issue Number: CDE-1464

## What is the new behavior?

Now we trigger validation if there is a change inside the control and validations works.

- [ ] Yes
- [X] No


projects/angular/src/forms/common/wrapped-control.spec.ts
```diff
@@ -193,6 +193,46 @@ class WithControlAndSuccess {
...
+        <label> Label </label>
+        <textarea testControl3 [(ngModel)]="object['control']"></textarea>
+        <clr-control-success>Successful!</clr-control-success>
+      </test-wrapper3>
+      <form-wrapper> </form-wrapper
+    ></form-wrapper>
+  `,
+})
+class WithDynamicNgControl {
+  object = {};
```
wghglory, line 230: object is a reserved keyword. It's better to use another variable name

wghglory:
Hello @dtsanevmw,

Could you please provide an estimated timeframe for the resolution of this issue? Our product heavily relies on it, and having an idea of when it will be fixed would greatly assist our planning. Thank you for your prompt attention to this matter.

projects/angular/src/forms/common/wrapped-control.spec.ts
dtsanevmw: Thanks!

projects/angular/src/forms/common/wrapped-control.spec.ts
```diff
@@ -193,6 +193,46 @@ class WithControlAndSuccess {
...
+@Component({
+  template: `
+    <form-wrapper>
+      <test-wrapper3>
+        <label> Label </label>
+        <textarea testControl3 [formControl]="form.get('control') || form.get('alternative')"></textarea>
+        <clr-control-success>Successful!</clr-control-success>
+      </test-wrapper3>
+      <form-wrapper> </form-wrapper
+    ></form-wrapper>
```
kevinbuhmann, line 205: The closing `>` should be on the previous line.

kevinbuhmann, line 205:
```suggestion
      <form-wrapper></form-wrapper>
    </form-wrapper>
```

projects/angular/src/forms/common/wrapped-control.spec.ts
```diff
@@ -193,6 +193,46 @@ class WithControlAndSuccess {
...
+@Component({
+  template: `
+    <form-wrapper>
+      <test-wrapper3>
+        <label> Label </label>
+        <textarea testControl3 [(ngModel)]="form['control']"></textarea>
+        <clr-control-success>Successful!</clr-control-success>
+      </test-wrapper3>
+      <form-wrapper> </form-wrapper
+    ></form-wrapper>
```
kevinbuhmann, line 226:
```suggestion
      <form-wrapper></form-wrapper>
    </form-wrapper>
```

projects/demo/src/app/forms/dynamic-controls/dynamic-controls.demo.ts
```diff
@@ -0,0 +1,131 @@
...
+    const signatureControl = this.form.controls['s3'].get('tlsSignature') as FormControl;
+
+    const value = signatureControl.value ? 'tlsSignature' : 'publicSigned';
+
+    this.certValidation.setValue(value);
+  }
+
+  private registerCertValidateChange() {
+    this.certValidation.valueChanges.subscribe(value => {
+      const tlsSignature = this.form.controls['s3'].get('tlsSignature') as FormControl;
```
kevinbuhmann, line 114:
```suggestion
      const tlsSignature: FormControl = this.form.controls['s3'].get('tlsSignature');
```

Does type annotation work instead?

projects/demo/src/app/forms/dynamic-controls/dynamic-controls.demo.ts
```diff
@@ -0,0 +1,131 @@
...
+  });
+
+  ngOnInit() {
+    this.registerPlatformChange();
+
+    this.registerCertValidateChange();
+  }
+
+  private registerPlatformChange() {
+    this.form.controls.platform.valueChanges.subscribe(value => {
```
kevinbuhmann, line 31: This subscription is never unsubsubscribed.

projects/demo/src/app/forms/dynamic-controls/dynamic-controls.demo.ts
```diff
@@ -0,0 +1,131 @@
...
+  private setCertValidationValue() {
+    const signatureControl = this.form.controls['s3'].get('tlsSignature') as FormControl;
+
+    const value = signatureControl.value ? 'tlsSignature' : 'publicSigned';
+
+    this.certValidation.setValue(value);
+  }
+
+  private registerCertValidateChange() {
+    this.certValidation.valueChanges.subscribe(value => {
```
kevinbuhmann, line 113: This subscription is never unsubsubscribed.

projects/demo/src/app/forms/dynamic-controls/dynamic-controls.demo.ts
dtsanevmw: It was the initial reproduction demo which I was going to clean after the first review but it stood like that. Cleaned it up.

projects/demo/src/app/forms/dynamic-controls/dynamic-controls.demo.ts
dtsanevmw: Same as above.

projects/demo/src/app/forms/dynamic-controls/dynamic-controls.demo.ts
```diff
@@ -4,128 +4,39 @@
...
+          this.form.addControl('control', new FormControl('test'));
+          this.form.get('control')?.markAsTouched();
+          this.form.get('control')?.updateValueAndValidity();
         }
       });
-    });
+  }
+
+  ngOnDestroy() {
+    this.destroyed.complete();
```
kevinbuhmann, line 40:
```suggestion
    this.destroyed.next();
    this.destroyed.complete();
```

The `takeUntil` operator will only unsubscribe when the `notifier` emits a value.

projects/demo/src/app/forms/dynamic-controls/dynamic-controls.demo.ts
```diff
@@ -4,128 +4,39 @@
...
-      setTimeout(() => {
-        if (tlsSignature.value) {
-          tlsSignature.markAsTouched();
-          tlsSignature.updateValueAndValidity();
+  ngOnInit() {
+    this.form
+      .get('platform')
+      ?.valueChanges.pipe(takeUntil(this.destroyed))
+      .subscribe(res => {
+        console.log(res);
```
kevinbuhmann, line 30: Was this for testing? If so, should this be removed?

projects/demo/src/app/forms/dynamic-controls/dynamic-controls.demo.ts
```diff
@@ -4,128 +4,39 @@
...
+        if (res === 'two') {
```
kevinbuhmann, line 31: You can use `filter(res => res === 'two')` instead.

projects/angular/src/forms/common/wrapped-control.ts
```diff
@@ -100,6 +115,22 @@ export class WrappedFormControl<W extends DynamicWrapper> implements OnInit, OnD
...
+      const changes = this.differ.diff(this.ngControl);
+      if (changes) {
+        changes.forEachChangedItem(change => {
+          if (
+            (change.key === CHANGE_KEYS.FORM || change.key === CHANGE_KEYS.MODEL) &&
+            change.currentValue !== change.previousValue
+          ) {
+            this.triggerValidation();
+          }
+        });
```
kevinbuhmann, line 129:
```suggestion
        let validationNeeded = false;

        changes.forEachChangedItem(change => {
          if (
            (change.key === CHANGE_KEYS.FORM || change.key === CHANGE_KEYS.MODEL) &&
            change.currentValue !== change.previousValue
          ) {
            validationNeeded = true;
          }
        });

       if (validationNeeded) {
         this.triggerValidation();
       }
```

Should we do something like this to prevent validation from being trigger more than once? Not sure if that's a possible issue or not.

projects/angular/src/forms/common/wrapped-control.ts
```diff
@@ -100,6 +115,22 @@ export class WrappedFormControl<W extends DynamicWrapper> implements OnInit, OnD
     }
   }
 
+  ngDoCheck() {
```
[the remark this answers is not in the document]
dtsanevmw: Yeah, makes more sense. Thanks!

projects/angular/src/forms/common/wrapped-control.ts
dtsanevmw: It would trigger only if the binded control is changed and since I'm tracking only that the items in `changes` will be always 1.

projects/angular/src/forms/common/wrapped-control.spec.ts
```diff
@@ -193,6 +193,44 @@ class WithControlAndSuccess {
...
+class WithDynamicFormControl {
+  form = new FormGroup<any>({
+    alternative: new FormControl(),
+  });
+
+  addControl() {
+    this.form.addControl('control', new FormControl('TEST'));
+  }
+}
+@Component({
```
kevinbuhmann, line 216: nit: missing blank line between classes

projects/angular/src/forms/common/wrapped-control.spec.ts
```diff
@@ -339,6 +377,29 @@ export default function (): void {
...
+        setupTest(this, WithDynamicFormControl, TestControl3);
+        const cb = jasmine.createSpy('cb');
+        const sub = this.ifControlStateService.statusChanges.subscribe((control: CONTROL_STATE) => cb(control));
+        expect(cb).toHaveBeenCalledWith(CONTROL_STATE.NONE);
+        this.fixture.componentInstance.addControl();
+        this.fixture.detectChanges();
+        expect(cb).toHaveBeenCalledWith(CONTROL_STATE.VALID);
+        sub.unsubscribe();
+      });
+      it('with ng-control directive', function (this: TestContext) {
```
kevinbuhmann, line 391: nit: missing blank line between tests

kevinbuhmann, line 391:
```suggestion

      it('with ng-control directive', function (this: TestContext) {
```

projects/angular/src/forms/common/wrapped-control.spec.ts
kevinbuhmann, line 216:
```suggestion

@Component({
```

projects/demo/src/app/forms/dynamic-controls/dynamic-controls.demo.ts
```diff
@@ -0,0 +1,44 @@
...
+    scenario: new FormControl('one'),
+  });
+
+  constructor(private cd: ChangeDetectorRef) {}
+
+  ngOnInit() {
+    this.form
+      .get('platform')
+      ?.valueChanges.pipe(
+        takeUntil(this.destroyed),
```
kevinbuhmann, line 29: This must be last to ensure all inner subscriptions are unsubscribed.

dtsanevmw: If it's last then it will go trough again and then complete the inner subscriptions but if it's first then it completes and doesn't pass any values to the next pipe operators.
Created an example where it can be seen: https://stackblitz.com/edit/typescript-3ckpfc?devToolsHeight=33&file=index.ts

kevinbuhmann: That is because all of the value are emitted synchronously. This is a bad test.

If you emit the values in different frames using `setTimeout`:
- With `takeUntil` first, it will continue to emit values after destroy and the inner subscription in the `mapCopy` operator is not unsubscribed. Both of these are memory leaks.
- With `takeUntil` last, both issues are solved.

https://stackblitz.com/edit/typescript-nqfkh9?file=index.ts

It might be better to just call `unsubscribe` explicitly in `ngOnDestroy` so none of this is an issue.

kevinbuhmann: I am looking into this further. I am worried out the inner subscriptions of subsequent operators not being unsubscribed.

kevinbuhmann: The inner subscriptions of subsequent operators are not unsubscribed if `takeUntil` is first.

Here is an example with four demos, `takeUntil` first and last with values emitted synchronously and asynchronously for both:

https://stackblitz.com/edit/typescript-nqfkh9?devToolsHeight=33&file=index.ts

With `takeUntil` first: While values are not processed after destroy synchronously or asynchronously, the inner subscription in the `mapCopy` operator is never unsubscribed. This is a memory leak.

With `takeUntil` last: Synchronously-emitted values are processed after destroy, but the inner subscription in the `mapCopy` operator is always unsubscribed. This is not a memory leak because once the next frame hits, everything is done.

It might be better to just call `unsubscribe` explicitly in `ngOnDestroy` so none of this is an issue.

kevinbuhmann: The inner subscriptions of subsequent operators are not unsubscribed if `takeUntil` is first.

Here is an example with four demos, `takeUntil` first and last with values emitted synchronously and asynchronously for both:

https://stackblitz.com/edit/typescript-nqfkh9?devToolsHeight=33&file=index.ts

With `takeUntil` first: While values are not processed after destroy synchronously or asynchronously, the inner subscription in the `destination` operator is never unsubscribed. This is a memory leak.

With `takeUntil` last: While values are processed after destroy if (and only if) emitted synchronously, the inner subscription in the `destination` operator is always unsubscribed. This is not a memory leak because everything is done once the next frame hits. I think it's likely for a source event to be emitted synchronously in the same frame when the component is being destroyed, so that probably a non-issue as well.

All that said, it might be better to just call `unsubscribe` explicitly in `ngOnDestroy` so none of this is an issue at all.

kevinbuhmann: The inner subscriptions of subsequent operators are not unsubscribed if `takeUntil` is first.

Here is an example with four demos, `takeUntil` first and last with values emitted synchronously and asynchronously for both:

https://stackblitz.com/edit/typescript-nqfkh9?devToolsHeight=33&file=index.ts

With `takeUntil` first: While values are not processed after destroy synchronously or asynchronously, the inner subscription in the `destination` operator is never unsubscribed. This is a memory leak.

With `takeUntil` last: While values are processed after destroy if (and only if) emitted synchronously, the inner subscription in the `destination` operator is always unsubscribed. This is not a memory leak because everything is done once the next frame hits. I think it's unlikely for a source event to be emitted synchronously in the same frame when the component is being destroyed, so that probably a non-issue as well.

All that said, it might be better to just call `unsubscribe` explicitly in `ngOnDestroy` so none of this is an issue at all.

kevinbuhmann: https://github.com/cartant/eslint-plugin-rxjs/blob/main/docs/rules/no-unsafe-takeuntil.md

dtsanevmw: Thanks for digging further! Moved it to be last

kevinbuhmann: Thanks!

kevinbuhmann (approved):

PR #1089, opened 2023-12-06
chore: add retries to playwright to prevent flaky tests

williamernest:
Update the playwright config to allow two retries for failed tests. When a test fails, that individual test will be retried twice, and if it fails every time, the last screenshot will be retained for different comparisons. This should help overcome the flaky button issues seen on this PR and #1088 .


williamernest:
I ran this five times consecutively without any flakiness.  
image

dtsanevmw (approved):

kevinbuhmann (approved):

PR #1102, opened 2023-12-12
feat(ng): update ng deps to remove angular 15

williamernest:
Upgrade dependencies and workflow to remove Angular 15.


williamernest:
Closing this until we can upgrade storybook to v7, as storybook v6 does not support angular v16.

PR #1103, opened 2023-12-12
chore(alert): update stories for alerts

williamernest:
Updates the alert storybook stories to CSFv3. Separates the stories into app level and standard alerts.


Jinnie (commented):
Lightway + Small case seems to be missing.
Not sure we want it, but we used to have it previously.
Approving, just want to be sure the above is considered.

Jinnie (approved):

PR #1104, opened 2023-12-12
chore(accordion): update storybook stories to csfv3 (port of #996)

williamernest:
Updates the accordion stories to follow CSFv3. Adds individual stories for multiple configurations. Adds inline comments/type in component to ensure a description and appropriate type is used in storybook docs mode.


Jinnie (approved):
Looks great. Both code-wise and storybook.

PR #1129, opened 2024-01-16
chore: delete visual snapshot update branch immediately

kevinbuhmann:
There was code to cleanup visual snapshot update branches if no changes, but that job only runs after after failed build. This resulted in visual snapshot update branches persisting after successful builds.

We can solve this by immediately deleting the visual snapshot branches. We just need the visual snapshot update commit in the repo. There is no need to keep the branches around for any length of time.

### DIFF
--- a/.github/workflows/pr-visual-snapshot-update-bot.yml
+++ b/.github/workflows/pr-visual-snapshot-update-bot.yml
@@ -50,11 +50,11 @@ jobs:
             # push changes
             git push $repositoryUrl HEAD:refs/heads/$updateBranchName --force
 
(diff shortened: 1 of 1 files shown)


kevinbuhmann:
> If that runs on successful build I'm not really sure that is what we want because most of the time is good to see the changes until we close the PR. Can this trigger once the PR get merged ?

The commit will still exist even if the branch is deleted.

kevinbuhmann:
> The commit will still exist even if the branch is deleted.

On that note, I think we should just make it delete the branch immediately. It doesn't need to persist.

dtsanevmw (approved):
That's perfect 👍

PR #1133, opened 2024-01-17
chore: disable spinner animation in visual regression tests

kevinbuhmann:
The spinner animation sometimes interferes with the visual regression tests.


kevinbuhmann:
The spinner is not the issue. The success button flashes an icon before the text.

Issue #816, opened 2023-07-17
Datepicker: min/max validation not updated


ekrismer:
## Describe the bug

The `min` and `max` input of the clarity date picker do not react to changes to the provided value.

## How to reproduce

See this stackblitz: https://stackblitz.com/edit/clarity-light-theme-clr15-xuikzh?file=src%2Fapp%2Fapp.component.html

1.  Update or clear `minDate`
2.  Update or clear `maxDate`

## Expected behavior

When clearing or updating the min/max dates the input should become valid.

## Versions

**Clarity version:**

@clr/angular `15.1.0`
@angular/core `15.0.4`
@cds/core `6.4.4`

## Additional notes

The validation feature was introduced in https://github.com/vmware-clarity/ng-clarity/pull/360

dtsanevmw:
Issue is fixed and merged in 15.x, 16.x and 17.x

PR #1138, opened 2024-01-18
fix(date-picker): trigger validation on min/max date change

venkateshr06:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #816  CDE-1584

- [ ] Yes
- [x] No


projects/angular/src/forms/datepicker/date-input.ts
```diff
@@ -299,4 +301,13 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
...
   }
+
+  private retriggerValidation() {
+    if (this.datepickerHasFormControl() && !this.usingNativeDatepicker()) {
+      this.control?.control?.updateValueAndValidity({ onlySelf: true, emitEvent: false }); // Added onlySelf, emitEvent purposefully inorder to prevent unnecessary valuechange triggers
+      if (this.control.control && this.control.control.touched) {
+        this.control.control.setErrors(this.control.control.errors); // Inorder to revalidate and apply validation errors(if any) only when form control is touched.
+      }
+    }
+  }
```
dtsanevmw, line 312:
```suggestion
 private triggerControlValidation() {
    if (this.datepickerHasFormControl()) {
      this.control.control?.updateValueAndValidity({ onlySelf: true, emitEvent: false });
      this.control.control?.setErrors(this.control.control.errors);
    }
  }
```

- Put the function where the first private functions start (you should receive an error for order probably from the lint)
- `usingNativeDatepicker` is  not needed as validation would happen always when `clrDate` is used on the input.
- I renamed it to `triggerControlValidation` because there is another function called `triggerValidation` which is for internal control state service.

dtsanevmw (changes requested):
Thanks! Just some comments around fitting it into project code structure

projects/demo/src/app/datepicker/datepicker-min-max.html
```diff
@@ -4,24 +4,35 @@
...
+  <clr-date-container>
+    <label for="dateControl">Date</label>
+    <input id="dateControl" name="date" type="date" clrDate formControlName="date" [min]="minDate" [max]="maxDate" />
+    <clr-control-error *clrIfError="'min'">Value must be on or after min date</clr-control-error>
+    <clr-control-error *clrIfError="'max'">Value must be on or before max date</clr-control-error>
+  </clr-date-container>
+</form>
+
+<form clrForm>
+  <h6>Template Driven Form</h6>
```
kevinbuhmann, line 20: Please put both headings either inside or outside the form.

projects/angular/src/forms/datepicker/date-input.ts
```diff
@@ -299,4 +301,13 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
       .pipe(filter(date => !!date))
       .subscribe(() => this.datepickerFocusService.focusInput(this.el.nativeElement));
   }
+
+  private retriggerValidation() {
+    if (this.datepickerHasFormControl() && !this.usingNativeDatepicker()) {
+      this.control?.control?.updateValueAndValidity({ onlySelf: true, emitEvent: false }); // Added onlySelf, emitEvent purposefully inorder to prevent unnecessary valuechange triggers
```
kevinbuhmann, line 307: I'm concerned about not emitting `statusChanges` here because it can change from valid to invalid or vice versa. I would rather emit extra events than skip events and possibly cause hard-to-find bugs. Consumers of `valueChanges` and `statusChanges` can use `distinctUntilChanged` if needed.

kevinbuhmann (changes requested):

projects/angular/src/forms/datepicker/date-input.ts
kevinbuhmann, line 307: Also, the `onlySelf` seems unnecessary. Why would a date form control have child controls? I'm mostly curious about this one because it's probably not a problem either way.

venkateshr06: Agree! statusChange event will not be triggered when emitEvent is set to false, but control setErrors method will trigger the validations and its already triggering the statusChange event. Removing emitEvent option would trigger additional valueChanges at application end on every min/max setters.

venkateshr06: onlySelf will prevent bubbling up event to the parent form element, have updated the code! Thanks.

projects/angular/src/forms/datepicker/date-input.ts
venkateshr06: Resolved!

projects/angular/src/forms/datepicker/date-input.ts
kevinbuhmann: Ah, got it. Thanks!

kevinbuhmann: Thanks!

projects/demo/src/app/datepicker/datepicker-min-max.html
kevinbuhmann: Thanks!

projects/angular/src/forms/datepicker/date-input.ts
```diff
@@ -192,6 +194,15 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
     }
   }
 
+  private triggerControlValidation() {
+    if (this.datepickerHasFormControl()) {
+      this.control?.control?.updateValueAndValidity({ emitEvent: false }); // Added emitEvent purposefully inorder to prevent unnecessary valuechange triggers
```
kevinbuhmann, line 199:
```suggestion
      // Set `emitEvent` to false to prevent unnecessary value change event. Status change event will be emitted below by `setErrors`.
      this.control.control?.updateValueAndValidity({ emitEvent: false });
```

The code is already checking if a control exists. 

Also, I'd put the comment on the preceding line for readability (instead of one long line) and clarify about the status change event here as well.

projects/angular/src/forms/datepicker/date-input.ts
```diff
@@ -192,6 +194,15 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
...
+      if (this.control.control && this.control.control.touched) {
```
kevinbuhmann, line 200:
```suggestion
      if (this.control.control?.touched) {
```

This can be simplified a bit.

projects/angular/src/forms/datepicker/date-input.ts
```diff
@@ -192,6 +194,15 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
...
+        this.control.control.setErrors(this.control.control.errors); // Inorder to revalidate and apply validation errors(if any) only when form control is touched & trigger status change event.
```
kevinbuhmann, line 201:
```suggestion
        // Apply validation errors (if any) and trigger status change event when control is touched.
        this.control.control.setErrors(this.control.control.errors);
```

Same here. I'd put the comment on the preceding line so that the line isn't so long.

I also reworded a bit. If I understand correctly, this line doesn't revalidate. That is done by `updateValueAndValidity`.

kevinbuhmann (changes requested):

kevinbuhmann (approved):

dtsanevmw (approved):
LGTM!

PR #1168, opened 2024-01-26
fix(accordion): header spacing and disabled panel content color

mivaylo:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-1603

- [x] Yes
- [ ] No

## Other information

This PR introduces changes in the API (removed obsolete custom properties) of the accordion / stepper and the appearance of the stepper.

There are spacing differences (hence it was marked as breaking change) and color differences. Previous version was not following figma specs or was not well defined, so this is marked as a "fix".

Removed obsolete custom property is:
--clr-accordion-header-button-padding

Spacing changes are:

-we used to have stepper header panel padding set to 18px on top and bottom, and the overall height of the header panel was calculated to 56.4px; overall height is calculated to exactly 56px;
-when a step is completed header panel used to have bigger height (60px) because of the size of the complete icon, now it stays to 56px on complete/error etc;
-gap between header panel container children was changed to 24px to reflect figma spec;
-there was smaller padding-left of the header panel container and some margin-left to the first child; changed to bigger padding-left and no margin for the child -> this helps with vertical alignments when there are two rows in smaller screens;
-top and bottom padding for header container was reduced from 18 to 8px for "breathing" space when there is a very long text or a small screen;

Preview link look in normal screen (padding right is 24 and padding left is 32 because of the box-shadow blue indicator which is 8px):

image

Smaller screen (big enough to have only one row though):

image

When we have two rows (they are now vertically aligned, the gap is 24px by design and that's why we have so big space between the rows):

image

When texts are long (there is this "breathing" space and texts are not fixed to the header container border:

image

BREAKING CHANGE: Changed spacing in stepper header


projects/angular/src/accordion/accordion-panel.html
```diff
@@ -41,7 +41,10 @@
         (@toggle.done)="collapsePanelOnAnimationDone(panel)"
         class="clr-accordion-content"
       >
-        <div class="clr-accordion-inner-content">
+        <div
+          class="clr-accordion-inner-content"
+          [class.clr-accordion-inner-content-disabled]="isAccordion && panel.disabled"
```
williamernest, line 46: I don't see the inner content disabled state specified in figma, can you verify with design that we do in fact want the content panel to render a disabled color. If so, we should set the styling of the inner content from the clr-accordion-panel-disabled class instead of adding a new one here.

dtsanevmw (approved):

PR #1231, opened 2024-02-13
fix(datagrid): row does not collapse if it crosses table height in safari

spadidapelli:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-1570

- [ ] Yes
- [x] No


projects/angular/src/data/datagrid/utils/key-navigation-grid.controller.ts
```diff
@@ -144,7 +146,13 @@ export class KeyNavigationGridController implements OnDestroy {
...
     const items = getTabableItems(activeCell);
     const item = activeCell.getAttribute('role') !== 'columnheader' && items[0] ? items[0] : activeCell;
-    item.focus();
+    if (scrollCell) {
+      item.focus();
+    } else {
+      item.focus({
+        preventScroll: true,
+      });
+    }
```
kevinbuhmann, line 155: This can be simplified to `item.focus({ preventScroll });`. I'd flip the boolean argument to make this work.

projects/angular/src/data/datagrid/utils/key-navigation-grid.controller.ts
```diff
@@ -111,8 +112,9 @@ export class KeyNavigationGridController implements OnDestroy {
             const activeItem = this.rows
               ? (Array.from(this.rows[y].querySelectorAll(this.config.keyGridCells))[x] as HTMLElement)
               : null;
+            const scrollItem = true;
```
kevinbuhmann, line 115: There is a better way to do this. You can change the signature of the `setActiveCell` to take an object as the second parameter so that the boolean literal is named there instead of in a variable that is used once.

kevinbuhmann (changes requested):

PR #1244, opened 2024-02-15
fix(modal): prevent error when modal is escaped with a focused input

kevinbuhmann:
If you have an input inside a modal (specifically an input with `[(ngModel)]`) and escape the modal with the input focused, an "expression change after it was checked" error might be thrown.

I think this is due to Angular updating the input's touched status after the element has been destroyed due to modal being closed.

Blurring (unfocusing) the active element before closing the modal solves this issue. I think this is A safe fix because the modal's trigger element is focused immediately after the modal is closed (so focus is not lost).

CDE-1662


projects/angular/src/modal/modal-stack.service.ts
```diff
@@ -45,6 +45,10 @@ export class ModalStackService {
 
   private onKeyUp(event: KeyboardEvent) {
     if (this.modalStack.length && normalizeKey(event.key) === Keys.Escape) {
+      // We blur the active element because escaping with an input element in focus could cause
```
williamernest, line 48: Thanks for quickly resolving this issue. Can you write a unit test to ensure we don't reintroduce this issue in the future?

kevinbuhmann: Yeah, I'll try.

kevinbuhmann: I am having trouble producing the error in a unit test.

williamernest: Ok. Thanks for looking into it.

williamernest (approved):

PR #1309, opened 2024-03-11
fix(tabs): fix aria-labelledby attribute

mivaylo:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-1354

- [ ] Yes
- [x] No

## Other information

This PR fixes getter method ariaLabelledBy to return correct tabLinkId. Previously nothing was found in the getter function so the aria-labelledby attrbute was not set. To be able to find the correct tabLink I had to expose a getter of the id of the tabLink.

### DIFF
--- a/projects/angular/src/layout/tabs/tab-link.directive.ts
+++ b/projects/angular/src/layout/tabs/tab-link.directive.ts
@@ -40,7 +40,7 @@ export class ClrTabLink {
 
   constructor(
     public ifActiveService: IfActiveService,
-    @Inject(IF_ACTIVE_ID) private id: number,
+    @Inject(IF_ACTIVE_ID) private _id: number,
     public el: ElementRef,
     private cfr: ComponentFactoryResolver,
     private viewContainerRef: ViewContainerRef,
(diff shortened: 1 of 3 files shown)


projects/angular/src/layout/tabs/tab-link.directive.ts
```diff
@@ -40,7 +40,7 @@ export class ClrTabLink {
 
   constructor(
     public ifActiveService: IfActiveService,
-    @Inject(IF_ACTIVE_ID) private id: number,
+    @Inject(IF_ACTIVE_ID) private _id: number,
```
kevinbuhmann, line 43:
```suggestion
    @Inject(IF_ACTIVE_ID) public readonly id: number,
```

We can just use `readonly` instead of the get accessor wrapper.

kevinbuhmann (changes requested):

projects/angular/src/layout/tabs/tab-link.directive.ts
mivaylo: @kevinbuhmann , makes more sense, thanks, updated

kevinbuhmann (approved):
LGTM, just need to update the public api file.

PR #1316, opened 2024-03-13
fix(combobox): enable screen readers to announce "no results" properly

mivaylo:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-1515

- [ ] Yes
- [x] No

## Other information

Current NVDA voiceover behavior  (notice for "USA" as there are no results it doesn't say "Dialog, list, no results"):
[current behaviour.webm](https://github.com/vmware-clarity/ng-clarity/assets/132376042/3678b55e-4b7a-4c6d-9d4e-021fc6ea87e1)

And with the changes you can hear how NVDA says "List, no results":
[after-changes.webm](https://github.com/vmware-clarity/ng-clarity/assets/132376042/7655a402-68a4-4442-b8d1-c48f1761865d)


projects/angular/src/forms/combobox/options.ts
```diff
@@ -44,7 +44,11 @@ let nbOptionsComponents = 0;
 
     <!-- Rendered if data set is empty -->
     <div *ngIf="emptyOptions">
-      <span class="clr-combobox-options-empty-text">
+      <span
+        class="clr-combobox-options-empty-text"
+        id="no-results-in-combobox"
```
kevinbuhmann, line 49: This must be a unique id because there could be more than on combobox on the page. You can use the `uniqueIdFactory` function for this.

Sorry if this isn't ready for review. It's not a draft, and I don't want this to be missed.

kevinbuhmann (changes requested):

projects/angular/src/forms/combobox/options.ts
mivaylo: Your comment is on point and it's my bad I didn't set this PR to draft.

I was actually thinking of having an id that is concatenation of the id of combobox + '-no-results', and I guess using this function makes more sense. Thanks!

On a different hand though the reality should be that we should never have two open dialogs showing "no results" at the same time (because the first one will close once we start dealing with the second combobox), but reusing id-s is not something we need and is counterintuitive

mivaylo: Updated and this how id-s are now for more than one combobox in the same page:

[multiple-ids.webm](https://github.com/vmware-clarity/ng-clarity/assets/132376042/fe028d68-af49-4223-a1d3-82b2acaaff40)

mivaylo: This is now "clr-id-0" and "clr-id-1", maybe a better naming would be "clr-no-results-id-1" and "clr-no-results-id-2"?

kevinbuhmann: >I was actually thinking of having an id that is concatenation of the id of the combobox + '-no-results'

That would be fine as well.

projects/angular/src/forms/combobox/options.ts
```diff
@@ -44,7 +44,11 @@ let nbOptionsComponents = 0;
 
     <!-- Rendered if data set is empty -->
     <div *ngIf="emptyOptions">
-      <span class="clr-combobox-options-empty-text">
+      <span
+        class="clr-combobox-options-empty-text"
+        id="{{ optionSelectionService.noResultsElementId }}"
```
kevinbuhmann, line 49:
```suggestion
        [id]="optionSelectionService.noResultsElementId"
```

This can be simplified.

kevinbuhmann (approved):

projects/angular/src/forms/combobox/options.ts
mivaylo: Should we move the id of the combobox to be kept in the option-selection.service.ts then so that all id-s are in the state which this service holds? I actually think using the uniqueIdFactory function it is a simpler solution. As for the naming maybe we could have "clr-id-1-no-results" to avoid parsing of the text that the function  returns.

kevinbuhmann (changes requested):
There should be a unit test for this. I am looking into your question about the element id. I will reply in that thread.

projects/angular/src/forms/combobox/options.ts
kevinbuhmann: We don't need the id in the service at all. The `ClrOptions` component has a `optionsId` property. We can just append to that.

kevinbuhmann: We don't need the id in the service at all. The `ClrOptions` component has a `optionsId` property. We can just append to that. See here: 1e7f2c0a4e2b89e60587113290540c2c81021dde

kevinbuhmann:
I think the PR title should reworded to describe the functional change instead of the technical change. I would say 'enable screen readers to announce "no results" properly' or similar.

projects/angular/src/forms/combobox/options.ts
mivaylo: It seems this solution is not full as on Mac OS the screen reader is still not announcing "No results". Once I fix it for Mac OS I will try to reuse your commit, thanks.

kevinbuhmann (approved):

PR #1323, opened 2024-03-20
chore(modal): add visual snapshots for open modal component

kevinbuhmann:
## PR Checklist

- [x] Tests for the changes have been added (for bug fixes / features)
- [N/A] Docs have been added / updated (for bug fixes / features)
- [N/A] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

New visual regression tests.

## What is the current behavior?

No visual regression tests for open modal component.

## What is the new behavior?

Visual regression tests for open modal component in all four sizes.

## Does this PR introduce a breaking change?

No.

PR #1331, opened 2024-03-25
chore(modal): prevent error when story hot reloads

kevinbuhmann:
This fixes a mistake I made in fe2970e9be9c30d8c60088330bb601e3e1390055 (#1323).


valentin-mladenov (approved):
LGTM

Jinnie (approved):
I hit this a couple of times recently :) 
Even had the same fix before I realized it can be avoided by simple refresh and decided to remove it.

kevinbuhmann:
> I hit this a couple of times recently :) Even had the same fix before I realized it can be avoided by simple refresh and decided to remove it.

Yeah, refreshing everytime is annoying though.

PR #1334, opened 2024-03-26
fix(a11y): stepper-completed step does not have accessible notification after completion

andyfeds:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

In the Stepper, when the step is complete, keyboard focus moves to the next step but there's no indication that the previous step is completed by the screen reader.

Issue Number: [CDE-699](https://jira.eng.vmware.com/browse/CDE-699)

## What is the new behavior?
The screen reader will announce that the current step is complete (or failed) and move focus to the next step or stay in the same step in case of an error.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/clarity.api.md
+++ b/projects/angular/clarity.api.md
@@ -318,9 +318,9 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
     // (undocumented)
     panelOpenChange: EventEmitter<boolean>;
     // (undocumented)
-    stepCompleteText(panelNumber: number): string;
+    protected stepCompleteText(panelNumber: number): string;
     // (undocumented)
-    stepErrorText(panelNumber: number): string;
+    protected stepErrorText(panelNumber: number): string;
     // (undocumented)
     togglePanel(): void;
     // (undocumented)
--- a/projects/angular/clarity.api.md
+++ b/projects/angular/clarity.api.md
@@ -320,7 +320,7 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
     // (undocumented)
     stepCompleteText(panelNumber: number): string;
     // (undocumented)
-    stepFailedText(panelNumber: number): string;
+    stepErrorText(panelNumber: number): string;
     // (undocumented)
     togglePanel(): void;
     // (undocumented)
@@ -1006,7 +1006,7 @@ export interface ClrCommonStrings {
     // (undocumented)
     stepComplete: string;
     // (undocumented)
-    stepFailed: string;
+    stepError: string;
     success: string;
     // (undocumented)
     timelineStepCurrent: string;
--- a/projects/angular/clarity.api.md
+++ b/projects/angular/clarity.api.md
@@ -318,6 +318,10 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
     // (undocumented)
     panelOpenChange: EventEmitter<boolean>;
     // (undocumented)
+    stepCompleteText(panelNumber: number): string;
+    // (undocumented)
+    stepFailedText(panelNumber: number): string;
+    // (undocumented)
     togglePanel(): void;
     // (undocumented)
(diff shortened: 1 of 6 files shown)


mivaylo:
Do we intend to backport this to v16? and to v15 if we do it before 28th of March (this is the end of support date for v15)?

mivaylo:
> Do we intend to backport this to v16? and to v15 if we do it before 28th of March (this is the end of support date for v15)?

although we are adding new API - so this seems more like a "feat" rather than a "fix" to me, but it still is not a breaking change, so it could be backported

projects/angular/src/utils/i18n/common-strings.default.ts
```diff
@@ -115,4 +115,8 @@ export const commonStringsDefault: ClrCommonStrings = {
    * Datagrid footer; sr-only text after the number of selected rows.
    */
   selectedRows: 'Selected rows',
+
+  // Accordion/Stepper
+  stepComplete: 'Step {STEP} complete',
+  stepFailed: 'Step {STEP} failed',
```
kevinbuhmann, line 121:
```suggestion
  stepError: 'Error in step {STEP}',
```

I think this would be better.

andyfeds:
> > Do we intend to backport this to v16? and to v15 if we do it before 28th of March (this is the end of support date for v15)?
> 
> although we are adding new API - so this seems more like a "feat" rather than a "fix" to me, but it still is not a breaking change, so it could be backported

Yeah, its not a breaking change, just changes the behaviour for screen readers.

projects/angular/src/utils/i18n/common-strings.default.ts
andyfeds: Yes, makes sense. Made the change.

kevinbuhmann: Please also rename `stepFailed` to `stepError`.

projects/angular/src/accordion/stepper/stepper-panel.spec.ts
```diff
@@ -88,13 +88,13 @@ describe('ClrStep Reactive Forms', () => {
       fixture.detectChanges();
 
       const statusMessage = fixture.nativeElement.querySelector('button .clr-sr-only');
-      expect(statusMessage.innerText.trim()).toBe('Error');
+      expect(statusMessage.innerText.trim()).toBe(`Error in step ${mockStep?.index + 1}`);
```
kevinbuhmann, line 91: The expected value should not be computed at runtime. Please replace the template literal with a static string literal. This applies to both tests.

kevinbuhmann (changes requested):

projects/angular/src/accordion/stepper/stepper-panel.spec.ts
andyfeds: Made these changes

kevinbuhmann (approved):

projects/angular/src/accordion/stepper/stepper-panel.spec.ts
kevinbuhmann: Thanks!

projects/angular/src/utils/i18n/common-strings.default.ts
kevinbuhmann: Thanks!

projects/angular/src/accordion/accordion-panel.ts
```diff
@@ -111,6 +111,14 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
     return `clr-accordion-header-${id}`;
   }
 
+  stepCompleteText(panelNumber: number) {
```
kevinbuhmann, line 114:
```suggestion
  protected stepCompleteText(panelNumber: number) {
```

You can mark these methods `protected` so that they don't show up in public API.

projects/angular/clarity.api.md
```diff
@@ -318,9 +318,9 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
     // (undocumented)
     panelOpenChange: EventEmitter<boolean>;
     // (undocumented)
-    stepCompleteText(panelNumber: number): string;
+    protected stepCompleteText(panelNumber: number): string;
```
andyfeds, line 321: @kevinbuhmann  they still do show up in the public api

kevinbuhmann: Okay, I misremembered something then. Sorry about that. In any case, these don't need to to be `public` anyway.

kevinbuhmann (approved):

PR #1342, opened 2024-03-27
chore(storybook): remove dead `setupStorybook` function

kevinbuhmann:
This function was used for the old non-CSFv3 stories.

CDE-1871


valentin-mladenov (approved):
Nice! LGTM

PR #1346, opened 2024-04-04
feat(datagrid): optimise layout rendering

dtsanevmw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

- When computing header widths we get each header boundingClientRect and update the width separately which creates multiple reflows and slow us down.
- Cells and Headers are listening for changes and holding column states.
- Calculating userDefinedWidth triggers reflow each time when accessing getBoundingClientRect.
- Expandable animation cycle is triggered as being changed initially which trigger setStyle.

Issue Number: CDE-1743

## What is the new behavior?

- All the widths are batched before traversing trough headers and their width being set.
- Main renderer is handling the state listening and pass the changes to cells and headers to update.
- We clone the target element before we calculate boundingClientRect outside the grid so we don't trigger reflow of the whole grid and just receive measurements for that specific element.
- We set the expanded state to null to skip the animation cycle on initial load.

- [ ] Yes
- [X] No


projects/angular/src/utils/dom-adapter/dom-adapter.ts
```diff
@@ -14,10 +14,16 @@ import { Injectable } from '@angular/core';
...
+    We clone the element and take its measurements from outside the grid
+    so we don't trigger reflow for the whole datagrid.
+  */
   userDefinedWidth(element: HTMLElement): number {
-    element.classList.add('datagrid-cell-width-zero');
-    const userDefinedWidth = this.clientRect(element).width;
-    element.classList.remove('datagrid-cell-width-zero');
+    const clonedElement = element.cloneNode(true) as HTMLElement;
+    clonedElement.classList.add('datagrid-cell-width-zero');
+    document.body.appendChild(clonedElement);
```
valentin-mladenov, line 24: Pitfall: https://developer.mozilla.org/en-US/docs/Web/API/Node/cloneNode
`CloneNode` possible `id` duplication, possibly the `name` attributes as well.
Change the cloned node ID to something else. Add `-clone` at the end of it.
```suggestion
    clonedElement.id = clonedElement.id + '-clone';
    document.body.appendChild(clonedElement);
```

projects/angular/src/data/datagrid/render/header-renderer.ts
```diff
@@ -63,20 +60,29 @@ export class DatagridHeaderRenderer implements OnDestroy {
...
+      this.widthSet = true;
+      this.autoSet = true;
+    }
+  }
+
+  setHidden(state: ColumnState) {
+    if (state.hidden) {
+      this.renderer.addClass(this.el.nativeElement, HIDDEN_COLUMN_CLASS);
+    } else {
+      this.renderer.removeClass(this.el.nativeElement, HIDDEN_COLUMN_CLASS);
```
valentin-mladenov, line 85: Is there a reason behind the move of `setWidth` and `setHidden` from their original places?

valentin-mladenov (commented):
Check my comments.

projects/angular/src/data/datagrid/render/header-renderer.ts
dtsanevmw: They are public now so we can access them from main-renderer and we have a specific order for class members otherwise the lint will blow up.

projects/angular/src/utils/dom-adapter/dom-adapter.ts
dtsanevmw: We are removing the clone instantly so shouldn't be a problem

valentin-mladenov: I pointed it out to be 100% sure in uniqueness of the `id`. We don't know how datagrid may be used and is low effort to update the `id`, unless it leads to some performance degradation.
There might be very slim corner case that might lead to some unexpected results from not changing the `id`.

projects/angular/src/data/datagrid/datagrid.spec.ts
```diff
@@ -1261,13 +1261,6 @@ export default function (): void {
         expect(calculationColumns.length).toBe(2);
       });
 
-      it('moves the rows into the calculation container', function () {
```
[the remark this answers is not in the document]
dtsanevmw: Reverted it.

projects/angular/src/data/datagrid/render/main-renderer.ts
```diff
@@ -213,6 +218,47 @@ export class DatagridMainRenderer implements AfterContentInit, AfterViewInit, Af
...
+          default:
+            break;
+        }
+      });
+    }
+  }
+
+  private listenForColumnChanges() {
+    this.columnsService.columns.forEach((column, index) => {
+      this.columnsService.emitStateChange(column, { changes: [], columnIndex: index } as ColumnStateDiff);
```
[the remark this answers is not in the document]
dtsanevmw: I checked and we need it due to the service applying the new state as 'diff' on top of the old state so if there are no changes it will keep the old ones. So I'll keep it.

Jinnie (approved):

projects/angular/src/data/datagrid/render/main-renderer.ts
kevinbuhmann, line 252:
```suggestion
      this.columnsService.emitStateChange(column, { changes: [], columnIndex: index });
```

This type assertion is unnecessary.

dtsanevmw: Thanks, fixed in [cb1c20d](https://github.com/vmware-clarity/ng-clarity/pull/1346/commits/cb1c20da71a5cca6b781799d9b36f8d5fb2b80c6)

PR #1376, opened 2024-04-25
feat(wizard): add back inline wizards in 17.x

kevinbuhmann:
This is a port of 8a623cb66aa5cad4e0870801f995ddc3a41c9afe (#1068) to 17.x.

This reverts commit 7470769480b843dc332adbd25f67624ce4979e56 (#419), removes the deprecated comments, and adds a comment explaining when inline wizards are used.

Inline wizards are used by vSphere plugins to display a wizard inside an iframe that is inside a modal.

This was previously landed in 15.x and 16.x, but I missed 17.x.

CDE-154


valentin-mladenov (approved):
LGTM

PR #1381, opened 2024-05-02
feat(datepicker): add support for date range and different date formats

venkateshr06:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-1846

- [ ] Yes
- [ ] No

### DIFF
--- a/projects/angular/src/layout/nav/_links.clarity.scss
+++ b/projects/angular/src/layout/nav/_links.clarity.scss
@@ -9,37 +9,42 @@
 
 @include mixins.exports('links.clarity') {
   a {
-    &:link {
-      /*
+    /*
       * Due to the higher specificity of `a:link` over `.btn`, `.nav-link`, `.dropdown-item`, `.alert-action` and `badge`
       * it overrides the styles of both which is not required.
       * Also `.btn, .nav-link, .dropdown-item, .alert-action, .badge` doesn't need these styles, so we don't add them in these cases.
       */
-      &:not(.btn, .nav-link, .dropdown-item, .alert-action, .badge) {
-        @include mixins.generate-typography-token('LINK-13');
+    &:not(.btn, .nav-link, .dropdown-item, .alert-action, .badge, .clr-treenode-link) {
+      @include mixins.generate-typography-token('LINK-13');
+      &:link {
+        color: nav-variables.$clr-link-color;
+        &:hover {
+          color: nav-variables.$clr-link-hover-color;
+          cds-icon,
+          clr-icon {
+            color: nav-variables.$clr-link-hover-color;
+          }
+        }
+        &:active {
+          color: nav-variables.$clr-link-active-color;
+          cds-icon,
+          clr-icon {
+            color: nav-variables.$clr-link-active-color;
+          }
+        }
       }
-    }
-    &:active {
-      cds-icon,
-      clr-icon {
-        color: nav-variables.$clr-link-active-color;
-      }
-    }
-    &:hover {
-      cds-icon,
-      clr-icon {
-        color: nav-variables.$clr-link-hover-color;
-      }
-    }
-    &:visited {
-      cds-icon,
-      clr-icon {
+      &:visited {
         color: nav-variables.$clr-link-visited-color;
-      }
-      &:hover {
         cds-icon,
         clr-icon {
+          color: nav-variables.$clr-link-visited-color;
+        }
+        &:hover {
           color: nav-variables.$clr-link-visited-color-hover;
+          cds-icon,
+          clr-icon {
+            color: nav-variables.$clr-link-visited-color-hover;
+          }
         }
       }
     }
--- a/projects/angular/src/utils/_reboot.clarity.scss
+++ b/projects/angular/src/utils/_reboot.clarity.scss
@@ -408,24 +408,4 @@
   table {
     border-spacing: 0;
   }
-
-  a:link {
-    color: nav-variables.$clr-link-color;
-  }
-
-  a:visited {
-    color: nav-variables.$clr-link-visited-color;
-
-    &:hover {
-      color: nav-variables.$clr-link-visited-color-hover;
-    }
-  }
-
-  a:hover {
-    color: nav-variables.$clr-link-hover-color;
-  }
-
-  a:active {
-    color: nav-variables.$clr-link-active-color;
-  }
 }


projects/angular/src/forms/common/abstract-container.ts
```diff
@@ -27,6 +27,7 @@ export abstract class ClrAbstractContainer implements DynamicWrapper, OnDestroy,
   @ContentChild(ClrControlHelper) controlHelperComponent: ClrControlHelper;
 
   control: NgControl;
+  secondaryControl: NgControl;
```
Jinnie, line 30: Introducing a secondary control here is a bit controversial to me
1) we're adding case-specific code to a common parent. It will be only used in the date-range picker and nowhere else.
2) the concept of secondary control itself is a bit unusual. It leads to having a component with 2 roots. Why don't we add a common wrapper to the 2 inputs and add all range-specific code there?
3) it requires changes to many other core classes, which poses a high risk of regressions at seemingly unrelated components.

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
```diff
@@ -26,56 +26,139 @@
...
+
+  // }
+
   &:active {
     color: datepicker-variables.$clr-calendar-btn-active-color;
   }
 }
 
 @include mixins.exports('datepicker.clarity') {
+  // .datepicker {
```
Jinnie, line 97: left-over comments

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
```diff
@@ -26,56 +26,139 @@
...
   .datepicker {
     display: flex;
-    flex-direction: column;
-
-    //Dimensions
-    margin-top: datepicker-variables.$clr-calendar-margin-top;
-    padding: datepicker-variables.$clr-calendar-padding;
+    flex-direction: row;
+    margin-top: var(--cds-global-space-4);
+    // padding: var(--cds-global-space-7);
```
Jinnie, line 119: left-over comment

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
```diff
@@ -26,56 +26,139 @@
...
-    border: tokens.$cds-alias-object-border-width-100 solid datepicker-variables.$clr-calendar-border-color;
-    border-radius: tokens.$cds-alias-object-border-radius-100;
-    box-shadow: tokens.$cds-alias-object-shadow-100;
-    overflow: hidden;
-    z-index: map.get(variables.$clr-layers, dropdown-menu);
+    background: var(--clr-calendar-background-color);
+    border: var(--cds-alias-object-border-width-100) solid var(--clr-calendar-border-color);
+    border-radius: var(--cds-alias-object-border-radius-100);
+    box-shadow: var(--cds-alias-object-shadow-100);
+    // overflow: hidden;
```
Jinnie, line 126: left-over comment

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
```diff
@@ -26,56 +26,139 @@
...
+    .datepicker-view-manager {
+      padding: var(--cds-global-space-7);
+    }
+    &.has-range-option {
+      width: calc(datepicker-variables.$clr-calendar-width + datepicker-variables.$clr-calendar-options-width);
+    }
+  }
+  .options-wrapper {
+    width: datepicker-variables.$clr-calendar-options-width;
+    //  border-right: 1px solid tokens.$cds-alias-object-border-color;
```
Jinnie, line 137: left-over comment

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
```diff
@@ -26,56 +26,139 @@
...
+      border-radius: 0;
+    }
+    &:not(.is-excluded) {
+      background: datepicker-variables.$clr-calendar-in-range-cell-border-color;
+    }
   }
-
   &:focus {
-    //Only because we have defined a background color for focus above
+    border: 1px solid datepicker-variables.$clr-calendar-btn-focus-border-color !important;
```
Jinnie, line 46: 1px needs to get tokenized

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
```diff
@@ -26,56 +26,139 @@
...
-
   &:active {
     background: datepicker-variables.$clr-calendar-btn-active-color-bg;
   }
 }
 
+@mixin generate-current-date-style() {
+  &.is-today {
+    color: datepicker-variables.$clr-calendar-today-date-cell-color;
+    border: 1px solid datepicker-variables.$clr-calendar-today-date-cell-border-color;
```
Jinnie, line 57: 1px needs to get tokenized

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
```diff
@@ -26,56 +26,139 @@
...
+}
+
 @mixin generate-calendar-hover-active-color-style() {
-  &:hover,
-  &:focus {
+  &:hover {
     color: datepicker-variables.$clr-calendar-btn-hover-focus-color-text;
   }
 
+  // &:focus {
```
Jinnie, line 87: left-over comment

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
```diff
@@ -26,56 +26,139 @@
...
-    border-radius: tokens.$cds-alias-object-border-radius-100;
-    box-shadow: tokens.$cds-alias-object-shadow-100;
-    overflow: hidden;
-    z-index: map.get(variables.$clr-layers, dropdown-menu);
+    background: var(--clr-calendar-background-color);
+    border: var(--cds-alias-object-border-width-100) solid var(--clr-calendar-border-color);
+    border-radius: var(--cds-alias-object-border-radius-100);
+    box-shadow: var(--cds-alias-object-shadow-100);
+    // overflow: hidden;
+    z-index: 1050;
```
Jinnie, line 127: better add this to the z-indices map and retrieve it from there

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
```diff
@@ -26,56 +26,139 @@
...
+    flex-shrink: 0;
+    border-top-left-radius: var(--cds-alias-object-border-radius-100);
+    border-bottom-left-radius: var(--cds-alias-object-border-radius-100);
+    .range-option {
+      padding: var(--cds-global-space-5);
+      padding-left: var(--cds-global-space-7);
+      color: datepicker-variables.$clr-calendar-option-text-color;
+      cursor: pointer;
+      &.active {
+        border-left: 3px solid datepicker-variables.$clr-calendar-active-option-border-color;
```
Jinnie, line 148: 3px needs to get tokenized

projects/angular/src/forms/datepicker/_datepicker.clarity.scss
```diff
@@ -97,6 +180,15 @@
     width: datepicker-variables.$clr-switcher-width;
   }
 
+  .year-view-switcher {
+    flex: 0 0 auto;
+    align-self: center;
+  }
+
+  .calendar-header.in-monthpicker {
+    margin-bottom: 8px;
```
Jinnie, line 189: tokenize please

projects/angular/src/forms/datepicker/_properties.datepicker.scss
```diff
@@ -13,9 +13,14 @@
       --clr-calendar-background-color: #{tokens.$cds-alias-object-container-background};
       --clr-calendar-border-color: #{tokens.$cds-alias-object-border-color-shade};
 
+      --clr-calendar-option-text-color: #{tokens.$cds-alias-object-interaction-color};
+      // --clr-calendar-options-background-color: #f3f5f7; // #{tokens.$cds-alias-object-vertical-navigation-background-color};
```
Jinnie, line 17: left-over comment

projects/angular/src/forms/datepicker/_variables.datepicker.scss
```diff
@@ -18,6 +18,7 @@ $clr-calendar-width: calc(
   calc(7 * #{tokens.$cds-global-space-11}) + calc(2 * #{tokens.$cds-global-space-7}) +
     calc(2 * #{tokens.$cds-alias-object-border-width-100})
 );
+$clr-calendar-options-width: 150px;
```
Jinnie, line 21: handle through baselinePx

projects/angular/src/forms/datepicker/date-input-base.spec.ts
```diff
@@ -202,25 +202,25 @@ export default function () {
           expect(context.clarityDirective.inputType).toBe('date');
         });
 
-        it('sets the selectedDay if the value of the input is valid', () => {
-          const testEl = document.createElement('INPUT') as HTMLInputElement;
-          testEl.value = '01/02/2015';
+        // it('sets the selectedDay if the value of the input is valid', () => {
```
Jinnie, line 205: remove or make these tests work??

projects/angular/src/forms/datepicker/date-input-base.spec.ts
```diff
@@ -291,14 +291,14 @@ export default function () {
           expect(context.clarityElement.type).toBe('date');
         });
 
-        it('listens to the input change events', () => {
-          spyOn(context.clarityDirective, 'onValueChange');
+        // it('listens to the input change events', () => {
```
Jinnie, line 294: remove or make work??

projects/angular/src/forms/datepicker/date-input-base.ts
```diff
@@ -19,98 +18,79 @@ import {
...
 import { WrappedFormControl } from '../common/wrapped-control';
 import { ClrDateContainer } from './date-container';
-import { DayModel } from './model/day.model';
 import { DateFormControlService } from './providers/date-form-control.service';
 import { DateIOService } from './providers/date-io.service';
 import { DateNavigationService } from './providers/date-navigation.service';
 import { DatepickerEnabledService } from './providers/datepicker-enabled.service';
 import { DatepickerFocusService } from './providers/datepicker-focus.service';
-import { datesAreEqual } from './utils/date-utils';
+// import { datesAreEqual } from './utils/date-utils';
```
Jinnie, line 38: leftover comment

projects/angular/src/forms/datepicker/date-input-base.ts
```diff
@@ -154,7 +128,7 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
     // I need the renderer to set the value property on the input to make sure that if the user has supplied a Date
     // input object, we reflect it with the right date on the input field using the IO service. I am not sure if
     // these are major issues or not but just noting them down here.
-    this.processInitialInputs();
+    // this.processInitialInputs();
```
Jinnie, line 131: leftover comment

projects/angular/src/forms/datepicker/date-input-base.ts
```diff
@@ -250,12 +162,14 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
       } else {
         this.renderer.setProperty(this.el.nativeElement, 'value', dateString);
       }
+      this.validateDateRange();
     } else {
       this.renderer.setProperty(this.el.nativeElement, 'value', '');
+      // this.control?.control?.setValue(null, { emitEvent: false });
```
Jinnie, line 168: leftover comment

projects/angular/src/forms/datepicker/model/calendar-view.model.ts
```diff
@@ -158,9 +177,14 @@ export class CalendarViewModel {
    * Initialize the selected day if the day is in the calendar.
    */
   private initializeSelectedDay(): void {
-    if (this.selectedDay && this.isDayInCalendarView(this.selectedDay)) {
-      this.currMonthDayViews[this.selectedDay.date - 1].isSelected = true;
-    }
+    this.setSelectedDay(this.selectedDay, true);
+    this.setSelectedDay(this.selectedEndDay, true);
+    // if (this.selectedDay && this.isDayInCalendarView(this.selectedDay)) {
```
Jinnie, line 182: leftover comments

projects/angular/src/forms/datepicker/providers/date-io.service.ts
```diff
@@ -73,22 +78,73 @@ export class DateIOService {
...
-      const dateStr: string = dateNo > 9 ? dateNo.toString() : '0' + dateNo;
-      const monthStr: string = monthNo > 9 ? monthNo.toString() : '0' + monthNo;
-      if (this.localeDisplayFormat === LITTLE_ENDIAN) {
-        return dateStr + this.delimiters[0] + monthStr + this.delimiters[1] + date.getFullYear();
-      } else if (this.localeDisplayFormat === MIDDLE_ENDIAN) {
-        return monthStr + this.delimiters[0] + dateStr + this.delimiters[1] + date.getFullYear();
-      } else {
-        return date.getFullYear() + this.delimiters[0] + monthStr + this.delimiters[1] + dateStr;
-      }
+      // const dateNo: number = date.getDate();
```
Jinnie, line 136: left over comments

projects/angular/src/forms/datepicker/providers/date-io.service.ts
```diff
@@ -97,26 +153,31 @@ export class DateIOService {
...
-    } else if (this.localeDisplayFormat === MIDDLE_ENDIAN) {
-      // firstPart is month && secondPart is date
-      return this.validateAndGetDate(thirdPart, firstPart, secondPart);
-    } else {
-      // secondPart is month && thirdPart is date
-      return this.validateAndGetDate(firstPart, secondPart, thirdPart);
-    }
+    const dateParts: string[] = extractDateParts(date, this.localeDisplayFormat.format);
+    return this.validateAndGetDate(dateParts[0], dateParts[1], dateParts[2]);
+    // const dateParts: string[] = date.match(USER_INPUT_REGEX);
```
Jinnie, line 158: leftover comments

Jinnie (commented):
First, thanks for the great amount of work on this exciting feature and all the attention to fine details and various use-cases!
Reviewing this may take some more iterations, but right now I have some high level questions and concerns that we may further discuss offline.

My goals, as reviewer, are for now the following:
- Ensure we're not breaking current behavior/s and provide incremental API changes where possible.
- Make sure we are providing a solid and well structured architecture of the various sub-components.
- Validate agains the design requirements and guidelines.

First off, my biggest concern is that the implementation diverges too much from the initial specs that were presented to the team. This leads to unexpected scope explosion and features and architecture that were not synced on planning phase.

The changes on the abstract and other common classes bring concerns (explained in more detail in the lower comments). They bring (seemingly) unnecessary specificity to a high hierarchy level, and drag along changes in event model, subscriptions, validation and more.

The API and naming of the existing date picker has been change witch causes breaking changes, which we would prefer to avoid.

As we talked on the demos, some of the use-cases are not realistic - like the month range (without year). For example - if you select Feb to Sept it may make some sense, but what happens if you select Sept to Feb?

A large volume of commented code is left. I added comments to most, but not all cases, please check. Especially in tests, where it's not clear if these tests are obsolete, or just temporarily commented and still needs to get satisfied.

projects/angular/src/forms/common/abstract-container.ts
naga-zone: Looks like the Container is tightly coupled with control element. As per existing code, only one control is allowed for each container. Making it 2 different containers(for primary and secondary control) will create 2 instances, and both controls cant talk to each other, which is an issue for date range picker. I agree that this will cause additional regression test for other components using the Abstract Container. But, we need to gracefully handle the second control scenario for all other components.

projects/demo/src/app/datepicker/datepicker-date-input-explicit-wrapper.html
```diff
@@ -8,7 +8,7 @@
 <div clrForm>
   <h6>With [clrDate]</h6>
   <p></p>
-  <clr-date-container>
+  <clr-date-container [dateFormat]="'yyyy/MM'">
```
Jinnie, line 11: Is this testing something specific here? We'd better keep format overrides in their own page.

projects/demo/src/app/datepicker/datepicker-in-reactive-forms.ts
```diff
@@ -14,9 +14,32 @@ import { FormControl, FormGroup } from '@angular/forms';
...
 })
 export class DatepickerInReactiveForms {
-  dateForm = new FormGroup({ date: new FormControl('03/05/2018'), name: new FormControl('Jane') });
+  dateForm = new FormGroup({
+    date: new FormControl('03/05/2018'),
+    startDate: new FormControl(),
+    endDate: new FormControl(),
+  });
+  // minDate = '02/02/2021';
+  // maxDate = '02/24/2029';
```
Jinnie: Leftover comments.

projects/demo/src/app/datepicker/datepicker-in-reactive-forms.ts
```diff
@@ -14,9 +14,32 @@ import { FormControl, FormGroup } from '@angular/forms';
...
+
+  endDateChanged(date: Date) {
+    console.log(date);
+  }
+
+  ngOnInit() {
+    // this.minDate = '2024-02-05';
+    // this.maxDate = '2024-02-22';
+    // this.dateForm.get('date').patchValue('02/02/2024')
+  }
```
Jinnie: More leftover comments in an empty ngOnInit()

projects/demo/src/app/datepicker/datepicker.demo.html
```diff
@@ -23,6 +23,18 @@ <h2>Datepicker</h2>
...
     <li><a [routerLink]="['./locale-data']">Locale Data</a></li>
+    <li>
+      <a [routerLink]="['./date-range-picker']"
+        >Date Range Picker
+        <span class="badge badge-success"> New </span>
+      </a>
+    </li>
+    <li>
+      <a [routerLink]="['./enhanced-date-picker']"
+        >Enhanced Date Picker
```
Jinnie, line 34: Maybe we should name it "Custom Date Format" or something in that direction. "Enhanced" is too broad.

projects/demo/src/app/datepicker/datepicker.demo.scss
```diff
@@ -22,3 +22,59 @@
 .clr-example-input {
   display: inline-block;
 }
+
+.date-range-section {
+  margin: 0 20px;
```
Jinnie, line 27: This file also needs value tokenization and baselining. 
We discussed the details offline, so should be no problem.

Jinnie (commented):
It would really be nice to have the dateFormat related work in a separate PR.
As it fights with default browser localization and provider-based localization we don't even know if we want it, or when we want it released and now it's tied to the range component.

And some more observations and first use impressions:
- Dragging a range, as opposed to sequential clicking, does not work. Not sure this is valid use case, but I hit it several times before understanding what happens. May need to sync with UX.
- Trying to only change the end date on already existing value resets the start date. UX comment needed.
- Sequentially clicking backwards, from later to earlier date, just resets the start date. UX comment needed.
- The dash (-) separating start and end date is too much closer to end date making it look like negative value instead of range. I know the technical reasons for this, but some reiteration with Design might help here.
- Nit: after selecting a range option, the next time you open the picker it shows "Custom range". Not critical, but would be nice if we could resolve back the range option and pre-select it if it's matching.
- Nit: If we violate validation restrictions in both fields we're receiving just one error message.

projects/demo/src/app/datepicker/datepicker.demo.scss
Jinnie: It's demo code. My previous comment may be disregarded.

kevinbuhmann:
>It would really be nice to have the dateFormat related work in a separate PR.
>As it fights with default browser localization and provider-based localization we don't even know if we want it, or when we want it released and now it's tied to the range component.

I agree. These two feature need to be separated. There is too much in this PR and these things need to reviewed separately.

projects/angular/src/forms/common/if-control-state/if-control-state.service.ts
```diff
@@ -34,11 +35,26 @@ export class IfControlStateService implements OnDestroy {
           // Subscribe to the status change events, only after touched
           // and emit the control
           this.subscriptions.push(
-            this.control.statusChanges.subscribe(() => {
+            this.control.statusChanges?.subscribe(() => {
```
Jinnie, line 38: Don't we need to subscribe to ngControlService.controlChanges here too, same as we do with the additional changes below?
We're only subscribing to statusChanges of the primary control, but we're not reestablishing the status subscriptions if control gets changed.

projects/angular/src/forms/common/if-control-state/if-control-state.service.ts
```diff
@@ -34,11 +35,26 @@ export class IfControlStateService implements OnDestroy {
...
+      }),
+
+      this.ngControlService.additionalControlChanges.subscribe((controls: NgControl[]) => {
+        if (controls) {
+          this.additionalControls = controls;
+          // Subscribe to the status change events, only after touched
+          // and emit the control
+          this.additionalControls?.forEach((control: NgControl) => {
+            this.subscriptions.push(
+              control?.statusChanges?.subscribe(() => {
```
Jinnie, line 52: If additional controls get changed, won't we need to unsubscribe from the previous statusChanges subscriptions before we register the new ones?
This may apply for the primary control subscriptions too, if my previous comment gets resolved.

projects/angular/src/forms/common/if-control-state/if-error.ts
```diff
@@ -35,27 +35,30 @@ export class ClrIfError extends AbstractIfState {
    * @param state CONTROL_STATE
    */
   protected override handleState(state: CONTROL_STATE) {
-    if (this.error && this.control) {
+    if (this.error && this.control && !!this.control.invalid) {
       this.displayError(this.control.hasError(this.error));
+    } else if (this.error && !!this.additionalControls?.length) {
+      const invalidControl = this.additionalControls?.filter(control => control.hasError(this.error))[0];
```
Jinnie, line 41: Is it impossible that the filter will return 0 elements and trying to access [0] will throw an error?
Or we rely on the assumption that some control set this.error and this is not the main control, but _should_ be some of the additionalControls?

projects/angular/src/forms/common/wrapped-control.ts
```diff
@@ -131,14 +128,25 @@ export class WrappedFormControl<W extends DynamicWrapper> implements OnInit, DoC
...
     }
   }
 
   ngDoCheck() {
-    if (this.differ) {
-      const changes = this.differ.diff(this.ngControl);
+    this.triggerDoCheck(this.differ, this.ngControl);
+  }
+
+  triggerDoCheck(differ, ngControl) {
```
Jinnie, line 147: Why do we need this as separate public method? I don't see it used anywhere else.

projects/angular/src/forms/common/if-control-state/if-control-state.service.ts
```diff
@@ -34,11 +35,26 @@ export class IfControlStateService implements OnDestroy {
...
         }
+      }),
+
+      this.ngControlService.additionalControlChanges.subscribe((controls: NgControl[]) => {
+        if (controls) {
+          this.additionalControls = controls;
+          // Subscribe to the status change events, only after touched
+          // and emit the control
+          this.additionalControls?.forEach((control: NgControl) => {
+            this.subscriptions.push(
```
Jinnie, line 51: How do we cleanup old subscriptions if this change occurs?

projects/angular/src/forms/common/wrapped-control.ts
```diff
@@ -131,14 +128,25 @@ export class WrappedFormControl<W extends DynamicWrapper> implements OnInit, DoC
...
-    if (this.ngControlService && this.ngControl) {
-      this.ngControlService.setControl(this.ngControl);
+    if (this.ngControlService && this._ngControl) {
+      if (!this.ngControlService.getControl()) {
+        this.ngControl = this._ngControl;
+        this.ngControlService.setControl(this.ngControl);
+        this.differ = this.differs.find(this._ngControl).create();
+      } else {
+        this.ngControl = this.ngControlService.getControl();
+        this.ngControlService.addAdditionalControl(this._ngControl);
```
Jinnie, line 138: I am getting a bit confused how this works with multiple additional controls. It's probably only me, but I'd be happy to get some more details about it, thanks!

projects/angular/src/forms/datepicker/date-input-base.ts
```diff
@@ -264,33 +175,57 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
...
-      this.previousDateChange = null;
+  protected datepickerHasFormControl() {
+    return !!this.control;
+  }
+
+  /**
+   * Incase of date range error, both start & end date field valdiation has to be triggered
+   * if either of the field gets updated
+   */
+  protected validateDateRange() {
```
Jinnie, line 186: I believe this belongs to the range specific children. They may have another common parent, that extends this class, to avoid code duplication between start and end, though.

Jinnie (commented):
A lot of code is duplicated between the date-range-start/end-input directives.

I would also like to have some UX input whether next/prev month grayed-out dates should be excluded from the range visualization, if it extends between adjacent months:
Screenshot 2024-06-13 at 14 26 25

projects/angular/src/forms/common/if-control-state/if-error.ts
venkateshr06: array[0] will return undefined only, It wont be returning any error. Even in `strict mode` the behavior is same. Its already gracefully handled in subsequent methods (displayError()) where this undefined value is passed.

projects/angular/src/forms/common/if-control-state/if-control-state.service.ts
venkateshr06: 'ngControlService.controlChanges' already there in line#32. Code is already available, request you to check complete file.

projects/angular/src/forms/common/if-control-state/if-control-state.service.ts
venkateshr06: Set control is going to happen only once per control if I am not wrong. I doubt that we can change the control i mean replace the existing control with new one. It will happen when container itself is completely destroyed and recreated. Also this is an existing implementation which can be found in other file as well(select-container).

Jinnie: Do you think it might become more explicit if we add take(1)?
We won't even need to add it to the subscriptions array if we do.  We're using the take(1) approach in several components and services already.

projects/angular/src/forms/datepicker/date-input-base.ts
venkateshr06: Yes, since both start date and end date component are using this kept in date-input-base component. Anyways I have wrapped it inside condition to check if its date range picker.

projects/angular/src/forms/common/wrapped-control.ts
venkateshr06: Basically its multi level inheritance used here. WrappedControl is the base class for DateInputBase which is inturn base component for Single, Start & End datepicker components. DateRangeStartInput will be considered as Primary Control & DateRangeEndInput will be considered as additional Control (since primary control is already set i.e. DateRangeStart)

projects/angular/src/forms/common/wrapped-control.ts
venkateshr06: Basically added it as a separate methods in order to support Multiple Form controls(Primary & Additional Controls). Removed Purposefully Keyvalue differ for additional Controls now its added back.

projects/angular/src/forms/common/if-control-state/if-control-state.service.ts
venkateshr06: Totally agree, but since its an existing approach in all other files didn't change the approach. Need to do some research why it was added in this way. Hence keeping it in subscriptions mode as of now.

projects/angular/src/forms/common/wrapped-control.ts
Jinnie: I think I got it, thanks!

projects/angular/src/forms/common/wrapped-control.ts
Jinnie: It's only called from the ngDoCheck method above. If you want to keep it for better code structure and readability, I am okay with this, but it can be private.

projects/angular/src/forms/common/if-control-state/if-control-state.service.ts
Jinnie: My concern is that we're adding more subscriptions inside a subscribe block. This may lead to leaks. It should be okay to add take(1) here. It will make the code safer if it really only gets called once, or will throw explicit error if it unexpectedly gets called more than once, and save us tedious performance/memory-leak debugging.

PR #1392, opened 2024-05-09
fix(a11y): stack-view - no proper heading hierarchy maintained in the page

andyfeds:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

No proper heading hierarchy maintained in the page in case of a stack-block component, as the role is fixed as "heading".
If a user wants to change or omit the role, they can't, because our current implementation forces the heading role and level.
Issue Number: [CDE-1634](https://jira.eng.vmware.com/browse/CDE-1634)

## What is the new behavior?
The below changes are made as mentioned in the JIRA ticket:
- If no role is provided, we add role="heading" by default and add the aria-level based on common component usage.
- The user can change the role or the aria level (if the role is not "heading", we omit the aria-level attribute). 
- If the user decides they don't want the "role" attribute at all, they can pass in an undefined value "undefined" and we won't add the role or aria-level attribute.

A new input `clrSbAriaRole` is added to the stack-block, so the user can pass/configure the aria role.

- [x] Yes
- [ ] No


sandeepkulki (approved):
Changes are approved

sandeepkulki (approved):

andyfeds:
Contains breaking changes, will be part of v18 release.

PR #1403, opened 2024-05-28
fix(modal): fall back to modal id for dialog `aria-labelledby` attribute

kevinbuhmann:
## PR Checklist

## PR Type

Bugfix

If you set `[clrModalLabelledById]` to an empty string, the `aria-labelledby` attribute of the modal dialog element is set to an empty value.

This behavior can be seen in the modal stories since the input is defaulted to an empty string. We can't default it to the modal id because that is not known until the component is instantiated.

Issue Number: CDE-2050

## What is the new behavior?

If you set `[clrModalLabelledById]` to an empty string, the `aria-labelledby` attribute of the modal dialog element is set to the modal id.

No.

### DIFF
--- a/projects/angular/src/modal/modal.html
+++ b/projects/angular/src/modal/modal.html
@@ -19,7 +19,7 @@
     role="dialog"
     aria-modal="true"
     [attr.aria-hidden]="!_open"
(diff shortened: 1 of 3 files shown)


Jinnie (approved):
Good catch

PR #1405, opened 2024-05-29
fix(a11y): datagrid detail pane has role dialog with no aria label

andyfeds:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`aria-describedby` is used on the datagrid detail pane instead of `aria-labelledby` 

Issue Number: [CDE-645](https://jira.eng.vmware.com/browse/CDE-645)

## What is the new behavior?

- By default if the user does not add anything: `aria-labelledby` will be added by the library with the id of the modal title.
- The user can set additional `aria-labelledby` (using the [clrDetailAriaLabelledBy] input) which could have a value referencing other text on the page , it can have multiple space separated id's. In this case the default title id will be added along with the user defined IDs.
- The user can add `aria-label` (using [clrDetailAriaLabel] input), in this case the aria-label provided will be used. 
- If both `aria-label` & `aria-labelledby` is added by the user, then only `aria-labelledby` will be added to the modal.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/data/datagrid/datagrid-detail.spec.ts
+++ b/projects/angular/src/data/datagrid/datagrid-detail.spec.ts
@@ -6,7 +6,7 @@
  */
 
 import { Component } from '@angular/core';
-import { async } from '@angular/core/testing';
+import { waitForAsync } from '@angular/core/testing';
 
 import { ClrDatagridDetail } from './datagrid-detail';
 import { TestContext } from './helpers.spec';
@@ -34,12 +34,13 @@ export default function (): void {
     });
 
     describe('View', function () {
-      let context: TestContext<ClrDatagridDetail, FullTest>;
+      let context: TestContext<ClrDatagridDetail, DataGridPaneTestComponent>;
       let detailService: DetailService;
 
       beforeEach(function () {
-        context = this.create(ClrDatagridDetail, FullTest, [DetailService]);
+        context = this.create(ClrDatagridDetail, DataGridPaneTestComponent, [DetailService]);
         detailService = context.getClarityProvider(DetailService);
+        detailService.id = 'clr-id-1';
         context.detectChanges();
       });
 
@@ -53,7 +54,7 @@ export default function (): void {
         expect(context.clarityElement.innerHTML).not.toContain(content);
       });
 
-      it('hides content with the esc key', async(() => {
+      it('hides content with the esc key', waitForAsync(() => {
         spyOn(detailService, 'close');
         detailService.open({});
         context.detectChanges();
@@ -81,6 +82,44 @@ export default function (): void {
         expect(messages[0].innerText).toBe('Start of row details');
         expect(messages[1].innerText).toBe('End of row details');
       });
+
+      it('should have aria-labelledby attribute by default', function () {
+        detailService.open({});
+        context.detectChanges();
+        const detailDialog = context.clarityElement.querySelector('[role="dialog"]');
+        const headerTitle = detailDialog.querySelector('.datagrid-detail-header-title');
+        expect(detailDialog.hasAttribute('aria-labelledby')).toEqual(true);
+        expect(detailDialog.getAttribute('aria-labelledby')).toBe(headerTitle.getAttribute('id'));
+      });
+
+      it('should allow a custom id value for aria-labelledby along with the default id', function () {
+        context.testComponent.ariaLabelledBy = 'customId';
+        detailService.open({});
+        context.detectChanges();
+        const detailDialog = context.clarityElement.querySelector('[role="dialog"]');
+        const headerTitle = detailDialog.querySelector('.datagrid-detail-header-title');
+        expect(detailDialog.getAttribute('aria-labelledby')).toBe(headerTitle.getAttribute('id') + ' customId');
+      });
+
+      it('should add only aria-label attribute if clrDetailAriaLabel input is available', () => {
+        context.testComponent.ariaLabel = 'customLabel';
+        detailService.open({});
+        context.detectChanges();
+        const detailDialog = context.clarityElement.querySelector('[role="dialog"]');
+        expect(detailDialog.hasAttribute('aria-labelledby')).toEqual(false);
+        expect(detailDialog.hasAttribute('aria-label')).toEqual(true);
+        expect(detailDialog.getAttribute('aria-label')).toEqual('customLabel');
+      });
+
+      it('should add only aria-labelledby attribute if both clrDetailAriaLabelledBy and clrDetailAriaLabel inputs are available', () => {
+        context.testComponent.ariaLabelledBy = 'customId';
+        context.testComponent.ariaLabel = 'customLabel';
+        detailService.open({});
+        context.detectChanges();
+        const detailDialog = context.clarityElement.querySelector('[role="dialog"]');
+        expect(detailDialog.hasAttribute('aria-labelledby')).toEqual(true);
+        expect(detailDialog.hasAttribute('aria-label')).toEqual(false);
+      });
     });
   });
 }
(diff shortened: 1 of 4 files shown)


projects/angular/src/data/datagrid/datagrid-detail.ts
```diff
@@ -38,8 +39,27 @@ import { DetailService } from './providers/detail.service';
...
   @ContentChild(ClrDatagridDetailHeader) header: ClrDatagridDetailHeader;
 
+  @Input('clrDetailAriaLabelledBy') ariaLabelledBy: string;
+  @Input('clrDetailAriaLabel') ariaLabel: string;
+
   constructor(public detailService: DetailService, public commonStrings: ClrCommonStringsService) {}
 
+  get labelledBy(): string {
+    if (this.ariaLabelledBy) {
+      return this.header ? this.header.titleId + ' ' + this.ariaLabelledBy : this.ariaLabelledBy;
```
kevinbuhmann, line 49:
```suggestion
      return this.header ? `${this.header.titleId} ${this.ariaLabelledBy}` : this.ariaLabelledBy;
```

projects/angular/src/data/datagrid/datagrid-detail.ts
```diff
@@ -38,8 +39,27 @@ import { DetailService } from './providers/detail.service';
...
+    } else {
+      return this.ariaLabel ? null : this.header ? this.header.titleId : '';
```
kevinbuhmann, line 51:
```suggestion
      return this.ariaLabel || this.header?.titleId || '';
```

This is much easier to read than a nested ternary expression.

kevinbuhmann: Also, `this.ariaLabel` is not valid for `aria-label` because it is not a list of element IDs.

projects/angular/src/data/datagrid/datagrid-detail.ts
```diff
@@ -27,7 +27,8 @@ import { DetailService } from './providers/detail.service';
       role="dialog"
       [id]="detailService.id"
       aria-modal="true"
-      [attr.aria-describedby]="header ? header.titleId : ''"
+      [attr.aria-labelledby]="labelledBy"
+      [attr.aria-label]="label"
```
kevinbuhmann, line 31: Only one of these should be set.

"Don't use both on the same element because aria-labelledby will take precedence over aria-label if both are applied."
 https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label#:~:text=Don%27t%20use%20both%20on,label%20if%20both%20are%20applied.

kevinbuhmann (changes requested):

projects/angular/src/data/datagrid/datagrid-detail.ts
andyfeds: By default `aria-labelledby` is added by the component, We need to give the user the flexibility to add an `aria-label` incase they do not want to use `aria-labelledby`. So if the user adds an `aria-label` using the input [clrDetailAriaLabel]="myLabel", then we do not set `aria-labelledby` and set only `aria-label`.

projects/angular/src/data/datagrid/datagrid-detail.ts
kevinbuhmann: Oh, I misread the code.

projects/angular/src/data/datagrid/datagrid-detail.ts
kevinbuhmann: Okay, I didn't realize one of the attributes isn't applied even though it is bound.

projects/angular/src/data/datagrid/datagrid-detail.ts
```diff
@@ -38,8 +39,32 @@ import { DetailService } from './providers/detail.service';
...
+    if (this.ariaLabelledBy) {
+      return this.header ? `${this.header.titleId} ${this.ariaLabelledBy}` : this.ariaLabelledBy;
+    } else {
+      // If aria-label is set by the end user, do not set aria-labelledby
+      if (this.ariaLabel) {
+        return null;
+      } else {
+        return this.header?.titleId || '';
+      }
+    }
```
kevinbuhmann, line 57:
```suggestion
    } else if (this.ariaLabel) {
      // If aria-label is set by the end user, do not set aria-labelledby.
      return null;
    } else {
      return this.header?.titleId || '';
    }
```

This can be flattened.

projects/angular/src/data/datagrid/datagrid-detail.ts
```diff
@@ -38,8 +39,32 @@ import { DetailService } from './providers/detail.service';
...
+  }
+
+  get label(): string {
+    if (!this.ariaLabelledBy) {
+      return this.ariaLabel ? this.ariaLabel : null;
```
kevinbuhmann, line 62:
```suggestion
      return this.ariaLabel || null;
```

kevinbuhmann (changes requested):

projects/angular/src/data/datagrid/datagrid-detail.ts
andyfeds: Changed as suggested.

kevinbuhmann (changes requested):
Please add unit tests including tests to ensure both attributes are never both present and that `aria-labelledby` takes precedence over `aria-label`.

andyfeds:
> Please add unit tests including tests to ensure both attributes are never both present and that `aria-labelledby` takes precedence over `aria-label`.

Added unit tests

projects/angular/src/data/datagrid/all.spec.ts
```diff
@@ -81,6 +82,7 @@ describe('Datagrid', function () {
     DatagridCellSpecs();
     DatagridFilterSpecs();
     DatagridColumnSpecs();
+    DatagridDetailSpecs();
```
kevinbuhmann, line 85: I think the tests should be included and fixed separately to avoid including unrelated test fixes in the same commit as this bug fix. That can be confusing to readers. I was confused at first. I have opened #1451 for this.

projects/angular/src/data/datagrid/datagrid-detail.spec.ts
```diff
@@ -53,25 +54,25 @@ export default function (): void {
         expect(context.clarityElement.innerHTML).not.toContain(content);
       });
 
-      it('hides content with the esc key', async(() => {
+      it('hides content with the esc key', waitForAsync(() => {
```
kevinbuhmann, line 57:
```suggestion
      it('hides content with the esc key', async(() => {
```

This is another unrelated change. Please revert. It will be fixed when #1440 is merged.

projects/angular/src/data/datagrid/all.spec.ts
andyfeds: Oh yes, I meant to discuss this with you. The older tests were commented out and were failing.
Makes sense to move it to a new PR. I have approved https://github.com/vmware-clarity/ng-clarity/pull/1451
Please check.

projects/angular/src/data/datagrid/datagrid-detail.spec.ts
andyfeds: Done

projects/angular/src/data/datagrid/all.spec.ts
kevinbuhmann: Thanks, my PR is merged now.

projects/angular/src/data/datagrid/datagrid-detail.spec.ts
```diff
@@ -89,3 +128,16 @@ export default function (): void {
...
+    <clr-dg-detail [clrDetailAriaLabelledBy]="ariaLabelledBy" [clrDetailAriaLabel]="ariaLabel">
+      <clr-dg-detail-header>Title</clr-dg-detail-header>
+      <clr-dg-detail-body>${content}</clr-dg-detail-body>
+    </clr-dg-detail>
+  `,
+})
+class DataGridPaneTestComponent {
+  ariaLabelledBy: string;
+  ariaLabel: string;
+}
```
kevinbuhmann, line 143: I think you can just merge this into `FullTest`.

kevinbuhmann (approved):

PR #1410, opened 2024-05-29
chore(date-picker): style label correctly in stories

kevinbuhmann:
The `ClrFormsModule` needs to be imported so the `ClrLabel` directive is in scope.


kevinbuhmann:
> Adding `ClrFormsModule` have to be done in `.storybook/stories/datepicker/date-container.stories.ts` as well.

Good catch. Thanks! I also needed to adjust the CSS to prevent centering of the label.

valentin-mladenov (approved):
LGTM

PR #1415, opened 2024-05-30
fix(file-input): add vertical spacing between label and browse button

kevinbuhmann:
Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [x] If applicable, have a visual design approval

## PR Type

Bugfix

## What is the current behavior?

No vertical spacing between the label and an the browse button.

Issue Number: CDE-1668

## What is the new behavior?

Vertical spacing between the label and an the browse button.

## Does this PR introduce a breaking change?

No.

## Other Information

This commit also removes dead `.clr-control-label` styles. The label is not nested in the wrapper. That is why I didn't add the margin on the label.

PR #1416, opened 2024-05-30
fix(file-input): add vertical spacing between label and browse button

kevinbuhmann:
This is a backport of e378c6092db054fe8fddcc029e674f0e19988df2 (#1415) to 16.x.

This commit also removes dead `.clr-control-label` styles. The label is not nested in the wrapper. That is why I didn't add the margin on the label.

CDE-1668


dtsanevmw (approved):

valentin-mladenov (changes requested):
LGTM

valentin-mladenov (approved):

PR #1427, opened 2024-06-07
fix(datagrid): retain focus on closing detail

dtsanevmw:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When closing detail in datagrid trough the 'X' button it loses the focus for a second which triggers a scroll inside and fails to focus the button triger sometimes.

Issue Number: CDE-2043

## What is the new behavior?

Overlay detail is no longer setting the datagrid to `display:none` which block us from return the focus to the trigger at the right moment and we moved the focus trigger just before the close so we skip the moment where the focus is lost and it focuses the body.

- [ ] Yes
- [X] No


Jinnie (approved):
LGTM!

PR #1428, opened 2024-06-07
fix(datagrid): previously hidden columns visible after detail close

dtsanevmw:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Changing the page with detail open closes the detail and the columns which were hidden are still hidden.

Issue Number: CDE-2067

## What is the new behavior?

Columns will be visible again once the detail is closed.

- [ ] Yes
- [X] No

### DIFF
--- a/projects/angular/src/data/datagrid/render/main-renderer.ts
+++ b/projects/angular/src/data/datagrid/render/main-renderer.ts
@@ -19,7 +19,7 @@ import {
   QueryList,
   Renderer2,
 } from '@angular/core';
-import { merge, Subscription } from 'rxjs';
+import { Subscription } from 'rxjs';
 
 import { DomAdapter } from '../../../utils/dom-adapter/dom-adapter';
 import { DatagridColumnChanges } from '../enums/column-changes.enum';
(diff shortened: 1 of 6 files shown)


projects/angular/src/data/datagrid/render/main-renderer.ts
```diff
@@ -220,14 +226,18 @@ export class DatagridMainRenderer implements AfterContentInit, AfterViewInit, Af
   }
 
   private columnStateChanged(state) {
+    // eslint-disable-next-line eqeqeq
+    if (!this.headers || state.columnIndex == null) {
+      return;
+    }
```
valentin-mladenov, line 232:
```suggestion
    if (!this.headers || state.columnIndex === null || state.columnIndex === undefined) {
      return;
    }
```

valentin-mladenov (approved):
One comment check if it make sense.

projects/angular/src/data/datagrid/render/main-renderer.ts
dtsanevmw: using with two '==' checks for both null and undefined so it's okey.

valentin-mladenov: Yeah I know, my Idea was to remove the `// eslint-disable-next-line eqeqeq` exception.
It is better not to have any eslint disables exceptions unless is unavoidable and this is avoidable.

Also using `==` might affect some user application. 
Better be safe than sorry.

PR #1446, opened 2024-06-14
fix(datagrid): use proxy for detail context change

dtsanevmw:
## PR Checklist

## PR Type

Angular deprecated replacing the full context for `EmbeddedViewRef` which is used for datagrid detail and they suggest using a Proxy which is also used in `NgTemplateOutlet`. Introduced with https://github.com/angular/angular/pull/51887 and reported internally that it create problems in v17.
`Swapping out the context object for EmbeddedViewRef is no longer supported. Support for this was introduced with v12.0.0, but this pattern is rarely used. There is no replacement, but you can use simple assignments in most cases, or Object.assign , or alternatively still replace the full object by using a Proxy (see NgTemplateOutlet as an example).`

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [X] No


Jinnie (approved):
Thanks!

PR #1460, opened 2024-06-24
feat(datepicker): add support for date range

venkateshr06:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-1846

## What is the new behavior?
Adding support for Date Range picker. Ability to select start & end date using single form field.

- [ ] Yes
- [x] No


projects/angular/src/forms/datepicker/date-input-base.ts
```diff
@@ -264,33 +175,52 @@ export class ClrDateInput extends WrappedFormControl<ClrDateContainer> implement
...
-      this.previousDateChange = null;
+  protected datepickerHasFormControl() {
+    return !!this.control;
+  }
+
+  /**
+   * Incase of date range error, both start & end date field valdiation has to be triggered
+   * if either of the field gets updated
+   */
+  protected validateDateRange() {
```
Jinnie, line 186: nit: I still prefer this to be in the child/ren class/es where this code will apply if possible.

Jinnie (changes requested):
Thanks for separating this from the date format code.
Some, mostly minor, comments follow.
I also want to keep track on 2 design questions:
1) If selection spans over adjacent months, I think we should mark the grayed-out days as selected, too:
image

2) We should find a way to have the separating dash centered:
(might try right-aligning the second date)

image

dtsanevmw (commented):
There are visual changes that are made from design team that are not specifically made for the date range picker and affecting the default date picker. These changes are also fixing some a11y issues and are required even without the date range so should be in a separate PR.

projects/angular/src/forms/datepicker/date-single-input.ts
```diff
@@ -0,0 +1,123 @@
...
+import { datesAreEqual } from './utils/date-utils';
+
+@Directive({
+  selector: '[clrDate]',
+  host: {
+    '[class.clr-input]': 'true',
+  },
+  providers: [DatepickerFocusService],
+})
+export class ClrDateInput extends ClrDateInputBase implements AfterViewInit {
```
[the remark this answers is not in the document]
venkateshr06: Most of the common code are extracted to date-input-base class, only where we need to keep track of 'previousDateChange' value is retained in their respective Input classes. Code related to 'updateDate()' method is kept in their respective classes due to 'previousDateChange' value & control value dependency

projects/angular/src/forms/datepicker/date-input-base.ts
venkateshr06: To avoid code redundancy its kept in base class, also this method will be triggered only for Date range not for single date selection (already handled conditionally).

Jinnie (changes requested):
A default values story seems to work okay, we may leave it. Something like this:

```
export const DefaultDate: StoryObj = {
  render: DateRangePickerTemplate,
  args: {
    clrRangeStartDate: Date.now(),
    clrRangeEndDate: Date.now() + 2592000000,
  },
};
```

venkateshr06:
> A default values story seems to work okay, we may leave it. Something like this:
> 
> ```
> export const DefaultDate: StoryObj = {
>   render: DateRangePickerTemplate,
>   args: {
>     clrRangeStartDate: Date.now(),
>     clrRangeEndDate: Date.now() + 2592000000,
>   },
> };
> ```

Yes have already fixed the existing known issue in date picker. So it should work for DateRangePicker as well.

Jinnie (approved):
Looks okay to me as code now.
I would still expect it gets written design feedback before it gets merged.

venkateshr06:
> Looks okay to me as code now. I would still expect it gets written design feedback before it gets merged.

Resolved. Received email confirmation from Design Team.

PR #1470, opened 2024-07-04
fix(select): disable state color for single option

valentin-mladenov:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

1. Select disable color is using `cds-alias-typography-disabled`
2. A single disabled select option doesn't have proper colors set

Issue Number: CDE-2138 and CDE-2145

## What is the new behavior?
1. Select disable color is using `cds-alias-object-interaction-color-disabled`
2. A single disabled select option have proper colors set

- [ ] Yes
- [x] No


dtsanevmw (changes requested):
This mixin is using `cds-alias-typography-disabled` as underlying token but for this case as it's interaction item like in the dropdown it should use `cds-alias-object-interaction-color-disabled`

valentin-mladenov:
> This mixin is using `cds-alias-typography-disabled` as it's meaning is to be for the labels but for this case as it's interaction item like in the dropdown it should use `cds-alias-object-interaction-color-disabled`

@dtsanevmw Changing the rules to use token instead of the mixin is more broader task. The change will affect all select/option disable variations.

This PR only targets single option disable state to be aligned with the rest variants.

.storybook/stories/select/select.stories.ts
```diff
@@ -37,7 +37,9 @@ const selectTemplate: StoryFn = args => ({
     <clr-select-container>
       <label>Options</label>
       <select clrSelect>
-        <option *ngFor="let _ of createArray(optionCount); let i = index" [value]="i + 1">Option {{ i + 1 }}</option>
+        <option *ngFor="let _ of createArray(optionCount); let i = index" [value]="i + 1" [disabled]="i === 1">
```
kevinbuhmann, line 40:
```suggestion
        <option *ngFor="let _ of createArray(optionCount); let i = index" [value]="i + 1" [disabled]="i === 0">
```

Should this one be consistent with the multiple select story?

valentin-mladenov: Doesn't really matter.

I changed the single selection story. 
Commit e8fb100

dtsanevmw (approved):

PR #1490, opened 2024-07-23
chore(datepicker): use static dates in date range picker screenshots

kevinbuhmann:
This prevents the visual regression tests from failing due to the current date being different.

I also changed the default value in the single date picker story to be more readable.


valentin-mladenov (approved):
LGTM

PR #1511, opened 2024-08-08
chore: refactor active element assertions

kevinbuhmann:
- Use `toBe` instead of `toEqual`.
- Write assertions as `expect(actualValue).toBe(expectedValue)`.

### DIFF
--- a/projects/angular/src/wizard/wizard.spec.ts
+++ b/projects/angular/src/wizard/wizard.spec.ts
@@ -798,7 +798,7 @@ export default function (): void {
           wizard.pageCollection.lastPage.makeCurrent();
           context.detectChanges();
           const titleString = context.hostElement.querySelector('.modal-title').textContent.trim();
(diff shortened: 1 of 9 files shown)


projects/angular/src/wizard/wizard.spec.ts
```diff
@@ -798,7 +798,7 @@ export default function (): void {
           wizard.pageCollection.lastPage.makeCurrent();
           context.detectChanges();
           const titleString = context.hostElement.querySelector('.modal-title').textContent.trim();
-          expect(titleString).toEqual(document.activeElement.textContent.trim());
+          expect(document.activeElement.textContent.trim()).toEqual(titleString);
```
valentin-mladenov, line 801:
```suggestion
          expect(document.activeElement.textContent.trim()).toBe(titleString);
```

valentin-mladenov (approved):

PR #1518, opened 2024-08-13
chore: remove commitlint

kevinbuhmann:
Since we squash commits when merging PRs, there isn't a need to check the format of individual commit messages.


Jinnie (approved):
Nice, thanks!

PR #1521, opened 2024-08-22
chore(ci): remove Slack post from release process

kevinbuhmann:
The Slack channel is readonly now prior to the Slack instance being shut down.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
           NPM_TOKEN: ${{secrets.NPM_TOKEN}}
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -27,7 +27,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
           NPM_TOKEN: ${{secrets.NPM_TOKEN}}
-          SLACK_WEBHOOK: ${{secrets.SLACK_WEBHOOK}}
```
valentin-mladenov, line 30: @kevinbuhmann did you remove the `SLACK_WEBHOOK` from the secrets as well?

valentin-mladenov (approved):
LGTM. Check my comment.

.github/workflows/release.yml
kevinbuhmann: I will.

PR #1527, opened 2024-08-29
chore(forms): remove control height unit tests

kevinbuhmann:
These tests are flaky and we have visual regression tests.


valentin-mladenov (approved):
LGTM.
I would prefer to comment them out instead of deleting them. but it fine.

kevinbuhmann:
> I would prefer to comment them out instead of deleting them.

We can always find this in the repo history if needed.

PR #1540, opened 2024-09-04
fix(combobox): stop event propagation on option click

dtsanevmw:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Propagating click events on option click.

Issue Number: CDE-2210

## What is the new behavior?

Stops the propagation of click events.

- [ ] Yes
- [ ] No


Jinnie (approved):
I am okay with this.
In general, we've been trying to avoid imposing behaviors that the users cannot override.
But if this is a real case for them, it would mean they have nested actionable elements, which is at least not accessible.
So, let's start with the combobox and gradually we can add this to the other elements that use popovers. Because we'll have the same issue with dropdowns, signposts, etc.
Also, we may need to start doing this for Esc buttons.

projects/angular/src/forms/combobox/option.ts
```diff
@@ -78,8 +78,9 @@ export class ClrOption<T> implements OnInit {
     }
   }
 
-  @HostListener('click')
-  onClick() {
+  @HostListener('click', ['$event'])
+  onClick(event) {
```
kevinbuhmann, line 82:
```suggestion
  onClick(event: MouseEvent) {
```

kevinbuhmann (changes requested):

projects/angular/src/forms/combobox/option.ts
dtsanevmw: Added in [ed14c8a](https://github.com/vmware-clarity/ng-clarity/pull/1540/commits/ed14c8a6dc53d57bc5d0ac4fa65096fa323d915b)

PR #1566, opened 2024-09-26
chore: clean up parameter properties

kevinbuhmann:
- remove unused parameter properties (manual)
- use parameter properties instead of a separate property (lint rule)
- remove unnecessary `this` when accessing parameter properties (lint rule for detecting unneded `this.`, manual removal of no-longer-needed properties)


valentin-mladenov (approved):
LGTM

Issue #1569, opened 2024-09-27
Datagrid rendering issue


vzdendyak:
**This is a bug report for the `@clr` Angular or UI versions of the design system.
For the web-component implementation of Clarity (`@cds`), visit [vmware-clarity/core](https://github.com/vmware-clarity/core).**

**_If you are a VMware employee or a contractor in VMware, please use our support space in Google Chat to raise Clarity issues._**

## Describe the bug

I'm encountering some performance issues with the datagrid component. Specifically, I've noticed the following problems:

* Column Width Jumps: During the rendering process, the column widths appear to "jump" or adjust noticeably.

https://github.com/user-attachments/assets/33e989ee-3941-4051-9df3-e0f1c9a51437

<img width="807" alt="image" src="https://github.com/user-attachments/assets/4b0c1518-e251-4832-b3a4-7776708506f2">

 
## How to reproduce

Here is the [stackblitz link](https://stackblitz.com/edit/clarity-light-theme-clr17-umsycv?file=src%2Fapp%2Fapp.component.ts).

I generate data dynamically and use the setTimeout to imitate the data loading process. The issue happens when the data is loaded and initially displayed in the Datagrid. It's visible that the column width is incorrect and it takes some time to recalculate the column width and display the grid correctly.

1.  Open the example
2. Look at the datagrid column width change

Sometimes it renders fast, so please try to refresh the page several times.

## Expected behavior

The datagrid data initial render should be smooth as it was in the previous versions.

## Versions

**Clarity version:**

- [x] v17.3.0

**Framework version:**
- [x] Angular 17
- [x] Angular 18

**Device:**

- Type: [Laptop]
- OS: [MacOS, Windows]
- Browser [Chrome, Opera, Firefox]  (IMPORTANT: I can't reproduce this on Safari)

## Additional notes

This issues affects our application seriously and stops us from migration to Angular 18 and Clarity 17.3.0

PR #1494, opened 2024-07-25
fix(datagrid): nested in accordion columns have wrong width

valentin-mladenov:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?
Nested in accordion datagrid columns have wrong width.
image

Issue Number: CDE-2208

## What is the new behavior?
Nested in accordion datagrid columns have proper width.
image

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #1577, opened 2024-10-02
fix(datagrid): remove setTimeout and check visibility for calculations

dtsanevmw:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

There was added `setTimeout` https://github.com/vmware-clarity/ng-clarity/pull/1494 in the renderer to fix an issue with datagrid being in accordion but that creates different issue - #1569. The main root cause of the wrong widths inside the accordion come from that the content of accordion is using separate component with `ng-content` and that is projected inside `accordion-panel` wrapped in *ngIf which seems okey but Angular life cycles go trough that content component before the condition on the ngIf is met so the calculations are being done wrongly on invisible datagrid. In header-renderer scrollWidth is 0 due to that the component is not actually visible after it's visible "shouldStabilizeColumns" is already false and no more re-calculations are being done.

Issue Number: #1569

## What is the new behavior?

Do the calculations when the datagrid is actually visible in the DOM.

- [ ] Yes
- [X] No


projects/angular/src/data/datagrid/datagrid.ts
```diff
@@ -127,6 +127,7 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
 
   selectAllId: string;
   hasVirtualScroller: boolean;
+  isVisible = false;
```
kevinbuhmann, line 130: I don't think this should publicly writable.

projects/angular/src/data/datagrid/datagrid.ts
```diff
@@ -347,6 +349,10 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
             this._calculationRows.insert(row._view);
           });
         }
+      }),
+      this.visiblityChange.subscribe(visibility => {
```
kevinbuhmann, line 353: Do we need a subject and a subscription for this? Why not just a `setVisibility` method. I think this subscribe callback happens synchronously upon `next`, so this feels like unnecessary indirection.

kevinbuhmann (changes requested):

kevinbuhmann (approved):
LGTM, I didn't test anything though.

projects/angular/src/data/datagrid/render/main-renderer.ts
```diff
@@ -59,6 +60,7 @@ export class DatagridMainRenderer implements AfterContentInit, AfterViewInit, Af
   private _heightSet = false;
   private shouldStabilizeColumns = true;
   private subscriptions: Subscription[] = [];
+  private isVisible = false;
```
kevinbuhmann, line 63: Wait, you don't need this property. You just return the value.

kevinbuhmann (changes requested):
See comment.

projects/angular/src/data/datagrid/render/main-renderer.ts
kevinbuhmann: Maybe rename `setVisibility` to `checkAndUpdateVisibility` and return the value.

projects/angular/src/data/datagrid/render/main-renderer.ts
```diff
@@ -287,4 +289,14 @@ export class DatagridMainRenderer implements AfterContentInit, AfterViewInit, Af
       this.columnsSizesStable = true;
     }
   }
+
+  private setVisibility() {
+    if (this.el.nativeElement.offsetParent) {
+      this.isVisible = true;
+      this.datagrid.datagrid.nativeElement.style.visibility = 'visible';
```
valentin-mladenov, line 291: Setting it to empty string is enough. Test it just in case.
```suggestion
      this.datagrid.datagrid.nativeElement.style.visibility = '';
```

projects/angular/src/data/datagrid/render/main-renderer.ts
```diff
@@ -287,4 +289,14 @@ export class DatagridMainRenderer implements AfterContentInit, AfterViewInit, Af
...
+    } else {
+      this.isVisible = false;
+      this.datagrid.datagrid.nativeElement.style.visibility = 'hidden';
+    }
+  }
```
kevinbuhmann, line 301:
```suggestion
  private checkAndUpdateVisibility() {
    if (this.el.nativeElement.offsetParent) {
      this.datagrid.datagrid.nativeElement.style.visibility = 'visible';
      return true;
    } else {
      this.datagrid.datagrid.nativeElement.style.visibility = 'hidden';
      return false;
    }
  }
```

projects/angular/src/data/datagrid/render/main-renderer.ts
valentin-mladenov, line 291: Rename to `clrDatagridComponent`. Accessing it through `this.datagrid.datagrid` feels odd.

```suggestion
      this.clrDatagridComponent.datagrid.nativeElement.style.visibility = 'visible';
```

kevinbuhmann (approved):

valentin-mladenov (commented):
Check my comments

projects/angular/src/data/datagrid/render/main-renderer.ts
kevinbuhmann: I don't think the `clr` prefix is needed.

kevinbuhmann: I think what Daniel has is better. Setting it to `'visible'` has a obvious effect. Setting it to an empty string might do the same thing, but it not as readable.

Issue #1501, opened 2024-07-31
Datagrid: hiding the last column adds whitespaces at the end and shows scrollbar


fichtip:
**This is a bug report for the `@clr` Angular or UI versions of the design system.
For the web-component implementation of Clarity (`@cds`), visit [vmware-clarity/core](https://github.com/vmware-clarity/core).**

**_If you are a VMware employee or a contractor in VMware, please use our support space in Google Chat to raise Clarity issues._**

## Describe the bug

Hiding the last column of a datagrid ends up in showing some whitespaces a the end of the datagrid in addition to the scrollbar.

Showing all columns
![grafik](https://github.com/user-attachments/assets/bb8a0b31-579c-4ea3-a246-5f7b62cf5345)

Hiding non-last column
![grafik](https://github.com/user-attachments/assets/f4eee8fa-d806-4c05-a0d2-da8b8bee6392)

Hiding last column
![grafik](https://github.com/user-attachments/assets/fea917b1-1730-4734-8286-709ce3ef34cd)

## How to reproduce

This behavior can also be seen with the provided example in the [documentation](https://clarity.design/documentation/datagrid/code/hide-show).

1.  Open link of documentation from above
2.  Click on 'Manage Column'
3.  unselect 'Favorite color'

## Expected behavior

No whitespaces and no scrollbar when hiding the last element.

## Versions

**Clarity version:**

```
"@cds/core": "^6.12.0",
"@clr/angular": "^17.2.1",
"@clr/ui": "^17.2.1",
```

**Framework version:**

`"@angular/core": "~17.1.2",`

**Device:**

- Type: Lenovo ThinkPad P53
- OS: Windows 11
- Browser Firefox, Edge
- Version FF 128.0.3, Edge 127.0.2651.74 

## Additional notes

_Add any other notes about the problem here._

dtsanevmw:
Fixed in [b4a4935](https://github.com/vmware-clarity/ng-clarity/commit/b4a4935b9103fbb04a1f7944e11872f2eb7f5709) and it will be in the next release.

PR #1584, opened 2024-10-08
fix(datagrid): hide separator from last column dynamically

dtsanevmw:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently we have a CSS selector for `:last-child` to hide the separator from the column but when the last column gets hidden that's happening only visually and it's still in the DOM so that's not working.

Issue Number: CDE-2317, #1501 

## What is the new behavior?

Update showSeparator states when hidden state changes.

- [ ] Yes
- [X] No


projects/angular/src/data/datagrid/datagrid-column.ts
```diff
@@ -157,6 +160,18 @@ export class ClrDatagridColumn<T = any>
...
+    return this.el.nativeElement.classList.contains(HIDDEN_COLUMN_CLASS);
+  }
+
+  get showSeparator() {
+    return this._showSeparator;
+  }
+  set showSeparator(value: boolean) {
+    this._showSeparator = value;
+    this.changeDetectorRef.markForCheck();
+  }
```
kevinbuhmann, line 173: Would a `BehaviorSubject` and `async` pipe work here instead of manually calling `markForCheck`? I think that would be cleaner and less code.

kevinbuhmann (approved):

PR #1590, opened 2024-10-12
fix(nav): register close button icon and fix invalid html

kevinbuhmann:


Jinnie (approved):

projects/angular/src/layout/nav/nav-level.ts
```diff
@@ -27,15 +28,17 @@ import { ResponsiveNavigationService } from './providers/responsive-navigation.s
...
+  ClarityIcons.addIcons(windowCloseIcon);
+
   const closeButton = document.createElement('button');
   closeButton.setAttribute('aria-label', ariaLabel);
   closeButton.setAttribute('aria-hidden', 'true');
   closeButton.innerHTML = `
     <cds-icon
       inner-offset="1"
-      shape="close"
+      shape="window-close"
```
valentin-mladenov, line 39: Aren't we using `times` icon for close buttons? 

```suggestion
      shape="times"
```

kevinbuhmann: We use `times` for the datagrid detail pane and `windowClose` for the modal. I think it should be `windowClose` in all three places, but I will change it back to `times` for this PR.

valentin-mladenov: Thank you. I think `times` is the proper icon for all places as well.

PR #1616, opened 2024-11-12
fix(accordion): add heading role and aria level to title

mivaylo:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Accordion and Stepper headers don't have `role` and `aria-level` assigned.

Issue Number: CDE-2437

## What is the new behavior?
Accordion and Stepper headers have `role=heading` and `aria-level` can be assigned from the user with defaul value of 3.

- [ ] Yes
- [x] No


dtsanevmw (commented):
Check the implementation where we are checking if the component is nested so we put 4 instead of 3 for the default value

kevinbuhmann:
This is a breaking change because it can cause pages with legal heading structure to suddenly have illegal heading structure due to the new headings.

projects/angular/src/accordion/accordion-panel.ts
```diff
@@ -42,6 +44,11 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
   @Input('clrAccordionPanelDisabled') @HostBinding('class.clr-accordion-panel-disabled') disabled = false;
   @Input('clrAccordionPanelOpen') panelOpen = false;
 
+  /**
+   * Level of the accordion/stepper heading from 1 to 7.
+   */
+  @Input('clrAriaLevel') _ariaLevel: number;
```
dtsanevmw, line 50:
```suggestion
  @Input('clrAccordionAriaLevel') _ariaLevel: number;
```

dtsanevmw: Yeah, missed the 'Panel' part. We corrected it offline

valentin-mladenov: fixed in commit: aa02e3139de0d67e433c4d3fa3ea3cbb7bddb584

projects/angular/src/accordion/accordion-panel.ts
```diff
@@ -42,6 +44,12 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
   @Input('clrAccordionPanelDisabled') @HostBinding('class.clr-accordion-panel-disabled') disabled = false;
   @Input('clrAccordionPanelOpen') panelOpen = false;
 
+  /**
+   * Level of the accordion/stepper heading from 1 to 7.
+   */
+  @Input('clrAccordionPanelAriaLevel') _ariaLevel: number;
+  @Input('clrAccordionPanelHeading') heading = false;
```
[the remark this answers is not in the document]
valentin-mladenov: Corrected in commit 8c5995b

projects/angular/src/accordion/accordion-panel.spec.ts
```diff
@@ -36,6 +38,8 @@ class TestComponent {
   open = false;
   disabled = false;
   showDescription = false;
+  showHeading = false;
+  ariaLevel = 3;
```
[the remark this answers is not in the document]
valentin-mladenov: Corrected in commit 8c5995b

projects/angular/src/accordion/accordion-panel.spec.ts
```diff
@@ -22,6 +22,8 @@ import { AccordionService } from './providers/accordion.service';
     <clr-accordion>
       <clr-accordion-panel
         [(clrAccordionPanelOpen)]="open"
+        [clrAccordionPanelAriaLevel]="ariaLevel"
+        [clrAccordionPanelHeading]="showHeading"
```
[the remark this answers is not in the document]
valentin-mladenov: Corrected in commit 8c5995b

projects/angular/src/accordion/accordion-panel.ts
```diff
@@ -42,6 +45,12 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
   @Input('clrAccordionPanelDisabled') @HostBinding('class.clr-accordion-panel-disabled') disabled = false;
   @Input('clrAccordionPanelOpen') panelOpen = false;
 
+  @Input('clrAccordionPanelHeading') heading = false;
+  /**
+   * Level of the accordion/stepper heading from 1 to 7.
+   */
+  @Input('clrAccordionPanelHeadingLevel') _explicitHeadingLevel: HeadingLevel;
```
kevinbuhmann, line 52:
```suggestion
  @Input('clrAccordionPanelHeadingLevel') explicitHeadingLevel: HeadingLevel;
```

I don't think a underscore is needed here. That's usually only used for fields that back a property of the same same. That's not really the case here. I think "explicit" replaces the needed for the underscore while also making the code more understandable.

valentin-mladenov: Corrected in commit: 37c1f1b

projects/angular/src/accordion/accordion-panel.ts
```diff
@@ -42,6 +45,12 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
   @Input('clrAccordionPanelDisabled') @HostBinding('class.clr-accordion-panel-disabled') disabled = false;
   @Input('clrAccordionPanelOpen') panelOpen = false;
 
+  @Input('clrAccordionPanelHeading') heading = false;
```
kevinbuhmann, line 48:
```suggestion
  @Input('clrAccordionPanelHeading') headingEnabled = false;
```

kevinbuhmann: A property just named `heading` is confusing to read. This property does not hold a "heading". It holds a feature flag.

kevinbuhmann, line 48:
```suggestion
  @Input('clrAccordionPanelHeadingEnabled') headingEnabled = false;
```

A property just named `heading` is confusing to read. This property does not hold a "heading". It holds a feature flag.

kevinbuhmann (changes requested):

projects/angular/src/accordion/accordion-panel.ts
```diff
@@ -52,12 +61,23 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
   private _panelIndex: number;
 
   constructor(
+    @Optional()
+    @SkipSelf()
+    public parent: ClrAccordionPanel,
```
kevinbuhmann, line 66: Why is this `public`? I think it should either be `protected` or `private`.

projects/angular/src/accordion/stepper/stepper-panel.ts
```diff
@@ -45,6 +46,9 @@ export class ClrStepperPanel extends ClrAccordionPanel implements OnInit {
   private subscriptions: Subscription[] = [];
 
   constructor(
+    @Optional()
+    @SkipSelf()
+    public override parent: ClrAccordionPanel,
```
kevinbuhmann, line 51: Sam here. I think this should either be `protected` or `private`.

valentin-mladenov: It can be `protected`, but not `private`.

projects/angular/src/accordion/stepper/stepper-panel.spec.ts
```diff
@@ -124,6 +132,18 @@ describe('ClrStep Reactive Forms', () => {
 
       expect(input.focus).toHaveBeenCalled();
     }));
+
+    it('should have the aria-level attribute set to 4', () => {
```
kevinbuhmann, line 136:
```suggestion
    it('should allow the aria-level attribute to be set to an explicit value', () => {
```

This test is for an explicit heading value, so I think the test description should say that.

projects/angular/src/accordion/accordion-panel.spec.ts
```diff
@@ -268,5 +276,15 @@ describe('ClrAccordionPanel', () => {
       fixture.detectChanges();
       expect(panelElement.querySelector('.clr-accordion-header-has-description')).toBeTruthy();
     });
+
+    it('should have the aria-level attribute set to 5', () => {
```
kevinbuhmann, line 280:
```suggestion
    it('should allow the aria-level attribute to be set to an explicit value', () => {
```

Same here.

projects/angular/src/accordion/accordion-panel.spec.ts
```diff
@@ -226,6 +232,8 @@ describe('ClrAccordionPanel', () => {
       expect(headerButton.getAttribute('aria-controls')).toBe(panelRegion.getAttribute('id'));
       expect(headerButton.getAttribute('disabled')).toBe(null);
 
+      expect(panelHeading).toBeNull();
```
kevinbuhmann, line 235: This should be extracted to a test specifically for the feature flag not being enabled. Also, there should be tests for the feature flag being enabled without an explicit heading level value for a parent and nested accordion. Same for the stepper.

projects/angular/src/accordion/accordion-panel.ts
kevinbuhmann: I think this one can be `protected` as well.

projects/angular/src/accordion/stepper/stepper-panel.spec.ts
```diff
@@ -22,11 +23,18 @@ import { ClrStepperModule } from './stepper.module';
 @Component({
   template: `
     <form clrStepper [formGroup]="form">
-      <clr-stepper-panel formGroupName="groupName">test step</clr-stepper-panel>
+      <clr-stepper-panel
+        [clrAccordionPanelHeading]="showHeading"
+        [clrAccordionPanelHeadingLevel]="headingLevel"
```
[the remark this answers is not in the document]
kevinbuhmann: You might be able to `override` the properties and set the value on the `super` property to just change the `@Input()` alias for the stepper. That would be preferable so that the accordion input names don't even work on stepper panels.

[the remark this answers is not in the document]
valentin-mladenov: I'll remove the test.

projects/angular/src/accordion/accordion-panel.ts
valentin-mladenov: It will be.

projects/angular/src/accordion/accordion-panel.ts
```diff
@@ -42,6 +45,12 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
   @Input('clrAccordionPanelDisabled') @HostBinding('class.clr-accordion-panel-disabled') disabled = false;
   @Input('clrAccordionPanelOpen') panelOpen = false;
 
+  @Input('clrAccordionPanelHeading') heading = false;
+  /**
+   * Level of the accordion/stepper heading from 1 to 7.
```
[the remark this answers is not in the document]
valentin-mladenov: Corrected in commit: cc43f46

projects/angular/src/accordion/stepper/stepper-panel.ts
valentin-mladenov: Corrected in commit: cc43f46

projects/angular/src/accordion/accordion-panel.ts
valentin-mladenov: Corrected in commit: cc43f46

projects/angular/src/accordion/stepper/stepper-panel.spec.ts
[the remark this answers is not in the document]
valentin-mladenov: Corrected in commit: cc43f46

projects/angular/src/accordion/accordion-panel.spec.ts
kevinbuhmann: This is redundant now, so I think it should be removed.

projects/angular/src/accordion/accordion-panel.spec.ts
```diff
@@ -176,6 +181,13 @@ describe('ClrAccordionPanel', () => {
         accordionPanel.togglePanel();
         expect(component.change).toHaveBeenCalledWith(false);
       });
+
+      it('when [clrAccordionPanelHeadingEnabled] is not set at all, heading role should not be present', () => {
```
kevinbuhmann, line 185: Why is this in the "TypeScript API" test block instead of the "View" test block with other test? It's a test for an input and a DOM attribute (or lack thereof), so I think it should be moved.

kevinbuhmann, line 185: This test does not belong in the "Template API -> Output (clrAccordionPanelOpenChange)" test block. It should be in the "View" test block with other test.

kevinbuhmann (changes requested):
See my comment. Also, there should be tests for the default values (both parent and nested).

projects/angular/src/accordion/stepper/stepper-panel.ts
```diff
@@ -53,7 +57,7 @@ export class ClrStepperPanel extends ClrAccordionPanel implements OnInit {
     ifExpandService: IfExpandService,
     cdr: ChangeDetectorRef
   ) {
-    super(commonStrings, stepperService, ifExpandService, cdr);
+    super(parent, commonStrings, stepperService, ifExpandService, cdr);
```
kevinbuhmann, line 60:
```suggestion
    super(null, commonStrings, stepperService, ifExpandService, cdr);
```

We discussed that this feature is not supported in the stepper. The `ClrAccordionPanel` panel dependency should also be removed.

projects/angular/src/accordion/accordion-panel.ts
```diff
@@ -52,12 +61,23 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
   private _panelIndex: number;
 
   constructor(
+    @Optional()
+    @SkipSelf()
+    protected parent: ClrAccordionPanel,
```
kevinbuhmann, line 66:
```suggestion
    private parent: ClrAccordionPanel,
```

This can be `private`. 1) The override in the stepper is not needed since this is an accordion-only feature. 2) Even if we did support this feature in the stepper, you do not need to override the property. You just need to pass a parameter to the `super` constructor.

kevinbuhmann: Also, I would put this on one line like everything else.

```
@Optional() @SkipSelf() private parent: ClrAccordionPanel,
```

projects/angular/src/accordion/accordion-panel.ts
```diff
@@ -52,12 +61,23 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
...
     private cdr: ChangeDetectorRef
   ) {}
 
+  get headingLevel() {
+    if (this.explicitHeadingLevel) {
+      return this.explicitHeadingLevel;
+    }
+
+    return this.parent ? 4 : 3;
+  }
```
kevinbuhmann, line 79: nit: It feels odd to me to put this before the `id` property.

valentin-mladenov: Corrwcted in commit 2cb2400

projects/angular/src/accordion/accordion-panel.ts
valentin-mladenov: Corrwcted in commit 2cb2400

projects/angular/src/accordion/stepper/stepper-panel.ts
valentin-mladenov: Corrwcted in commit 2cb2400

projects/angular/src/accordion/accordion-panel.spec.ts
valentin-mladenov: Corrwcted in commit 2cb2400

kevinbuhmann (changes requested):
There is still no test for the default heading level for a parent and nested accordion.

projects/angular/src/accordion/accordion-panel.spec.ts
```diff
@@ -36,6 +39,40 @@ class TestComponent {
...
+          </clr-accordion>
+        </clr-accordion-content>
+      </clr-accordion-panel>
+    </clr-accordion>
+  `,
+})
+class TestNestedAccordionComponent {
   change = state => {
     return state;
   };
```
kevinbuhmann, line 78: This should be removed in the `TestNestedAccordionComponent` because it is not used.

projects/angular/src/accordion/accordion-panel.spec.ts
```diff
@@ -268,5 +305,45 @@ describe('ClrAccordionPanel', () => {
...
+      const panelHeading = panelElement.querySelector('[role="heading"]');
+      expect(panelHeading).not.toBeNull();
+      expect(panelHeading.getAttribute('aria-level')).toBe('5');
+    });
+
+    it('when [clrAccordionPanelHeadingEnabled] is not set at all, heading role should not be present', () => {
+      const fixture: ComponentFixture<TestNoBindingComponent> = TestBed.createComponent(TestNoBindingComponent);
+      fixture.detectChanges();
+
+      const accordionPanel = fixture.debugElement.query(By.directive(ClrAccordionPanel)).nativeElement as HTMLElement;
```
kevinbuhmann, line 325:
```suggestion
      const accordionPanel: HTMLElement = fixture.debugElement.query(By.directive(ClrAccordionPanel)).nativeElement;
```

Type annotations are generally better than type assertions when possible to avoid overriding certain compiler type checks.

projects/angular/src/accordion/accordion-panel.spec.ts
```diff
@@ -268,5 +305,45 @@ describe('ClrAccordionPanel', () => {
...
+
+      expect(panelHeading).toBeNull();
+    });
+
+    it('default heading aria-level should be present in nested accordion', () => {
+      const fixture: ComponentFixture<TestNestedAccordionComponent> =
+        TestBed.createComponent(TestNestedAccordionComponent);
+      fixture.detectChanges();
+
+      const accordionPanel = fixture.debugElement.query(By.directive(ClrAccordionPanel)).nativeElement as Element;
```
kevinbuhmann, line 336:
```suggestion
      const accordionPanel: HtmlElement = fixture.debugElement.query(By.directive(ClrAccordionPanel)).nativeElement;
```

projects/angular/src/accordion/accordion-panel.ts
```diff
@@ -42,6 +45,12 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
   @Input('clrAccordionPanelDisabled') @HostBinding('class.clr-accordion-panel-disabled') disabled = false;
   @Input('clrAccordionPanelOpen') panelOpen = false;
 
+  @Input('clrAccordionPanelHeadingEnabled') headingEnabled = false;
+  /**
+   * Level of the accordion/stepper heading from 1 to 6.
```
kevinbuhmann, line 50:
```suggestion
   * Level of the accordion/stepper heading from 1 to 6. This only works if `[clrAccordionPanelHeadingEnabled]` is set to `true`.
```

I just realized that this documentation comment is incomplete. We might also want to document `[clrAccordionPanelHeadingEnabled]` as well.

projects/angular/src/accordion/accordion-panel.html
```diff
@@ -1,6 +1,10 @@
 <ng-container *ngIf="panel | async; let panel">
   <div [ngClass]="getPanelStateClasses(panel)">
-    <div class="clr-accordion-header">
+    <div
+      class="clr-accordion-header"
+      [attr.role]="headingEnabled ? 'heading' : null"
+      [attr.aria-level]="headingEnabled ? headingLevel : null"
```
kevinbuhmann, line 6:
```suggestion
      [attr.role]="headingEnabled || explicitHeadingLevel ? 'heading' : null"
      [attr.aria-level]="headingEnabled || explicitHeadingLevel ? headingLevel : null"
```

I don't think the current implementation is quite right. If the user sets an explicit heading level, it should work without the feature flag being enabled (setting the heading level implicitly means the user wants the accordion header to be `role="heading`).

valentin-mladenov: It is right.

If there is not flag: `role="heading"` will be always applyed. 
If there is not flag: ` [attr.aria-level]="headingLevel"` will be applyed. 

`explicitHeadingLevel` is only for that `explicit`, `headingLevel` will alway have value.

projects/angular/src/accordion/accordion-panel.spec.ts
valentin-mladenov: Corrected in commit: d9b21e4

projects/angular/src/accordion/accordion-panel.spec.ts
valentin-mladenov: Corrected in commit: d9b21e4

projects/angular/src/accordion/accordion-panel.spec.ts
valentin-mladenov: Corrected in commit: d9b21e4

projects/angular/src/accordion/accordion-panel.html
valentin-mladenov: Logic updated in ac47ae2

projects/angular/src/accordion/accordion-panel.spec.ts
```diff
@@ -268,5 +296,44 @@ describe('ClrAccordionPanel', () => {
       fixture.detectChanges();
       expect(panelElement.querySelector('.clr-accordion-header-has-description')).toBeTruthy();
     });
+
+    it('should allow the aria-level attribute to be set to an explicit value', () => {
+      expect(panelElement.querySelector('[role="heading"]')).toBeNull();
+
+      fixture.componentInstance.headingLevel = 5;
+      fixture.componentInstance.showHeading = true;
```
kevinbuhmann, line 304: I'm going to approve this, but this line could be removed now.

kevinbuhmann (approved):

projects/angular/src/accordion/accordion-panel.html
kevinbuhmann: Thanks!

projects/angular/src/accordion/accordion-panel.spec.ts
```diff
@@ -36,11 +39,36 @@ class TestComponent {
   open = false;
   disabled = false;
   showDescription = false;
+  showHeading = false;
+  headingLevel: HeadingLevel = 3;
```
kevinbuhmann, line 43:
```suggestion
  headingLevel: HeadingLevel;
```

This change will fix the tests.

projects/demo/src/app/accordion/accordion.demo.html
```diff
@@ -68,6 +68,26 @@ <h2>Angular - nested datagrid</h2>
...
+<h2>Angular - nested accordion</h2>
+
+<clr-accordion>
+  <clr-accordion-panel [clrAccordionPanelOpen]="true">
+    <clr-accordion-title>Parent Title 1</clr-accordion-title>
+    <clr-accordion-content>
+      <clr-accordion>
+        <clr-accordion-panel
+          [clrAccordionPanelHeadingEnabled]="true"
+          [clrAccordionPanelHeadingLevel]="3"
```
kevinbuhmann, line 80:
```suggestion
          [clrAccordionPanelHeadingLevel]="3"
```

projects/angular/src/accordion/accordion-panel.html
```diff
@@ -1,6 +1,10 @@
 <ng-container *ngIf="panel | async; let panel">
   <div [ngClass]="getPanelStateClasses(panel)">
-    <div class="clr-accordion-header">
+    <div
+      class="clr-accordion-header"
+      [attr.role]="headingEnabled || explicitHeadingLevel ? 'heading' : null"
+      [attr.aria-level]="headingEnabled || explicitHeadingLevel ? headingLevel : null"
```
kevinbuhmann, line 6:
```suggestion
      [attr.role]="headingEnabled || (explicitHeadingLevel && headingEnabled !== false) ? 'heading' : null"
      [attr.aria-level]="headingEnabled || (explicitHeadingLevel && headingEnabled !== false) ? headingLevel : null"
```

I just realized I made a mistake before... I'm sorry this is difficult to get right. If the user sets an explicit heading level *and* disables the heading feature flag, it probably should not render the `role="heading"`.

kevinbuhmann (changes requested):
Sorry for the multiple reviews. This is difficult to get right. Feature flags are difficult. I found a bug. See my comment.

Also, the accordion panel story needs the new inputs to be set in the template.

```html
      [clrAccordionPanelHeadingEnabled]="clrAccordionPanelHeadingEnabled"
      [clrAccordionPanelHeadingLevel]="clrAccordionPanelHeadingLevel"
 ```

projects/angular/src/accordion/accordion-panel.html
kevinbuhmann: My thought process is that the explicit heading level implicitly enables the heading, but an explicitly disabling of the heading should take precendence.

valentin-mladenov: corrected

valentin-mladenov: Change reverted. It's not a bug, but a feature.

Providing `clrAccordionPanelHeadingEnabled=false` and `clrAccordionPanelHeadingLevel=3` is unsupported scenario.

kevinbuhmann (approved):

kevinbuhmann (changes requested):
Wait, the story still needs to be updated.

kevinbuhmann (approved):
Nevermind, the story is updated.

PR #1612, opened 2024-11-06
feat(datagrid): add ability to unsort columns

valentin-mladenov:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?
Once turned on column sort can only be rotated between `ASC` and `DESC ` when clicking on the column, but can't be turned off. 

Issue Number: CDE-2416

## What is the new behavior?
Column sort can be turned off by clicking third time on the column.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #1617, opened 2024-11-13
feat(datagrid): add ability to unsort datagrid colums

valentin-mladenov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Once turned on column sort can only be rotated between `ASC` and `DESC`  when clicking on the column, but can't be turned off.

Issue Number: CDE-2439, CDE-2416

## What is the new behavior?

- Column sort can be turned off by clicking third time on the column.
- The feature can be turned on/off per column by `clrDgDisableUnsort` flag on column level.

- [x] Yes
- [ ] No

## Other information

BREAKING CHANGE: 
1. Datagrid columns now have UNSORTED state which activates on the third click on sortable column. It can be disabled per column trough `clrDgDisableUnsort` input.
2. Unsorted sortable columns have visible `two-way-arrows` icon.


valentin-mladenov:
follow up of #1612

projects/angular/src/data/datagrid/datagrid-column.ts
```diff
@@ -202,16 +203,12 @@ export class ClrDatagridColumn<T = any>
...
-      if (comparator) {
-        this._sortBy = comparator;
-      } else {
-        if (this.field) {
-          this._sortBy = new DatagridPropertyComparator(this.field);
-        } else {
-          delete this._sortBy;
-        }
-      }
+      delete this._sortBy;
```
kevinbuhmann, line 211: Does this just flatten the `if` blocks? Mixing such refactors with another significant code change can make the change harder to read and understand. Please extract this to a separate PR.

projects/angular/src/data/datagrid/datagrid-column.ts
```diff
@@ -230,17 +227,18 @@ export class ClrDatagridColumn<T = any>
...
         break;
       case ClrDatagridSortOrder.DESC:
         this.sort(true);
         break;
+      // the Unsorted case happens when the current state is neither Asc nor Desc
+      case ClrDatagridSortOrder.UNSORTED:
+      default:
+        this._sort.clear();
+        this._sortDirection = null;
+        break;
```
kevinbuhmann, line 241: This case should be moved back to it's original position to minimize this diff and make the change more obvious without have to mentally compare the code.

Further, there are actually two places with the unsorted/default case listed first. I have fixed both in a separate PR: #1721.

projects/angular/src/data/datagrid/datagrid-column.ts
```diff
@@ -99,6 +99,7 @@ export class ClrDatagridColumn<T = any>
   @Input('clrFilterStringPlaceholder') filterStringPlaceholder: string;
   @Input('clrFilterNumberMaxPlaceholder') filterNumberMaxPlaceholder: string;
   @Input('clrFilterNumberMinPlaceholder') filterNumberMinPlaceholder: string;
+  @Input('clrDgSortDisableUnsort') clrDgSortDisableUnsort = false;
```
kevinbuhmann, line 102:
```suggestion
  @Input('clrDgDisableUnsort') disableUnsort = false;
```

The input alias is redundant (has "sort" twice) and the property name doesn't need the `clrDg` prefix.

kevinbuhmann (changes requested):
See my comments. Please also update the PR summary.
1. It references a `clrDgEnableUnsortColumns` input that does not exist in the current revision.
2. It states that isn't a breaking change, but requiring application code to opt out of unsorting is a breaking change. If we want to release this as a non-breaking change, it needs to be an opt-in feature.

projects/angular/src/data/datagrid/datagrid-column.ts
valentin-mladenov: PR for the change #1777

projects/angular/src/data/datagrid/datagrid-column.ts
valentin-mladenov: corrected in commit 8873189

valentin-mladenov:
WIll not be implemented

PR #1624, opened 2024-11-26
fix(a11y): signpost close aria label update and allow user to config aria label

spadidapelli:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-2461

- [ ] Yes
- [x] No


projects/angular/src/popover/signpost/signpost-content.ts
```diff
@@ -56,6 +56,8 @@ const POSITIONS: string[] = [
   host: { '[class.signpost-content]': 'true', '[id]': 'signpostContentId' },
 })
 export class ClrSignpostContent extends AbstractPopover implements OnDestroy {
+  @Input('clrSignpostCloseAriaLabel') signpostCloseAriaLabel: string;
```
dtsanevmw, line 59: Can we just use it as a default value?
```suggestion
  @Input('clrSignpostCloseAriaLabel') signpostCloseAriaLabel = this.commonStrings.keys.signpostClose;
```

projects/angular/src/popover/signpost/signpost-content.ts
```diff
@@ -40,7 +40,7 @@ const POSITIONS: string[] = [
       <div class="signpost-content-header">
         <button
           type="button"
-          [attr.aria-label]="commonStrings.keys.signpostClose"
+          [attr.aria-label]="signpostCloseAriaLabel || commonStrings.keys.signpostClose"
```
dtsanevmw, line 43:
```suggestion
          [attr.aria-label]="signpostCloseAriaLabel"
```

kevinbuhmann: The current implementation is better so that we default to the common string if a falsy value is bound to the input. This is a pattern we use in several places, and I think this one should be the same.

projects/angular/src/popover/signpost/signpost-content.ts
kevinbuhmann: See my other comment for why defaulting to the common string in the template is better (and consistent with existing code). https://github.com/vmware-clarity/ng-clarity/pull/1624#discussion_r1868210855

projects/angular/src/popover/signpost/signpost-content.ts
spadidapelli: @dtsanevmw They want to customise aria label with name of trigger element, so current implementation is better and we are using same pattern in other places also.

dtsanevmw: We have kind of mixed approaches and while If someone uses the input to bind to something that's 'falsy' is their responsibility I'm okey keeping it like you did it initially @spadidapelli.

dtsanevmw (approved):

PR #1620, opened 2024-11-18
feat(tabs): add actions section

dtsanevmw:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

1: Slot which is being projected inside the header so basically `<clr-tabs-actions>Whatever content you put here</clr-tabs-actions>`. 
2: Directive `clrTabsAction` which you put on the element you want to be interactable and part of the navigation of tabs ( arrows ).

- [ ] Bugfix
- [X] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: CDE-2391

## Does this PR introduce a breaking change?

- [ ] Yes
- [X] No

PR #1629, opened 2024-12-02
fix(tabs): export tabs actions position type

dtsanevmw:
Export tabs actions position type. Missed to add it in #1620


valentin-mladenov (approved):
LGTM

PR #1634, opened 2024-12-03
feat: add support for Angular 19

kevinbuhmann:


waliarubal:
I have forcefully upgraded to angular 19 and I don't see any issues happening yet. The only problem I see here is I can't use npm update command directly without using --force switch. Please provide a roadmap by when we will have official support for angular 19?

valentin-mladenov (approved):
LGTM

No way arround version switch in virtual scroll until droping supprot for old angular versions.

PR #1645, opened 2024-12-13
feat(datagrid): add string filter label

valentin-mladenov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Datagrid filter dialog window have no title.
image

Issue Number: CDE-2483

## What is the new behavior?
Datagrid string filter dialog window have label similar to number filter dialog.
image

- [ ] Yes
- [x] No


projects/angular/src/data/datagrid/_datagrid.clarity.scss
```diff
@@ -1411,10 +1411,20 @@
     font-weight: normal;
 
     .datagrid-filter-close-wrapper {
-      text-align: right;
+      display: flex;
+      justify-content: space-between;
+      align-items: center;
+      gap: #{tokens.$cds-global-space-3};
```
kevinbuhmann, line 1417:
```suggestion
      gap: tokens.$cds-global-space-3;
```

Is interpolation needed here? Check the `margin-bottom` one too.

projects/angular/src/data/datagrid/datagrid-filter.ts
```diff
@@ -155,7 +159,11 @@ export class ClrDatagridFilter<T = any>
 
   ngOnChanges() {
     setTimeout(() => {
-      this.setToggleButtonAriaLabel();
+      const columnTitle = this.getColumnTitle();
+
+      this.setToggleButtonAriaLabel(columnTitle);
+      this.setTitle(columnTitle);
+      this.cdr.markForCheck();
```
kevinbuhmann, line 166: Is this needed with the `sleep` in the story?

projects/angular/src/data/datagrid/datagrid-filter.ts
```diff
@@ -72,6 +73,7 @@ import { KeyNavigationGridController } from './utils/key-navigation-grid.control
       [attr.aria-label]="commonStrings.keys.datagridFilterDialogAriaLabel"
     >
       <div class="datagrid-filter-close-wrapper">
+        <span class="datagrid-filter-title">{{ filterTitle }}</span>
```
kevinbuhmann, line 76: Screen reader users might have difficulties with this not marked up as a heading. I'm not sure how to handle the heading level though.

There is an `aria-label` on the popup element, so maybe this is fine. I just want to make sure.

kevinbuhmann (changes requested):
See comments.

Also, the filter form control labels are larger than the filter popup title. I don't think that's correct.

projects/angular/src/data/datagrid/datagrid-filter.ts
kevinbuhmann: Oh, I see the `sleep` was removed from the story. Nvm.

projects/angular/src/data/datagrid/datagrid-filter.ts
kevinbuhmann: The popup `aria-label` doesn't have the column name. Maybe that should be changed.

valentin-mladenov: The filter trigger already have `aria-label` similar to the filter title: `{COLUMN} filter`. That is already announced to the user via screen reader.
Adding the title ahve more value for the regular user. The title will be read by the screen reader anyway.
First will be `popup aria-label` and then the title, if I'm not mistaken.

projects/angular/src/data/datagrid/_datagrid.clarity.scss
valentin-mladenov: corrected

valentin-mladenov:
> See comments.
> 
> Also, the filter form control labels are larger than the filter popup title. I don't think that's correct.

The filter title is following the signpost title design spec.

valentin-mladenov:
@kevinbuhmann the acceptance criteria is changed. The implementation is totaly different.

projects/angular/src/data/datagrid/built-in/filters/datagrid-string-filter.ts
```diff
@@ -72,6 +78,7 @@ export class DatagridStringFilter<T = any>
    */
   @ViewChild(ClrDatagridFilter) filterContainer: ClrDatagridFilter<T>;
 
+  filterLabel = '';
```
kevinbuhmann, line 81:
```suggestion
  labelValue = '';
```

There is a `label` property and a `filterLabel` property, and the naming is a bit ambiguous. I recommend renaming this property to match `placeholderValue` so the the label and placeholder property naming is consistent.

valentin-mladenov: corrected in commit 42d1bf1

kevinbuhmann (approved):

PR #1651, opened 2024-12-19
chore: remove unnecessary `_dynamic` and `DynamicWrapper`

kevinbuhmann:
The `_dynamic` property was set but never read.


valentin-mladenov:
@kevinbuhmann is there a reason why was created on the first place?

kevinbuhmann:
> @kevinbuhmann is there a reason why was created on the first place?

I have no clue. It might been used at some point, but not anymore.

valentin-mladenov (approved):
LGTM

PR #1661, opened 2025-01-13
fix(combobox): remove underline when no focus

valentin-mladenov:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Sometimes when nested `combobox` focus presists after focus move to other element.

Issue Number: CDE-2531

## What is the new behavior?
Nested `combobox` focus is removed after focus move to other element.

- [ ] Yes
- [x] No


projects/angular/src/forms/combobox/combobox.ts
```diff
@@ -257,14 +257,19 @@ export class ClrCombobox<T>
   }
 
   onBlur() {
-    this.onTouchedCallback();
-    if (this.control.control.updateOn === 'change' && this.control.control?.errors?.required) {
+    this.focused = false;
+
+    if (this.onTouchedCallback) {
+      this.onTouchedCallback();
+    }
```
kevinbuhmann, line 264:
```suggestion
    this.onTouchedCallback?.();
```

You can simplify this.

kevinbuhmann: But why do we need the conditional call? The previous code had an unconditional call.

projects/angular/src/forms/combobox/combobox.ts
```diff
@@ -360,7 +365,7 @@ export class ClrCombobox<T>
     if (this.controlStateService) {
       this.subscriptions.push(
         this.controlStateService.statusChanges.subscribe(invalid => {
-          this.invalid = this.control.control.touched && invalid === CONTROL_STATE.INVALID;
+          this.invalid = this.control?.control.touched && invalid === CONTROL_STATE.INVALID;
```
kevinbuhmann, line 368: Why this change?

projects/angular/src/forms/combobox/combobox.ts
valentin-mladenov: The function is assigned dynamically and in some cases is `undefined` or `null` .

projects/angular/src/forms/combobox/combobox.ts
valentin-mladenov: In some cases `this.control` is `undefined` or `null`.

projects/angular/src/forms/combobox/combobox.ts
```diff
@@ -257,14 +257,19 @@ export class ClrCombobox<T>
...
   onBlur() {
-    this.onTouchedCallback();
-    if (this.control.control.updateOn === 'change' && this.control.control?.errors?.required) {
+    this.focused = false;
+
+    if (this.onTouchedCallback) {
+      this.onTouchedCallback();
+    }
+
+    if (this.control?.control.updateOn === 'change' && this.control?.control?.errors?.required) {
```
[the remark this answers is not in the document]
valentin-mladenov: The initial condition was `if (this.control.control.updateOn === 'change' && this.control.control?.errors?.required) {` I just make sure the `this.control` have value.

I'll introduce check on all places.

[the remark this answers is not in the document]
valentin-mladenov: corrected in commit 254c211

projects/angular/src/forms/combobox/combobox.ts
valentin-mladenov: corrected in commit 254c211

projects/angular/src/forms/combobox/combobox.ts
```diff
@@ -257,14 +257,17 @@ export class ClrCombobox<T>
   }
 
   onBlur() {
-    this.onTouchedCallback();
-    if (this.control.control.updateOn === 'change' && this.control.control?.errors?.required) {
+    this.focused = false;
+
+    this.onTouchedCallback?.();
+
+    if (this.control?.control?.updateOn === 'change' && this.control?.control?.errors?.required) {
```
kevinbuhmann, line 264:
```suggestion
    if (this.control?.control?.updateOn === 'change' && this.control.control.errors?.required) {
```

This has unnecessary conditional access operators. It is *impossible* for these properties to be `undefined` or `null` in the second condition.

I'm going to take a closer look at this because I think there are unnecessary conditional access operators added in this PR that are not needed. The current code doesn't have these (and the one on this line in current is impossible to hit).

Is this an issue in the current code? Or what in this change introduces the need for conditional access? I think I'm missing something.

valentin-mladenov: I misunderstood your statement.

Corrected in commit aef0752

kevinbuhmann (changes requested):
First, this bug has nothing to with being nested in an accordion. It happens when the combobox has no form control set (no `ngModel`, `formControl`, or `formControlName`). Please update the PR title/description accordingly.

Second, there are extraneous conditional property access operators in this change. The fix should be restricted to only the necessary checks.

Last, there is no need to move the `this.focused = true;` line. It works fine at the end of the `onBlur` method.

See this commit for a working fix with minimal changes: b9a0abbffc81cd60e4382490b1fee35944c07042. Please remove the extra unnecessary changes to isolate this bug fix commit to only what's needed to fix the bug.

kevinbuhmann (approved):

kevinbuhmann:
Please update the PR title/summary before merging.

valentin-mladenov:
> First, this bug has nothing to with being nested in an accordion. It happens when the combobox has no form control set (no `ngModel`, `formControl`, or `formControlName`). Please update the PR title/description accordingly.
> 
> Second, there are extraneous conditional property access operators in this change. The fix should be restricted to only the necessary checks.
> 
> Last, there is no need to move the `this.focused = true;` line. It works fine at the end of the `onBlur` method.
> 
> See this commit for a working fix with minimal changes: [b9a0abb](https://github.com/vmware-clarity/ng-clarity/commit/b9a0abbffc81cd60e4382490b1fee35944c07042). Please remove the extra unnecessary changes to isolate this bug fix commit to only what's needed to fix the bug.

1. I don't say is a combobox accordion related bug. Just that is the place I observed it.
2. Separate PR for null refernce checks is created #1670 
3. Corrected in commit 834884a

kevinbuhmann:
>I don't say is a combobox accordion related bug. Just that is the place I observed it.

My problem is the commit message not accurately describing the bug. The bug happened when there was no control, not when nesting as you *did* say in the commit message.

>Separate PR for null refernce checks is created https://github.com/vmware-clarity/ng-clarity/pull/1670

I'm not sure why you split the null reference checks for the same bug into two separate PRs, but it's too late now.

PR #1666, opened 2025-01-13
Fix code scanning alert no. 3: Artifact poisoning

kevinbuhmann:


kevinbuhmann:
I don't think this actually fixes any vulnerability.

PR #1667, opened 2025-01-13
chore: extract visual diffs to temporary directory

kevinbuhmann:
This change simplifies the code a bit because we don't need to delete the diff files anymore.


.github/workflows/pr-visual-snapshot-update-bot.yml
```diff
@@ -25,21 +25,22 @@ jobs:
...
           workflow: ${{github.event.workflow_run.workflow_id}}
           run_id: ${{github.event.workflow_run.id}}
           if_no_artifact_found: ignore
       - name: Apply Changes
         id: apply-changes
         run: |
           # apply changes
-          git apply visual-snapshot-changes-*.diff/visual-snapshot-changes-*.diff
-          rm visual-snapshot-changes-*.diff/visual-snapshot-changes-*.diff
+          git apply ${{runner.temp}}/artifacts/visual-snapshot-changes-*.diff/visual-snapshot-changes-*.diff
```
valentin-mladenov, line 43: There is no point to remove diffs anymore, because now are stored in temp folder that will be removed once the job finishes. Right?

Just resolve this comment if the above statement is correct.

valentin-mladenov (approved):
LGTM

only one clarification comment

.github/workflows/pr-visual-snapshot-update-bot.yml
kevinbuhmann: Yes, that's correct.

PR #1668, opened 2025-01-14
fix(file-input): prevent validation error when `accept` is empty string

kevinbuhmann:
## PR Checklist

## PR Type

Bugfix

File accept error is set when `accept=""`.

## What is the new behavior?

File accept error is not set when `accept=""`.

No.

### DIFF
--- a/projects/angular/src/forms/file-input/file-input-validator.ts
+++ b/projects/angular/src/forms/file-input/file-input-validator.ts
@@ -29,7 +29,7 @@ export class ClrFileInputValidator implements Validator {
       errors.required = true;
     }
 
-    const accept = fileInputElement.accept?.split(',').map(type => type.trim());
+    const accept = fileInputElement.accept ? fileInputElement.accept.split(',').map(type => type.trim()) : null;
 
     if (files?.length > 0 && (accept?.length || this.minFileSize || this.maxFileSize)) {
       for (let i = 0; i < files.length; i++) {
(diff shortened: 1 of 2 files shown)


projects/angular/src/forms/file-input/file-input-validator.ts
```diff
@@ -29,7 +29,7 @@ export class ClrFileInputValidator implements Validator {
       errors.required = true;
     }
 
-    const accept = fileInputElement.accept?.split(',').map(type => type.trim());
+    const accept = fileInputElement.accept ? fileInputElement.accept.split(',').map(type => type.trim()) : null;
```
valentin-mladenov, line 32: Instead of `null` set `accept` to empty array `[]`. Then you will not need the null check.
 `accept?.length` -> `accept.length`

```suggestion
    const accept = fileInputElement.accept ? fileInputElement.accept.split(',').map(type => type.trim()) : [];
```

valentin-mladenov (changes requested):
Check my comment

projects/angular/src/forms/file-input/file-input-validator.ts
kevinbuhmann: I think `null` is more appropriate here. It doesn't accept zero file types. Instead, there is no accept validation at all. I'll remove the `length` check.

kevinbuhmann: This is updated.

valentin-mladenov (approved):
LGTM

PR #1681, opened 2025-01-23
fix(form-label): prevent id attribute from being removed

kevinbuhmann:


projects/angular/src/forms/common/label.ts
```diff
@@ -50,6 +50,10 @@ export class ClrLabel implements OnInit, OnDestroy {
   }
 
   ngOnInit() {
+    // Prevent id attributes from being removed by the `undefined` host binding.
+    // This happens when a `label` is used outside of a control container.
```
dtsanevmw, line 54: It was happening in control containers.

dtsanevmw (approved):

PR #1690, opened 2025-01-29
fix(datagrid): align action items focus/hover/selected with dropdown

kevinbuhmann:
Use the same styles as dropdown items for datagrid action items. This results in focus being indicated with an outline instead of focus and hover using the same background color.

CDE-2565

### DIFF
--- a/projects/angular/src/data/datagrid/STYLES.md
+++ b/projects/angular/src/data/datagrid/STYLES.md
@@ -2,58 +2,58 @@
 
 ## CSS Properties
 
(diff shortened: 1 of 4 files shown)


dtsanevmw (changes requested):
We should sync with design as there is no specification in Figma but I think we should have hover, :active, :focus, plus the additional states that we have on other components of this type if we count focus as selected. These are selected-hover, selected-active. 

I see there is .active class there but not sure when this is activated.

dtsanevmw (approved):

PR #1694, opened 2025-02-03
fix(datagrid): virtual scroller key navigation focus

valentin-mladenov:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Virtual scroller key navigation focus disapears in some edge cases.

Issue Number: CDE-2564

## What is the new behavior?
Virtual scroller key navigation focus disapears only when multitude of pages are loade in quick succession, which is expected.

- [ ] Yes
- [x] No


kevinbuhmann (changes requested):
Please describe the bug and fix in more detail the PR title/summary.

projects/angular/src/data/datagrid/utils/key-navigation-grid.controller.ts
```diff
@@ -224,7 +225,11 @@ export class KeyNavigationGridController implements OnDestroy {
         y = 0;
       }
     } else if (e.key === Keys.PageUp) {
-      y = y - itemsPerPage > 0 ? y - itemsPerPage + 1 : 1;
+      if (this.strategy === 'virtualScroller') {
+        y = numOfRows / 4 < y ? Math.floor(numOfRows / 4) : 1;
```
dtsanevmw, line 229: Why we do division exactly by 4?

valentin-mladenov: It's kinda magic number, but the reason is to put the next Index in the upper 1/4 of the total rows.

The calculation is as follows in virtual scroll.
`numOfRows` are all rendered rows. The count is static and don't change, except in some edge cases that don't matter.
Relativley speaking we have not visible 1/4 above, visible 2/4 and invisible 1/4 under.

`PageUp` tries to scroll up a page placing the index at the above 1/4. That will pull down the row and will make the item visible and centered.

This all is relative and `itemsPerPage` also have weight, but I didn't find stable algorithm to implement.
Using `numOfRows / 4` was the most stable and relaiable in my tests.

valentin-mladenov: Using `numOfRows / 4` is the most stable and relaiable algorithim.

The most important changes are in the `minBufferPx` and `maxBufferPx`. 
Users have to add significant buffer for virtual scroll to update the rows reliably.

kevinbuhmann:
```suggestion
        const threshold = numRows / 4;
        y = y >= threshold ? Math.floor(threshold) : 1;
```

I think the `numRows / 4` should be extracted so that it only happens once in the code. I also flipped the condition because I think it reads better that way, but maybe that's just me.

Please also explain the threshold logic (rename if desired) in a code comment so that it is is available to anyone reading the code.

valentin-mladenov: After close examination `PageUp` beahviour is actually fixed in #1695.

In this PR is left only a slight Key navigation focus optimization, all other changes are reverted.

projects/angular/src/data/datagrid/datagrid-virtual-scroll.directive.spec.ts
```diff
@@ -223,21 +224,30 @@ export default function (): void {
         expect(document.activeElement).toBe(headerCheckboxCell);
 
         grid.dispatchEvent(new KeyboardEvent('keydown', { key: Keys.PageDown }));
+        tick(1);
         // active checkbox input with ID clr-dg-row-cb364
+        console.log(document.activeElement);
+        console.log(grid.querySelectorAll('[type=checkbox]')[22]);
```
kevinbuhmann, line 230:
```suggestion
```

projects/angular/src/data/datagrid/datagrid-virtual-scroll.directive.spec.ts
```diff
@@ -223,21 +224,30 @@ export default function (): void {
...
         expect(document.activeElement).toBe(grid.querySelectorAll('[type=checkbox]')[22]);
 
         grid.dispatchEvent(new KeyboardEvent('keydown', { key: Keys.PageDown }));
         sleep();
+        tick(1);
         fixture.whenStable();
         fixture.whenRenderingDone();
         // active checkbox input with ID clr-dg-row-cb383
+        console.log(document.activeElement);
+        console.log(grid.querySelectorAll('[type=checkbox]')[41]);
```
kevinbuhmann, line 240:
```suggestion
```

projects/angular/src/data/datagrid/datagrid-virtual-scroll.directive.spec.ts
```diff
@@ -223,21 +224,30 @@ export default function (): void {
...
         expect(document.activeElement).toBe(grid.querySelectorAll('[type=checkbox]')[41]);
 
         grid.dispatchEvent(new KeyboardEvent('keydown', { key: Keys.PageUp }));
         sleep();
+        tick(1);
         fixture.whenStable();
         fixture.whenRenderingDone();
         // active checkbox input with ID clr-dg-row-cb360
+        console.log(document.activeElement);
+        console.log(grid.querySelectorAll('[type=checkbox]')[19]);
```
kevinbuhmann, line 250:
```suggestion
```

kevinbuhmann (changes requested):

projects/angular/src/data/datagrid/datagrid-virtual-scroll.directive.spec.ts
valentin-mladenov: Removed in 01d0b44

projects/angular/src/data/datagrid/datagrid-virtual-scroll.directive.spec.ts
valentin-mladenov: Removed in 01d0b44

projects/demo/src/app/datagrid/virtual-scroll-client-side/virtual-scroll-client-side.html
```diff
@@ -81,12 +81,7 @@ <h3>Dynamic rows with sorting</h3>
   />
 </clr-input-container>
 
-<clr-datagrid
-  *ngIf="{ users: users | async }; let data"
-  [(clrDgSelected)]="selectedUsers"
-  class="datagrid-compact"
-  style="height: 30rem"
->
+<clr-datagrid *ngIf="{ users: users | async }; let data" [(clrDgSelected)]="selectedUsers" style="height: 30rem">
```
kevinbuhmann, line 84: This change seems unrelated.

kevinbuhmann:
Please update the PR title and summary.

projects/demo/src/app/datagrid/virtual-scroll-client-side/virtual-scroll-client-side.html
valentin-mladenov: I decided to leave at one of the datagrid in non compact mode.

Can be revered if needed.

kevinbuhmann: It feels like an extraneous change unrelated to the fix, but I could be wrong.

valentin-mladenov:
After some consideration the solution might lead to some unforseen changes in beahviour.

It's better to close it for now and wait for related bugs to be raised.

PR #1545, opened 2024-09-05
fix(datagrid): skip focusing active cell if focus outside component

dtsanevmw:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

If you change the items inside datagrid trough outside filter and previously it was clicked on any cell then the datagrid will steal the focus.

Issue Number: CDE-2273 #1542

## What is the new behavior?

Keep focus on the element

## Does this PR introduce a breaking change?

- [ ] Yes
- [X] No

PR #1695, opened 2025-02-03
fix(datagrid): skip focusing active cell if focus is outside component

valentin-mladenov:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Keep focus on the element, but brake virtual scroll key navigation.

Issue Number: CDE-2564, CDE-2237, follow up of PR #1545

## What is the new behavior?
Keep focus on the element in all cases without side effects.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/data/datagrid/utils/key-navigation-grid.controller.ts
+++ b/projects/angular/src/data/datagrid/utils/key-navigation-grid.controller.ts
@@ -102,7 +102,13 @@ export class KeyNavigationGridController implements OnDestroy {
       fromEvent(this.grid, 'focusout')
         .pipe(takeUntil(this.destroy$))
         .subscribe(() => {
-          this.removeActiveCell();
+          setTimeout(() => {
+            if (this.grid.contains(document.activeElement)) {
+              return;
+            }
+
+            this.removeActiveCell();
+          });
         });
 
       fromEvent(this.grid, 'keydown')
--- a/projects/angular/src/data/datagrid/utils/key-navigation-grid.controller.ts
+++ b/projects/angular/src/data/datagrid/utils/key-navigation-grid.controller.ts
@@ -99,6 +99,12 @@ export class KeyNavigationGridController implements OnDestroy {
           this.removeActiveCell();
         });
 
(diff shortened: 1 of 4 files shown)


projects/angular/src/data/datagrid/utils/key-navigation-grid.controller.ts
```diff
@@ -99,6 +99,12 @@ export class KeyNavigationGridController implements OnDestroy {
           this.removeActiveCell();
         });
 
+      fromEvent(this.grid, 'focusout')
+        .pipe(takeUntil(this.destroy$))
+        .subscribe(() => {
+          this.removeActiveCell();
+        });
```
dtsanevmw, line 106: Is this needed? If they focus out and focus again the datagrid will set different active cell?

valentin-mladenov: It's not only needed, but is the actual fix.
The actual focus is determinated by `tabindex`, when focus again the grid will set it's cell as the active one.

projects/angular/src/data/datagrid/utils/key-navigation-grid.controller.ts
```diff
@@ -99,6 +99,18 @@ export class KeyNavigationGridController implements OnDestroy {
...
+        .subscribe(() => {
+          setTimeout(() => {
+            if (this.grid.contains(document.activeElement)) {
+              return;
+            }
+
+            this.removeActiveCell();
+          });
+        });
+
```
dtsanevmw, line 113: I think this should work too and if we could leave a simple comment explaining what we are catching here would be good 
```suggestion
      fromEvent(this.grid, 'focusout')
        .pipe(debounceTime(0), takeUntil(this.destroy$))
        .subscribe(() => {
          if (this.grid.contains(document.activeElement)) {
            return;
          }

          this.removeActiveCell();
        });
```

dtsanevmw (approved):

PR #1696, opened 2025-02-03
chore(breadcrumbs): add an example with href in storybook and demo

andyfeds:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: [CDE-2550](https://vmw-jira.broadcom.net/browse/CDE-2550)

- [ ] Yes
- [x] No


projects/angular/src/layout/breadcrumbs/breadcrumbs.html
```diff
@@ -17,15 +17,15 @@
         *ngIf="breadcrumb.routerLink"
         [routerLink]="breadcrumb.routerLink"
         [queryParams]="breadcrumb.queryParams"
-        [target]="breadcrumb.target"
+        [target]="breadcrumb.target ? breadcrumb.target : '_self'"
```
kevinbuhmann, line 20:
```suggestion
        [target]="breadcrumb.target || '_self'"
```

projects/angular/src/layout/breadcrumbs/breadcrumbs.html
```diff
@@ -17,15 +17,15 @@
...
         (click)="handleItemClick(breadcrumb)"
       >
         {{ breadcrumb.label }}
       </a>
       <a
         *ngIf="breadcrumb.href"
         [href]="breadcrumb.href"
-        [target]="breadcrumb.target"
+        [target]="breadcrumb.target ? breadcrumb.target : '_self'"
```
kevinbuhmann, line 28:
```suggestion
        [target]="breadcrumb.target || '_self'"
```

kevinbuhmann (commented):
The default target change is a behavior change in the component, so this is not just a `chore`. The examples do more than just demo the bug fix, so this should probably be two separate PRs.

kevinbuhmann (dismissed):
See comments.

andyfeds:
> The default target change is a behavior change in the component, so this is not just a `chore`. The examples do more than just demo the bug fix, so this should probably be two separate PRs.

Yeah I initially thought about doing that, ill split it into two PR's

andyfeds:
Added the default target change to a new PR: https://github.com/vmware-clarity/ng-clarity/pull/1697

kevinbuhmann (approved):

projects/demo/src/app/breadcrumbs/breadcrumbs-routing.demo.ts
```diff
@@ -0,0 +1,54 @@
...
+    console.log('Menu:', this.menuItems);
+    this.router.events
+      .pipe(
+        filter(event => event instanceof NavigationEnd),
+        distinctUntilChanged()
+      )
+      .subscribe(() => (this.menuItems = this.createBreadcrumbs(this.activatedRoute.root)));
+  }
+
+  createBreadcrumbs(route: ActivatedRoute, url = '', breadcrumbs: MenuItem[] = []): MenuItem[] {
```
kevinbuhmann, line 34: This can be a `function` outside of the component class.

projects/demo/src/app/breadcrumbs/breadcrumbs-routing.demo.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  constructor(private router: Router, private activatedRoute: ActivatedRoute) {}
+
+  ngOnInit(): void {
+    this.menuItems = this.createBreadcrumbs(this.activatedRoute.root);
+    console.log('Menu:', this.menuItems);
+    this.router.events
+      .pipe(
+        filter(event => event instanceof NavigationEnd),
+        distinctUntilChanged()
```
kevinbuhmann, line 29: I don't think this is needed.

projects/demo/src/app/breadcrumbs/breadcrumbs-routing.demo.ts
```diff
@@ -0,0 +1,54 @@
...
+      )
+      .subscribe(() => (this.menuItems = this.createBreadcrumbs(this.activatedRoute.root)));
```
kevinbuhmann, line 31:
```suggestion
    this.menuItems = this.router.events.pipe(
      filter(event => event instanceof NavigationEnd),
      startWith(undefined),
      map(() => createBreadcrumbs(this.activatedRoute.root))
    );
```

The subscription is never unsubscribed. This can be simplified though so that we use the `async` pipe instead of an explicit subscription.

kevinbuhmann (changes requested):
See my comments.

projects/demo/src/app/breadcrumbs/breadcrumbs-routing.demo.ts
kevinbuhmann: See eb7e794f0106efa4cbb3b13b5ffe23753ca5b35f for an example refactor.

projects/demo/src/app/breadcrumbs/breadcrumbs-routing.demo.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  createBreadcrumbs(route: ActivatedRoute, url = '', breadcrumbs: MenuItem[] = []): MenuItem[] {
+    const children: ActivatedRoute[] = route.children;
+
+    if (children.length === 0) {
+      return breadcrumbs;
+    }
+
+    for (const child of children) {
```
kevinbuhmann, line 41: Please refactor this without this loop. There's no need to loop when there's an unconditional return in the first iteration. Also, an iteration approach would probably be more straightforward than this recursive approach.

projects/demo/src/app/breadcrumbs/breadcrumbs-routing.demo.ts
andyfeds: Moved it to a service : `breadcrumb.demo.service.ts`

projects/demo/src/app/breadcrumbs/breadcrumbs-routing.demo.ts
andyfeds: Made changes are suggested in https://github.com/vmware-clarity/ng-clarity/commit/eb7e794f0106efa4cbb3b13b5ffe23753ca5b35f

projects/demo/src/app/breadcrumbs/breadcrumbs-routing.demo.ts
kevinbuhmann: Why? It's just a function that is used in one place. There is no need for a service class.

kevinbuhmann: And it's still a method that doesn't use `this` except to call itself. It should be a `function`. The only other member of the service class is a `BehaviorSubject` that can be part of the component class. (It should be a `Observable` though.)

projects/demo/src/app/breadcrumbs/breadcrumb.demo.service.ts
```diff
@@ -0,0 +1,38 @@
...
+import { ActivatedRoute } from '@angular/router';
+import { BehaviorSubject } from 'rxjs';
+
+import { MenuItem } from './breadcrumbs.demo.model';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class BreadcrumbDemoService {
+  breadcrumbs: BehaviorSubject<MenuItem[]> = new BehaviorSubject<MenuItem[]>([]);
```
kevinbuhmann, line 18: This is unused.

projects/demo/src/app/breadcrumbs/breadcrumb.demo.service.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  createBreadcrumbs(route: ActivatedRoute, url = '', breadcrumbs: MenuItem[] = []): MenuItem[] {
```
kevinbuhmann, line 20: This should just a be `function` as discussed in the other thread. There is no need for a class here. And it doesn't really need to be a separate file either.

projects/demo/src/app/breadcrumbs/breadcrumbs-routing.demo.ts
kevinbuhmann: Wait, I just realized that this was all existing code. You can (should) just put it back how it was in the `BreadcrumbsDemo` component. I will open a PR with my clean up ideas.

kevinbuhmann (approved):

Issue #1692, opened 2025-01-31
The Clarity wizard page shows with an offset instead of starting from the beginning


edeyanova:
## Describe the bug

The Clarity wizard page shows with an offset instead of starting from the beginning.

## How to reproduce

https://stackblitz.com/edit/stackblitz-starters-i8zgv4sd

1. Click the wizard button.
2. Scroll down to the middle of the first page.
3. Click the Next button.
4. Notice that the scroller is positioned the same as it was on the previous page.

## Expected behavior

The scroll bar is positioned at the top of the page.

## Versions

**Clarity version:**

- v17.x

**Framework version:**
Angular 16

**Device:**

- Type: any
- OS: any
- Browser any

LynxDelli:
This is by design. If you go to the next step, you will see an icon; the offset is for icons.

kevinbuhmann:
This is not by design. I will open a PR to fix.

PR #1698, opened 2025-02-03
fix(wizard): scroll to top of page when page is changed

kevinbuhmann:
## PR Type

Bug fix

When you navigate between pages in a wizard, the page is not scrolled to the top. This can be an issue when the pages have long content because you can start in the middle/end of the next page instead of the top.

Issue Number: CDE-2576, #1692 

## What is the new behavior?

The wizard page will be scroll to the top upon navigation between pages.

No.


andyfeds (approved):
Looks good to me

PR #1709, opened 2025-02-19
feat(wizard): add in-page wizard

kevinbuhmann:
## PR Checklist

## PR Type

Feature

Wizards must be shown in a modal. We have support for "inline" wizards, but those intended to be shown in an iframe that's in a modal. As such, inline wizards do not fill the whole page, use F-Pattern buttons, or prevent close.

Issue Number: CDE-2578

## What is the new behavior?

A wizard can be use in "in page" mode which can fill the page, uses F-Pattern buttons, and cannot be closed.

No.


projects/angular/src/modal/modal.html
```diff
@@ -22,10 +22,8 @@
     [attr.aria-labelledby]="labelledBy || modalId"
   >
     <div class="clr-sr-only">{{commonStrings.keys.modalContentStart}}</div>
-    <div class="modal-content-wrapper">
-      <!-- only used in wizards -->
-      <ng-content select=".modal-nav"></ng-content>
```
valentin-mladenov, line 27: `.modal-nav` is public CSS class. You cannot remove it from here.

valentin-mladenov (changes requested):
Check my comments.

There are a lot of changes in the modal that should not exist.

projects/angular/src/wizard/wizard.html
```diff
@@ -5,58 +5,91 @@
...
+          >
+            <ng-content select="clr-wizard-button"></ng-content>
+          </div>
+          <div
+            *ngIf="navService.currentPage && navService.currentPage.hasButtons"
+            class="clr-wizard-footer-buttons-wrapper"
+          >
+            <ng-template [ngTemplateOutlet]="navService.currentPage.buttons"></ng-template>
+          </div>
+        </div>
```
valentin-mladenov, line 91: Copying most of modal's HTML in wizard HTML is not a proper solution.

projects/angular/src/wizard/_wizard.clarity.scss
```diff
@@ -510,6 +511,19 @@
     }
   }
 
+  .clr-wizard--in-page {
+    display: block;
+    height: calc(100% + #{variables.$clr-content-area-padding-top} + #{variables.$clr-content-area-padding-bottom});
+    width: unset; // 100% from the inline wizard styles does not work
+
+    // offset content area padding
+    margin: -#{variables.$clr-content-area-padding-top} -#{variables.$clr-content-area-padding-right} -#{variables.$clr-content-area-padding-bottom} -#{variables.$clr-content-area-padding-left} !important;
```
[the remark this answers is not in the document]
valentin-mladenov: Removing all of those CSS is a better solution check my other comment for it.

projects/angular/src/wizard/wizard.ts
```diff
@@ -47,6 +47,7 @@ import { ClrWizardTitle } from './wizard-title';
     '[class.wizard-lg]': "size == 'lg'",
     '[class.wizard-xl]': "size == 'xl'",
     '[class.lastPage]': 'navService.currentPageIsLast',
+    '[class.clr-wizard--in-page]': 'inPage',
```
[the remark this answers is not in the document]
kevinbuhmann: I matched the existing `.clr-wizard--inline` class.

projects/angular/src/modal/modal.ts
```diff
@@ -150,8 +153,4 @@ export class ClrModal implements OnChanges, OnDestroy {
       this.modalStackService.trackModalClose(this);
     }
   }
-
-  scrollTop() {
-    this.bodyElementRef.nativeElement.scrollTo(0, 0);
-  }
```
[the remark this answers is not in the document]
kevinbuhmann: It is not documented for external use. It was only public so that the wizard could call it.

projects/angular/src/modal/modal.html
kevinbuhmann: I disagree because this slot was never documented for public use. It is an internal wizard implementation detail that is no longer needed.

projects/angular/src/modal/modal.ts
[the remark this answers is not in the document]
kevinbuhmann: But I added it back in case someone has found it and used it.

projects/angular/src/wizard/wizard.ts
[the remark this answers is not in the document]
valentin-mladenov: `.clr-wizard--inline` and other classes of this type are externaly provided to the wizard.
`.clr-wizard--in-page` is internal.

I generally don't like the duality of the class naming convention that we have and that's why I suggested the change.

Resolve the conversation if it make more sense to keep your way.

projects/angular/src/wizard/_wizard.clarity.scss
```diff
@@ -512,11 +512,15 @@
   }
 
   .clr-wizard--in-page {
-    display: block;
-    height: calc(100% + #{variables.$clr-content-area-padding-top} + #{variables.$clr-content-area-padding-bottom});
+    // extra specificity is needed here to overcome the following default style without using !important:
+    // .main-container .content-container .content-area > :first-child { margin-top: 0px; }
+    &#{&}#{&}.clr-wizard--in-page-fill-content-area {
```
valentin-mladenov, line 517: What's the point ot have `&#{&}#{&}` it will be converted to `.clr-wizard--in-page.clr-wizard--in-page.clr-wizard--in-page-fill-content-area`?

````
&.clr-wizard--in-page-fill-content-area {
```

projects/angular/src/wizard/_wizard.clarity.scss
```diff
@@ -510,6 +511,22 @@
     }
   }
 
+  .clr-wizard--in-page {
+    // extra specificity is needed here to overcome the following default style without using !important:
+    // .main-container .content-container .content-area > :first-child { margin-top: 0px; }
+    &#{&}#{&}.clr-wizard--in-page-fill-content-area {
```
valentin-mladenov, line 517: What's the point ot have `&#{&}#{&}` it will be converted to `.clr-wizard--in-page.clr-wizard--in-page.clr-wizard--in-page-fill-content-area`?

Single `&` is enough.
```suggestion
    &.clr-wizard--in-page-fill-content-area {
```

projects/angular/src/modal/modal.html
valentin-mladenov: but the `.modal-nav` class is documented in the modal API.

valentin-mladenov (changes requested):
check my comments.

projects/angular/src/modal/modal.html
kevinbuhmann: It says, "Modal nav class used only in wizards."

projects/angular/src/wizard/_wizard.clarity.scss
kevinbuhmann: A single `&` is not enough. I explained why in my code comment:

```
// extra specificity is needed here to overcome the following default style without using !important:
// .main-container .content-container .content-area > :first-child { margin-top: 0px; }
```

projects/angular/src/modal/modal.html
kevinbuhmann: It's also documented as a CSS class which is explicitly not public API. It is not documented as a slot. I removed the slot, not the CSS class.

kevinbuhmann: The `.modal-nav` CSS class is documented as "Modal nav class used only in wizards." That is still true. I removed the slot (which is _not_ documented), not the CSS class.

valentin-mladenov: Then you should remove from `_modal.clarity.scss` those lines also.
````
  // modal-nav is only used in wizards
  .modal .modal-nav {
    display: none;
  }
```

projects/angular/src/wizard/_wizard.clarity.scss
valentin-mladenov: Check the follow up comment with suggestion: https://github.com/vmware-clarity/ng-clarity/pull/1709#discussion_r2063877749

projects/angular/src/modal/modal.html
kevinbuhmann: I don't agree with that either because if a `.modal-nav` element is somehow present in a non-wizard modal, it should still be hidden.

projects/angular/src/wizard/_wizard.clarity.scss
```diff
@@ -510,6 +511,22 @@
...
+
+      // offset content area padding
+      margin: -#{variables.$clr-content-area-padding-top} -#{variables.$clr-content-area-padding-right} -#{variables.$clr-content-area-padding-bottom} -#{variables.$clr-content-area-padding-left};
+    }
+
+    .clr-wizard-footer-buttons-wrapper {
+      justify-content: flex-start;
+    }
+  }
+
```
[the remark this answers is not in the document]
kevinbuhmann: updated in 05792c0e0f99162b9cef43ca6643f696da68b9f2

valentin-mladenov (approved):
LGTM

PR #1712, opened 2025-02-20
fix(wizard): improve accessibility of step navigation

kevinbuhmann:
- The step nav is now a `nav` element.
- The step nav `aria-label` is now on the `nav` element and reworded.
- The step nav list now uses the `ol` and `li` elements for ordered list roles.
- The step nav buttons now have `aria-labelledby` to prevent individual focus on separate parts of the button text.

This change may require changes to custom wizard nav labels.

CDE-2595


projects/angular/src/utils/i18n/common-strings.default.ts
```diff
@@ -110,7 +110,7 @@ export const commonStringsDefault: ClrCommonStrings = {
   wizardStep: 'Step',
   wizardStepSuccess: 'Completed',
   wizardStepError: 'Error',
-  wizardStepnavAriaLabel: 'Step navigation',
+  wizardStepnavAriaLabel: 'Wizard steps',
```
dtsanevmw, line 113: Is this enough context for users to understand what 'Wizard' is?

projects/angular/src/wizard/wizard-stepnav-item.spec.ts
```diff
@@ -270,15 +270,32 @@ export default function (): void {
         it('should have id', () => {
           expect(myStepnavItem.hasAttribute('id')).toBeTruthy('stepnav item should have an id');
           const myId = myStepnavItem.getAttribute('id');
-          expect(myId).toBe(testItemComponent.id, 'stepnav item id should contain id');
+          expect(myId).toBe('mock-id', 'stepnav item id should contain id');
```
dtsanevmw, line 273: Can we use 'getStepItemIdForPage' that comes from pageCollection.getStepItemIdForPage() so we don't have hardcoded id that not clear where it comes from?

kevinbuhmann: I don't think we should. In my opinion, the expected value should *always* be a literal value instead of a computed value. Otherwise, the test isn't actually tested against a known expected value.

kevinbuhmann: I don't think we should. In my opinion, the expected value should *always* be a literal value instead of a computed value. Otherwise, the test isn't actually tested against a known expected value.

kevinbuhmann: I don't think we should. In my opinion, the expected value should always be a literal value instead of a computed value. Otherwise, the test isn't actually testing against a known expected value.

dtsanevmw: Yeah but 'should have id' is different from checking if exact value was set. Maybe we can rephrase the test.

kevinbuhmann: I personally don't think that's an issue. I wouldn't expect a "should have id" to just test that the id is present. I'd expect it to test the value as well.

projects/angular/src/utils/i18n/common-strings.default.ts
kevinbuhmann: I was asked by Guru to remove "navigation" from the nav label because it was redundant.

dtsanevmw (approved):

PR #1722, opened 2025-03-04
fix(side-panel): pinning with default modal size

Jinnie:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

If size is not explicitly specified through `clrSidePanelSize`, the pined side panel keeps overlapping the content.
Issue Number: CDE-2602

## What is the new behavior?

The side panel is displayed side bu side with the content, even if `clrSidePanelSize` is not explicitly specified.

- [ ] Yes
- [x] No


Jinnie:
We talked with @valentin-mladenov that the underlying problem that the default size is not explicitly set, but still reported as "undefined". We'd better fix the default to be "md". This should be done in the Modal API, which is currently broken. As per documentation, the default should be "md", not "undefined" with md-like styling.
So, I'll change this to work with non-undefined default.

valentin-mladenov (approved):
LGTM

PR #1736, opened 2025-03-10
chore(breadcrumbs): clean up breadcrumbs router link demo

kevinbuhmann:
- use observable for menu items
- use first-class function to generate breadcrumbs from route
- remove some duplicate code by using variables
- remove unnecessary `for` loop that just returned in the first iteration
- rewrite function to remove recursion


andyfeds (approved):
LGTM

PR #1737, opened 2025-03-11
chore(ci): add preview build branches

kevinbuhmann:
CDE-2618


scripts/update-prebuilt-branches.sh
```diff
@@ -0,0 +1,35 @@
+COMMIT_SHA=$(git log -1 --format='%H')
+
+# configure git
+git config user.name "Clarity Update Bot"
+git config user.email "noreply@github.com"
+
+# fetch prebuilt branch remotes
+git fetch origin main-prebuilt-clr-ui
+git fetch origin main-prebuilt-clr-angular
```
valentin-mladenov, line 9: Why do we need 2 branches?

PR #1781, opened 2025-04-10
feat(side-panel): add bottom positioning

idonchev-vmw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-2607

## What is the new behavior?
Side panels can be displayed at the bottom of the page or container.

- [ ] Yes
- [x] No


projects/angular/src/modal/_modal.clarity.scss
```diff
@@ -14,29 +14,50 @@
...
+  }
+
   .clr-modal-host {
     overflow: hidden;
     position: relative;
     .modal,
     .modal-backdrop {
       position: absolute;
     }
+    box-sizing: border-box;
```
idonchev-vmw, line 60: I don't remember why I added this. Might have been leftover of a temporary research.
Must check.

idonchev-vmw: Yes, not needed. Removed it.

projects/angular/src/modal/_modal.clarity.scss
```diff
@@ -14,22 +14,42 @@
...
   }
 
-  .clr-side-panel-pinned-xl {
+  .clr-side-panel-pinned-right-xl {
     padding-right: modal-variables.$clr-modal-xl-width !important;
   }
 
+  :root {
+    --modal-height-width-ratio: 0.5;
+  }
```
dtsanevmw, line 35: You can declare it as a SASS variable if we are going to use it only int he file for calculations

projects/angular/src/modal/_modal.clarity.scss
```diff
@@ -97,6 +116,34 @@
...
       }
     }
+
+    &.side-panel-bottom {
+      .modal:not(.modal-full-screen) {
+        top: initial;
+        width: 100%;
+        justify-content: end;
+        max-height: 100%;
+        min-height: calc(modal-variables.$clr-modal-sm-width * 0.6);
```
dtsanevmw, line 126: Should we use "modal-height-width-ratio" here instead of 0.6 too?

projects/angular/src/modal/_modal.clarity.scss
```diff
@@ -97,6 +116,34 @@
...
+        &:has(.modal-lg) {
+          height: calc(modal-variables.$clr-modal-lg-width * var(--modal-height-width-ratio));
+        }
+
+        &:has(.modal-xl) {
+          height: calc(modal-variables.$clr-modal-xl-width * var(--modal-height-width-ratio));
+        }
+      }
+      .modal-dialog {
+        max-height: 100%;
```
dtsanevmw, line 143: Do we need it for second time if we set it to .modal class?

projects/angular/src/modal/_modal.clarity.scss
idonchev-vmw: We need a minimum size which will allow to show Title and Buttons, without compressing any of them.
Without this min-size, we'd observe compression/cropping. Maybe we can have a more meaningful formula for the size, but I preferred it to be following the other formula.
If UX provides us with fixed values in future, we will replace all these with token values.

projects/angular/src/modal/_modal.clarity.scss
idonchev-vmw: You are right - this is a leftover from the first implementation, when the height was applied on the .moda-dialog instead of the .modal

projects/angular/src/modal/side-panel.ts
```diff
@@ -113,8 +131,16 @@ export class ClrSidePanel implements OnInit, OnDestroy, OnChanges {
...
+  @HostBinding('class.side-panel-bottom')
+  private get bottomPositionCssClass() {
+    return this.position === 'bottom';
+  }
+
   ngOnInit(): void {
     this.configuration.fadeMove = 'fadeLeft';
+    if (this.position === 'bottom') {
+      this.configuration.fadeMove = 'fadeUp';
+    }
```
dtsanevmw, line 143: Do we need that both here and in the setter above?

projects/angular/src/modal/_modal.clarity.scss
idonchev-vmw: Changed

projects/angular/src/modal/side-panel.ts
idonchev-vmw: Yes, I remember trying only with the setter and it didn't work as expected.

projects/angular/src/modal/_modal.clarity.scss
dtsanevmw: Can you leave a comment or use a token that can clarify what is 0.6 exactly.

.storybook/stories/modal/side-panel-inline.stories.ts
```diff
@@ -156,11 +156,16 @@ export const SidePanelExtraLarge: StoryObj = {
 
 export const SidePanelBottomSmall: StoryObj = {
   render: InlineSidePanelTemplate,
-  play: removeFocusOutline,
+  play: ({ canvasElement }) => {
+    removeFocusOutline({ canvasElement });
+    (canvasElement.querySelector('button.pinnable') as HTMLElement).click();
+  },
```
dtsanevmw, line 162: Could we create a function that can be used for these or expose "pinned" as Input ?

dtsanevmw (commented):
Now all the bottom stories are defaulted to backdrop: false and pinnable: true which is different from the default stories for right positioning.

.storybook/stories/modal/side-panel-inline.stories.ts
idonchev-vmw: Done

projects/angular/src/modal/_modal.clarity.scss
idonchev-vmw: Done

projects/angular/src/modal/modal-host.component.ts
```diff
@@ -0,0 +1,27 @@
...
+  template: `
+    <div *ngIf="scrollable" class="clr-modal-host-scrollable">
+      <ng-container *ngTemplateOutlet="modalContent"></ng-container>
+    </div>
+
+    <ng-container *ngIf="!scrollable" [ngTemplateOutlet]="modalContent"></ng-container>
+
+    <ng-template #modalContent>
+      <ng-content></ng-content>
+    </ng-template>
```
dtsanevmw, line 22: `scrollable` variable seems to not be used and the template can be optimised to be in one variant.

projects/angular/src/modal/side-panel.ts
```diff
@@ -109,30 +120,44 @@ export class ClrSidePanel implements OnInit, OnDestroy, OnChanges {
...
+
   private get hostElement(): HTMLElement {
     return (this.element.nativeElement as HTMLElement).closest('.clr-modal-host') || document.body;
   }
 
-  ngOnInit(): void {
-    this.configuration.fadeMove = 'fadeLeft';
+  @HostBinding('class.side-panel-bottom')
+  private get bottomPositionCssClass() {
+    return this.position === 'bottom';
```
dtsanevmw, line 140: Can we add it directly to the meta data of the component under host: {}?
```suggestion
'[class.side-panel-bottom]': 'position === 'bottom',
```

projects/angular/src/modal/side-panel.ts
```diff
@@ -158,15 +183,31 @@ export class ClrSidePanel implements OnInit, OnDestroy, OnChanges {
...
 
-  private displaySideBySide() {
-    this.hostElement.classList.add('clr-side-panel-pinned-' + this.size);
+  private updateModalState() {
+    if (!this.modal) {
+      return;
+    }
+    if (this.pinned) {
+      this.modal.stopClose = true;
+      this.sideBySideMargins();
```
dtsanevmw, line 192: Would it be better to name the function 'updateHostSize' (or something similar ) as it's updating the size class of the host element.
```suggestion
      this.updateHostSize();
      ```

File: projects/angular/src/modal/side-panel.ts [hunk 10]
```diff
@@ -158,15 +183,31 @@ export class ClrSidePanel implements OnInit, OnDestroy, OnChanges {
...
+      this.cleanupMargins();
+    }
   }
 
-  private displayOverlapping() {
-    this.hostElement.classList.forEach(className => {
-      if (className.startsWith('clr-side-panel-pinned-')) {
-        this.hostElement.classList.remove(className);
-      }
+  private cleanupMargins() {
```
dtsanevmw, line 199: Just a suggestion as it's more straightforward for what it's doing and this classes are adding padding currently.
```suggestion
  private cleanupSizingClasses() {
```

dtsanevmw (commented):
I added some suggestions. Do we need also any unit tests for the modalHost implementation?

projects/angular/src/modal/modal-host.component.ts
idonchev-vmw: This has to be configurable Input. 
The idea is that some host elements, like the .content-area, are scrollable by default and adding additional scrollable child will only mess them.
In the same time, non-scrollable host elements need to be made scrollable by the directive.
I've forgotten to add the Input, I'll add it now, thanks for catching this.

projects/angular/src/modal/side-panel.ts
idonchev-vmw: Will proceed in style with the decision taken on the above question.

projects/angular/src/modal/side-panel.ts
dtsanevmw: I proposed it to keep it consistent with that we already have a class set like that here and also on other components we are generally using that for classes

projects/angular/src/modal/side-panel.ts
idonchev-vmw: Done -> updatePinnedClassses

projects/angular/src/modal/side-panel.ts
idonchev-vmw: Done -> cleanupPinnedClassses

dtsanevmw (approved):

PR #1793, opened 2025-04-23
fix(datagrid): datagrid filter button aria-label fix

spadidapelli:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-2669

- [ ] Yes
- [X] No

### DIFF
--- a/projects/angular/src/data/datagrid/datagrid-filter.ts
+++ b/projects/angular/src/data/datagrid/datagrid-filter.ts
@@ -7,6 +7,7 @@
 
 import { isPlatformBrowser } from '@angular/common';
 import {
+  ChangeDetectorRef,
   Component,
   ElementRef,
   EventEmitter,
@@ -110,7 +111,8 @@ export class ClrDatagridFilter<T = any>
     private smartToggleService: ClrPopoverToggleService,
     @Inject(PLATFORM_ID) private platformId: any,
     private elementRef: ElementRef<HTMLElement>,
-    @Optional() private keyNavigation: KeyNavigationGridController
+    @Optional() private keyNavigation: KeyNavigationGridController,
+    private changeDetectorRef: ChangeDetectorRef
   ) {
     super(_filters);
     this.subs.push(
@@ -177,5 +179,6 @@ export class ClrDatagridFilter<T = any>
     this.toggleButtonAriaLabel = this.commonStrings.parse(this.commonStrings.keys.datagridFilterAriaLabel, {
       COLUMN: columnTitle || '',
     });
+    this.changeDetectorRef.detectChanges();
   }
 }


projects/angular/src/data/datagrid/datagrid-filter.ts
```diff
@@ -177,5 +179,6 @@ export class ClrDatagridFilter<T = any>
     this.toggleButtonAriaLabel = this.commonStrings.parse(this.commonStrings.keys.datagridFilterAriaLabel, {
       COLUMN: columnTitle || '',
     });
+    this.changeDetectorRef.detectChanges();
```
idonchev-vmw, line 182: You shouldn't need this, as the method is only run from within a setTimeout block, which triggers change detection:
```
  ngOnChanges() {
    setTimeout(() => {
      this.setToggleButtonAriaLabel();
    });
  }
  ```

**Review by idonchev-vmw — COMMENTED** [review]
The visual change that gets introduced is a bit concerning. As it does not have anything to cause it in your code, can you make sure your PR is up-to-date with the code from the main repo?
I am also not able to reproduce the original problem with the current state of your fork/branch, as storybook does not start at all. But on the demo app, it's also not reproducible when reverting your change.

**Comment by idonchev-vmw** [comment]
Note, the visual test fail is not related to your code, we have it on other PRs too. We're working to resolve it.

**Comment by idonchev-vmw** [comment]
Reviewing deeply, I realize that the current suggestion will patch the problem it is addressing. But we still have serious underlying problems.
My suggestion for changes to resolve the actual problem is like this:

1. Create a ColumnNameService that has getter/setter for name and change EventEmitter.
2. Provide the service in datagrid-column.
3. Move code that resolves the columnName from content to the column itself
    - Preferably use a ViewChild resolving to ElementRef instead of querySelectors.
4. In ngAfterViewInit, resolve the column name and update it to the service.
5. In datagrid-filter's OnInit subscribe to the ColumnNameService.change
    - still have to make a changeDetectorRef.detectChanges call, but only one, in the handler.

What do we gain:
- we drop the setTimeout
- we drop the querySelectors
- we drop the OnChanges handler
- we have a more standard flow of the data and relationship between the components

**Comment by idonchev-vmw** [comment]
Tests are failing because they need the new service to be provided.

**Review by idonchev-vmw — APPROVED** [review]

File: projects/angular/src/data/datagrid/datagrid-filter.ts [hunk 2]
```diff
@@ -153,29 +156,19 @@ export class ClrDatagridFilter<T = any>
...
-    });
+  ngOnInit() {
+    this.subs.push(
+      this.columnNameService.change.subscribe(name => {
+        this.toggleButtonAriaLabel = this.commonStrings.parse(this.commonStrings.keys.datagridFilterAriaLabel, {
+          COLUMN: name,
+        });
+        this.changeDetectorRef.detectChanges();
+      })
+    );
```
valentin-mladenov, line 167: Why not push the subscription with the other in the constructor?

valentin-mladenov (approved):
LGTM

Check my comment.

projects/angular/src/data/datagrid/datagrid-filter.ts
valentin-mladenov: ... or the other way around push them all in on init.

dtsanevmw (changes requested):
We discussed an alternative variant that I'll provide in a bit.

PR #1799, opened 2025-04-29
fix(accordion): opened content background color

valentin-mladenov:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?
Curent token used `cds-alias-object-app-background`.

Issue Number: CDE-2676

## What is the new behavior?
New token used `cds-alias-object-container-background`.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #1803, opened 2025-05-07
chore: make visual regression threshold more strict

valentin-mladenov:
Visual regression threshold needs to be stricter because some very slight color changes are not calculated.

related to #1799


valentin-mladenov:
Threshold is too strict.

PR #1787, opened 2025-04-14
fix(datagrid): key navigation for `ctrl+end`

valentin-mladenov:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?
Using `Ctrl+End` key navigation in datagrid with last expanded row focus will remain on the currently focuesd cell.

Issue Number: CDE-2664

## What is the new behavior?
Using `Ctrl+End` key navigation in datagrid with last expanded row focus will navigate to bottom rigth cell.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #1810, opened 2025-05-09
fix(datagrid): `Up` and `Down` arrow key navigation

valentin-mladenov:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Navigating with Up/Down Arrow keys over action cells will jump rows. Regression added by #1787 

Issue Number: CDE-2664

## What is the new behavior?
Navigating with Up/Down Arrow keys over action cells will land on proper row.

- [ ] Yes
- [x] No


projects/angular/src/data/datagrid/utils/key-navigation-grid.controller.spec.ts
dtsanevmw, line 1: Comparing text content won't be efficient as it can happen to have equal content inside. I think I saw the errors at one time and toEqual should work probably for deep comparison.

valentin-mladenov: We are interested in `.toBe()` because we want to compare references not vaues.

`.textContent` is removed and `compareActiveElementToElement` method is introduced in commit: 725d614

dtsanevmw (approved):
Additionally we can use "compareToActiveElement(element)" as a better wording. But it's fine either way

valentin-mladenov:
> Additionally we can use "compareToActiveElement(element)" as a better wording. But it's fine either way
method renamed in commit 50f2f6a

projects/angular/src/data/datagrid/utils/key-navigation-grid.controller.spec.ts
```diff
@@ -52,6 +52,10 @@ class TestComponent {
 
 export default function (): void {
   describe('Key navigation controller', function () {
+    function compareToActiveElement(element: HTMLElement): void {
```
kevinbuhmann, line 55: I would have named this `expectActiveElementToBe` because it's an expectation. I'd expect a "compare" method to return a boolean, not throw an error. Also, this name would keep make the test assertions easier to identify by using "expect" consistently.

PR #1831, opened 2025-05-22
fix(wizard): scroll step into view when current step changes

kevinbuhmann:
## PR Checklist

## PR Type

Bugfix

If there are enough steps to make the stepnav scroll, the current step can be out of view.

Issue Number: N/A

## What is the new behavior?

If there are enough steps to make the stepnav scroll, the current step will be scrolled into view to ensure it is always visible.

No.


projects/angular/src/wizard/wizard-stepnav-item.ts
```diff
@@ -158,4 +170,30 @@ export class ClrWizardStepnavItem {
...
+      tap(currentPage => {
+        if (
+          currentPage === this.page &&
+          !elementIsScrolledIntoView({ container: stepnavWrapperElement, element: stepnavItemElement })
+        ) {
+          stepnavItemElement.scrollIntoView({ behavior: 'smooth', block: 'center' });
+        }
+      })
+    );
+  }
```
dtsanevmw, line 188: We should skip going into tap if the page isn't current so we don't have to execute "elementIsScrolledIntoView" for all the pages on each change
```suggestion
    return this.navService.currentPageChanged.pipe(
      filter(currentPage => currentPage === this.page),
      tap(currentPage => {
        if (!elementIsScrolledIntoView({ container: stepnavWrapperElement, element: stepnavItemElement })
        ) {
          stepnavItemElement.scrollIntoView({ behavior: 'smooth', block: 'center' });
        }
      })
    );
  }
```

projects/angular/src/wizard/wizard-stepnav-item.ts
```diff
@@ -158,4 +170,30 @@ export class ClrWizardStepnavItem {
...
+}
+
+function elementIsScrolledIntoView({ container, element }: { container: HTMLElement; element: HTMLElement }) {
```
valentin-mladenov, line 191: Is there specific need for paasing a object instead of params?
The naming of the method can be improved.
Check the suggestion.

```suggestion
function elementIsInView(container: HTMLElement, element: HTMLElement) {
```

projects/angular/src/wizard/wizard-stepnav-item.ts
```diff
@@ -158,4 +170,30 @@ export class ClrWizardStepnavItem {
...
+  const elementTop = element.offsetTop;
+  const elementBottom = elementTop + element.clientHeight;
+
+  const containerTop = container.scrollTop;
+  const containerBottom = containerTop + container.clientHeight;
```
valentin-mladenov, line 196: Some `null` checks might be needed.

```suggestion
  const elementTop = element?.offsetTop;
  const elementBottom = elementTop + element?.clientHeight;

  const containerTop = container?.scrollTop;
  const containerBottom = containerTop + container?.clientHeight;
```

valentin-mladenov (approved):
LGTM

Check my comments and resolve them.

projects/angular/src/wizard/wizard-stepnav-item.ts
kevinbuhmann: >Is there specific need for paasing a object instead of params?

Yes. It prevents transposing the parameters.

PR #1851, opened 2025-05-28
fix(wizard): scroll current step into view (take two)

kevinbuhmann:
CDE-2764


projects/demo/src/app/wizard/wizard-basic.demo.html
```diff
@@ -199,11 +203,13 @@
...
-  <clr-wizard-page>
-    <ng-template clrPageTitle>Title for page 3</ng-template>
-    <!-- mandatory -->
-    <ng-template clrPageNavTitle>Step 3</ng-template>
-    <!-- optional -->
-    <p class="content-for-page-3">Content for step 3</p>
-  </clr-wizard-page>
+  <ng-container *ngIf="scrollingSteps">
+    <clr-wizard-page
+      *ngFor="let number of [3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30]"
```
dtsanevmw, line 208: Can you create a variable or a getter that can simply return 3 or 30 steps depending on scrollingSteps?
```suggestion
      *ngFor="let page of pages"
```

projects/angular/src/wizard/wizard-stepnav-item.ts
```diff
@@ -158,4 +176,22 @@ export class ClrWizardStepnavItem {
...
   }
+
+  private ensureCurrentStepIsScrolledIntoView() {
+    // Don't use "smooth" scrolling when the wizard is first opened to avoid a delay in scrolling the current step into view.
+    // The current step when the wizard is opened might not be the first step. For example, the wizard can be closed and re-opened.
+    let scrollBehavior: ScrollBehavior = 'auto';
+
+    return this.navService.currentPageChanged.pipe(
+      startWith(this.navService.currentPage),
+      tap(currentPage => {
```
dtsanevmw, line 187: Could we use skip instead of the variable "skipNextScroll" to skip the first emitted change? Or we need it specifically after the click?
```suggestion
      skip(1), 
      startWith(this.navService.currentPage),
      tap(currentPage => {
```

dtsanevmw (approved):
I left two suggestions but otherwise it looks good.

projects/demo/src/app/wizard/wizard-basic.demo.html
kevinbuhmann: Well, it would always return the same array because this is only rendered when "scrolling steps" is enabled. The first two pages are static with different content.

dtsanevmw: Yeah, but now we have 2 of the same pages that have a lot of paragraphs that are the same and these here so we can just use on NgFor for all pages instead of having static and dynamic ones and just change the number when needed.  It will save a lot of space and unnecessary code on that page.

kevinbuhmann: Okay, I refactored this.

PR #1852, opened 2025-05-29
feat(timeline): use density tokens

dtsanevmw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-2757

- [ ] Yes
- [X] No


projects/angular/src/timeline/_timeline.clarity.scss
```diff
@@ -23,6 +24,7 @@
     width: 100%;
     min-width: mixins.baselinePx(175);
     margin-left: tokens.$cds-global-space-5;
```
idonchev-vmw: This should not be "global-space" token. 
Actually, by design it does not exist at all, but I guess it might be a matter of a different (breaking) update.

projects/angular/src/timeline/_properties.timeline.scss
```diff
@@ -22,7 +23,8 @@
 
     // body (title)
     --clr-timeline-step-title-font-weight: #{tokens.$cds-alias-typography-font-weight-semibold};
+    --clr-timeline-icon-size: #{density.$clr-base-icon-size-3xl};
 
-    --clr-timeline-step-internal-spacing: #{tokens.$cds-global-space-6}; // internal property cannot be exposed
+    --clr-timeline-step-internal-spacing: #{density.$clr-base-vertical-offset-l}; // internal property cannot be exposed
```
idonchev-vmw, line 28: In Figma it's layout-space-s. 
Probably needs to be fixed in Figma.

projects/angular/src/timeline/_timeline.clarity.scss
```diff
@@ -103,8 +105,8 @@
   }
 
   .clr-timeline-step-title {
-    margin: timeline-variables.$clr-timeline-step-internal-spacing 0;
     color: timeline-variables.$clr-timeline-step-title-color;
+    margin-bottom: timeline-variables.$clr-timeline-step-internal-spacing;
```
idonchev-vmw: This is gap-xs by design. Way smaller.

projects/angular/src/timeline/_properties.timeline.scss
idonchev-vmw: nit: also, in vertical timeline, this should technically become horizontal-offset.

idonchev-vmw (commented):
Some inconsistencies with Figma. 
Most of them are either Figma errors, or design being ahead of implementation.
Marking them so they can be fixed in Figma or tracked/marked for future updates.

This should be gap-s by design.
image

projects/angular/src/timeline/_properties.timeline.scss
dtsanevmw: They are identical that's why even in designs on a lot of places is used similar approach to using just one token for the whole padding.  Although we can trigger discussion to remove these two and have only "clr-base-offest-*" instead of vertical and horizontal if there won't be any case where they are different.

projects/angular/src/timeline/_timeline.clarity.scss
dtsanevmw: Fixed in [e2cbcdc](https://github.com/vmware-clarity/ng-clarity/pull/1852/commits/e2cbcdcc26a652c987381c24457cc816f13da525)

projects/angular/src/timeline/_timeline.clarity.scss
dtsanevmw: Fixed in [e2cbcdc](https://github.com/vmware-clarity/ng-clarity/pull/1852/commits/e2cbcdcc26a652c987381c24457cc816f13da525). This was already handled by a gap so it's not needed anymore, it was creating additional space.

idonchev-vmw (commented):
Looks good, just one last thing.
There is a padding on .clr-timeline that I don't see in the specs.

projects/angular/src/timeline/_properties.timeline.scss
valentin-mladenov: I agree with @idonchev-vmw with:  `in vertical timeline, this should technically become horizontal-offset.`

Vertical and horizontal offsets not always have the same values.
Most of them have, but their values can change.

projects/angular/src/timeline/_variables.timeline.scss
```diff
@@ -6,6 +6,7 @@
  */
 
 @use '@cds/core/tokens/tokens.scss';
+@use '../layout/density/' as density;
```
LynxDelli: Do we need to add this dependency?

dtsanevmw: Fixed in [634b38d](https://github.com/vmware-clarity/ng-clarity/pull/1852/commits/634b38db9ce8ed3ae74accd9a146269517a05cdf)

idonchev-vmw (approved):

PR #1944, opened 2025-07-17
fix(datagrid): virtual scroll focus jumping

valentin-mladenov:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-2900

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/clarity.api.md
+++ b/projects/angular/clarity.api.md
@@ -1441,7 +1441,7 @@ export class ClrDatagridCell implements OnInit {
 export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, ClrDatagridFilterInterface<T>> implements OnDestroy, OnInit, OnChanges {
     // Warning: (ae-forgotten-export) The symbol "Sort" needs to be exported by the entry point index.d.ts
     // Warning: (ae-forgotten-export) The symbol "FiltersProvider" needs to be exported by the entry point index.d.ts
-    constructor(el: ElementRef<HTMLElement>, _sort: Sort<T>, filters: FiltersProvider<T>, vcr: ViewContainerRef, detailService: DetailService, changeDetectorRef: ChangeDetectorRef);
+    constructor(el: ElementRef<HTMLElement>, _sort: Sort<T>, filters: FiltersProvider<T>, vcr: ViewContainerRef, detailService: DetailService, changeDetectorRef: ChangeDetectorRef, commonStrings: ClrCommonStringsService);
     // (undocumented)
     get ariaSort(): "none" | "ascending" | "descending";
     // (undocumented)
@@ -1463,6 +1463,8 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, ClrDa
     filterValueChange: EventEmitter<any>;
     // (undocumented)
     get isHidden(): boolean;
+    // (undocumented)
+    ngAfterViewInit(): void;
     // (undocumented)
     ngOnChanges(changes: SimpleChanges): void;
     // (undocumented)
@@ -1486,6 +1488,8 @@ export class ClrDatagridColumn<T = any> extends DatagridFilterRegistrar<T, ClrDa
     set sortOrder(value: ClrDatagridSortOrder);
     // (undocumented)
     sortOrderChange: EventEmitter<ClrDatagridSortOrder>;
+    // (undocumented)
+    titleContainer: ElementRef<HTMLElement>;
     // (undocumented)
     set updateFilterValue(newValue: string | [number, number]);
     // (undocumented)
@@ -1638,7 +1642,7 @@ export class ClrDatagridDetailHeader implements AfterViewInit {
 // Warning: (ae-forgotten-export) The symbol "CustomFilter" needs to be exported by the entry point index.d.ts
 //
 // @public
-export class ClrDatagridFilter<T = any> extends DatagridFilterRegistrar<T, ClrDatagridFilterInterface<T>> implements CustomFilter, OnChanges, OnDestroy {
+export class ClrDatagridFilter<T = any> extends DatagridFilterRegistrar<T, ClrDatagridFilterInterface<T>> implements CustomFilter, OnDestroy {
     constructor(_filters: FiltersProvider<T>, commonStrings: ClrCommonStringsService, smartToggleService: ClrPopoverToggleService, platformId: any, elementRef: ElementRef<HTMLElement>, keyNavigation: KeyNavigationGridController);
     get active(): boolean;
     // (undocumented)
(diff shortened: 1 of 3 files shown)


projects/angular/src/data/datagrid/datagrid-virtual-scroll.directive.ts
```diff
@@ -360,8 +363,11 @@ export class ClrDatagridVirtualScrollDirective<T> implements AfterViewInit, DoCh
       const datagridRowElement = rootElements.find(rowElement => rowElement.tagName === 'CLR-DG-ROW');
       const rowRoleElement = datagridRowElement?.querySelector('[role="row"]');
 
-      // aria-rowindex should start with one, not zero, so we have to add one to the zero-based index
-      rowRoleElement?.setAttribute('aria-rowindex', (viewRef.context.index + 1).toString());
+      const newAriaRowIndex = (viewRef.context.index + 1).toString();
+      if (rowRoleElement?.getAttribute('aria-rowindex') !== newAriaRowIndex) {
+        // aria-rowindex should start with one, not zero, so we have to add one to the zero-based index
+        rowRoleElement?.setAttribute('aria-rowindex', newAriaRowIndex);
+      }
```
dtsanevmw, line 367: We can optimise it in some other time to trigger `updateAriaRowIndexes` function only if such update is needed. For example once new render range is triggered we can mark that update should be done.

projects/angular/src/data/datagrid/datagrid.ts
```diff
@@ -280,6 +280,28 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
           this.items.all = all;
         }
       }),
+      this.keyNavigation.nextCellCoordsEmitter.subscribe(cellCoords => {
+        if (!this.virtualScroll) {
+          return;
+        }
```
dtsanevmw: Move it outside and push it in to the subscriptions only `if (this.virtualScroll)` so we don't need to have unnecessary subscription in default datagrids too.

projects/angular/src/data/datagrid/datagrid-virtual-scroll.directive.ts
```diff
@@ -75,6 +76,8 @@ export class ClrDatagridVirtualScrollDirective<T> implements AfterViewInit, DoCh
   @Output() renderedRangeChange = new EventEmitter<ListRange>();
   @Input('clrVirtualPersistItems') persistItems = true;
 
+  keyNavItemCoords: CellCoordinates;
```
dtsanevmw: Can we save activeCellCoordinates inside key-navigation controller as this is the place to have these coordinates instead of controlling / saving them in other parts.

projects/angular/src/data/datagrid/utils/key-navigation-grid.controller.ts
```diff
@@ -46,26 +46,28 @@ export interface KeyNavigationGridConfig {
 export interface CellCoordinates {
   x: number;
   y: number;
+  ariaRowIndex?: string;
```
dtsanevmw, line 49: Can we set it to number and just do the transform in the setter only instead of the calculations when we are using these coordinateS?
```suggestion
  ariaRowIndex?: number;
```

dtsanevmw: Also we should use it as rowIndex instead as that's what is here. It's not aria attribute so we don't need this prefix to confuse us in the future.
```suggestion
  rowIndex?: string;
```

projects/angular/src/data/datagrid/datagrid.ts
```diff
@@ -280,6 +280,28 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
...
+
+        if (cellCoords.ariaRowIndex === this.virtualScroll?.keyNavItemCoords?.ariaRowIndex) {
+          this.virtualScroll.keyNavItemCoords = cellCoords;
+          return;
+        }
+
+        // aria-rowindex is always + 1. Check virtual scroller updateAriaRowIndexes method.
+        const rowIndex = Number(cellCoords.ariaRowIndex) - 1;
```
dtsanevmw: If you apply my other suggestions in the interface.
```suggestion
        const rowIndex = cellCoords.rowIndex - 1;
```

projects/angular/src/data/datagrid/datagrid.ts
```diff
@@ -295,23 +317,47 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
...
-          const active = this.keyNavigation.getActiveCell();
-          if (active) {
-            this.zone.runOutsideAngular(() => {
-              setTimeout(() => this.keyNavigation.setActiveCell(active));
-            });
-          }
+        if (this.virtualScroll && this.virtualScroll.keyNavItemCoords) {
+          this.zone.runOutsideAngular(() => {
+            this.focusCoords(this.virtualScroll.keyNavItemCoords);
+          });
```
dtsanevmw: There are duplicated checks and stuff for this function and it's only used here so we can shorter it a little and just use it here.
```suggestion
          this.zone.runOutsideAngular(() => {          
                const row = Array.from(this.rows).find(row => {
                  return row.el.nativeElement.children[0].ariaRowIndex === coords.ariaRowIndex;
                });
          
                if (!row) {
                  return;
                }
          
                const activeCell = row.el.nativeElement.querySelectorAll(this.keyNavigation.config.keyGridCells)[
                  coords.x
                ] as HTMLElement;
          
                this.keyNavigation.setActiveCell(activeCell);
                this.keyNavigation.focusElement(activeCell, { preventScroll: true });
          });
```

projects/angular/src/data/datagrid/utils/key-navigation-utils.ts
```diff
@@ -46,6 +46,16 @@ export class KeyNavigationUtils {
...
   }
 
+  setAriaRowIndexTo(cellCoords: CellCoordinates) {
+    let ariaRowIndex = this.rows[cellCoords.y].getAttribute('aria-rowindex');
+
+    if (!ariaRowIndex) {
+      ariaRowIndex = this.rows[cellCoords.y - 1].getAttribute('aria-rowindex');
+    }
+
+    cellCoords.ariaRowIndex = ariaRowIndex;
```
dtsanevmw, line 56: If you apply the other suggestions.
```suggestion
    cellCoords.rowIndex = Number(ariaRowIndex);
```

projects/angular/src/data/datagrid/datagrid-virtual-scroll.directive.ts
valentin-mladenov: The Idea behind is that KeyNavController to emit the coordinates and to be as stateless as possible.

Whomever is lsitning will act and/or save upon the change.

projects/angular/src/data/datagrid/datagrid-virtual-scroll.directive.ts
valentin-mladenov: I have created JIRA for it: CDE-2909

dtsanevmw (approved):

PR #1957, opened 2025-07-31
fix(modal): Update modal header CSS, remove unused code.

LynxDelli:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-2849

- [ ] Yes
- [X] No

### DIFF
--- a/projects/angular/src/modal/_modal.clarity.scss
+++ b/projects/angular/src/modal/_modal.clarity.scss
@@ -192,20 +192,16 @@
       align-items: flex-start;
     }
 
-    .modal-title,
-    .side-panel-title {
-      color: modal-variables.$clr-modal-title-color;
-      @include mixins.generate-typography-token('SECTION-20-STD');
-      margin: 0;
-    }
-
+    // this is with higher specificity than .modal-title, .side-panel-title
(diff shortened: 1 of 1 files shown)


projects/angular/src/modal/_modal.clarity.scss
```diff
@@ -192,20 +192,16 @@
...
     }
 
-    .modal-title,
-    .side-panel-title {
-      color: modal-variables.$clr-modal-title-color;
-      @include mixins.generate-typography-token('SECTION-20-STD');
-      margin: 0;
-    }
-
+    // this is with higher specificity than .modal-title, .side-panel-title
```
dtsanevmw: This is not needed as these two selectors got mixed together and wrong. The idea here is to have .modal-title to set only color, margin: 0 and padding: 0 so if someone set a different typography with cds-text to receive the font styles and keep the colors and spacings. However if they don't have `cds-text` then we should put font styles on top of the base styles that would come from the selector without `:not([cds-text])`.

dtsanevmw (approved):

PR #1963, opened 2025-08-06
 fix(datagrid): calculate average height for pageUp and pageDown navigation

valentin-mladenov:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/data/datagrid/utils/key-navigation-utils.ts
+++ b/projects/angular/src/data/datagrid/utils/key-navigation-utils.ts
@@ -43,8 +43,8 @@ export class KeyNavigationUtils {
   }
 
   get averageRowHeight() {
-    const heightSum = Array.from(this.rows.values()).reduce((initial, row) => {
-      return initial + row.clientHeight;
+    const heightSum = Array.from(this.rows.values()).reduce((sum, row) => {
+      return sum + row.clientHeight;
     }, 0);
 
     return Math.round(heightSum / this.rows.length);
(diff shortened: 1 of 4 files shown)


projects/angular/src/data/datagrid/utils/key-navigation-utils.ts
```diff
@@ -42,8 +42,16 @@ export class KeyNavigationUtils {
     return coordinates;
   }
 
+  get averageRowHeight() {
+    const heightSum = Array.from(this.rows.values()).reduce((initial, row) => {
+      return initial + row.clientHeight;
+    }, 0);
+
+    return Math.round(heightSum / this.rows.length);
+  }
```
dtsanevmw: We can use `requestAnimationFrame` to optimise it even better so we can get the clientHeights in a batch so we don't have to trigger reflows for each clientHeight.
```suggestion
  get averageRowHeight() {
      const rowHeights = [];
      requestAnimationFrame(() => {
          for (const row of this.rows.values()) {
            rowHeights.push(row.clientHeight);
          }
      });
      const sum = rowHeights.reduce((total, height) => total + height, 0);
      return Math.round(sum / rowHeights.length);
  }
```

dtsanevmw:
```suggestion
  get averageRowHeight() {
    const heightsSum = Array.from(this.rows.values()).reduce((sum, row) => {
      return sum + row.clientHeight;
    }, 0);

    return Math.round(sum / this.rows.length);
  }
```

dtsanevmw (approved):

PR #2003, opened 2025-09-10
fix(forms): propagate only clr classes to control container

valentin-mladenov:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-2530

- [X] Yes
- [ ] No

BREAKING CHANGE: only `clr-*` classes are going to be propagated to the control container. Other custom classes no longer are going to be propagated on the container.

### DIFF
--- a/projects/angular/src/data/datagrid/datagrid-row.spec.ts
+++ b/projects/angular/src/data/datagrid/datagrid-row.spec.ts
@@ -695,7 +695,7 @@ class ExpandTest {
 
 @Component({
   template: `
-    <clr-datagrid [clrDgSelected]="[]" [clrDgItemsTrackBy]="trackBy">
+    <clr-datagrid [clrDgSelected]="[]" [clrDgItemsIdentityFn]="trackBy">
       @for (item of items; track item.id) {
         <clr-dg-row [clrDgItem]="item" [clrDgSelectable]="clrDgSelectable"></clr-dg-row>
       }


projects/angular/src/forms/common/wrapped-control.ts
```diff
@@ -173,6 +172,15 @@ export class WrappedFormControl<W> implements OnInit, DoCheck, OnDestroy {
...
   }
 
+  private injectControlClassService(injector: Injector) {
+    if (!this.controlClassService) {
+      this.controlClassService = injector.get(ControlClassService, null);
+      if (this.controlClassService) {
+        this.controlClassService.initControlClass(this.renderer, this.el.nativeElement);
+      }
+    }
+  }
```
dtsanevmw, line 181: Do we need this function to be ran both in constructor and ngOnInit? Separate the injecting from doing `initControlClass` and leave the injection in the constructor and the `initControlClass` in ngOnInit.

projects/angular/src/forms/common/providers/control-class.service.ts
```diff
@@ -40,13 +40,20 @@ export class ControlClassService {
...
+        if (cssClass.startsWith('clr-')) {
+          controlCssClasses.push(cssClass);
+        }
+
+        if (cssClass.startsWith('clr-col')) {
+          renderer.removeClass(element, cssClass);
         }
       });
+
+      this.className = controlCssClasses.join(' ').trim();
```
dtsanevmw:
```suggestion
      const klasses = element.className.split(' ');
      this.className = klasses.filter((klass) => klass.startsWith('clr-')).join(' ');
      klasses.forEach(klass => {
        if (klass.startsWith('clr-col')) {
          renderer.removeClass(element, klass);
        }
      });
```

projects/angular/src/forms/common/wrapped-control.ts
valentin-mladenov: With current implementation, yes we do.

We have 2 cases. 
1. Single Input HTML element with `clrInput`. In this case the input don't have container initially and cannot inject `ControlClassService` in constructor.
2. Input HTML element with `clrInput` is put in `clr-input-container` initially. In this case the input have container and the `ControlClassService` can be injected in constructor.

projects/angular/src/forms/common/providers/control-class.service.ts
valentin-mladenov: I did it in 1 loop to minimize the iterations over `klasses/cssClasses` array.

valentin-mladenov: Corrected in commit 1ce13f3

projects/angular/src/forms/common/wrapped-control.ts
valentin-mladenov: Corrected in commit 1ce13f3

PR #2014, opened 2025-09-19
fix(file-input): validate accept entries using suffix matching

dtsanevmw:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently the file input validators have a regex that take all dots from the file name and taking it as the extension, for example: "file-name.0.0.1.zip" would return extension to be ".0.01.zip" while we have `accept=".zip"` and that would fail.

Issue Number: CDE-2971

## What is the new behavior?

Use suffix matching by each entry in accept so we can compare `.zip` to the file name and dynamically take only the last part of the file name.

- [ ] Yes
- [X] No

### DIFF
--- a/projects/angular/src/forms/file-input/file-input-validator.spec.ts
+++ b/projects/angular/src/forms/file-input/file-input-validator.spec.ts
@@ -170,6 +170,135 @@ function fileInputValidatorSpec(testComponent: Type<TestComponent>) {
 
     expect(getErrorMessages(nativeElement)).toBe('File size too large');
   });
+
+  it('accept: ".gz" should accept both readme.gz and archive.tar.gz (suffix match)', () => {
+    fixture.componentInstance.accept = '.gz';
+    fixture.detectChanges();
+
+    selectFiles(fileInputElement, [new File(['+'.repeat(100)], 'readme.gz')]);
+    fixture.detectChanges();
+    expect(nativeElement.querySelector('clr-control-error')).toBeNull();
+
+    selectFiles(fileInputElement, [new File(['+'.repeat(100)], 'archive.tar.gz')]);
+    fixture.detectChanges();
+    expect(nativeElement.querySelector('clr-control-error')).toBeNull();
+  });
+
+  it('accept: ".tar.gz" should accept archive.tar.gz and reject readme.gz', () => {
+    fixture.componentInstance.accept = '.tar.gz';
+    fixture.detectChanges();
+
+    selectFiles(fileInputElement, [new File(['+'.repeat(100)], 'archive.tar.gz')]);
+    fixture.detectChanges();
+    expect(nativeElement.querySelector('clr-control-error')).toBeNull();
+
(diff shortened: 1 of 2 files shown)


projects/angular/src/forms/file-input/file-input-validator.ts
```diff
@@ -64,4 +67,45 @@ export class ClrFileInputValidator implements Validator {
...
+        if (entry.endsWith('/*')) {
+          const prefix = entry.slice(0, entry.length - 1); // keep trailing slash
+          if (type.startsWith(prefix)) {
+            return true;
+          }
+        } else if (type === entry) {
+          return true;
+        }
+        continue;
+      }
```
valentin-mladenov, line 103: Consider the following.

```suggestion
      if (entry.startsWith('.')) {
        const depth = (entry.match(/\./g) || []).length;
        if (this.getSuffixByDepth(name, depth) === entry) {
          return true;
        }
        continue;
      }

      // MIME check 
      if (entry.endsWith('/*')) {
        const prefix = entry.slice(0, entry.length - 1); // keep trailing slash
        if (type.startsWith(prefix)) {
          return true;
        }
      } else if (entry.includes('/') && type === entry) {
        return true;
      }
```

projects/angular/src/forms/file-input/file-input-validator.spec.ts
```diff
@@ -170,6 +170,135 @@ function fileInputValidatorSpec(testComponent: Type<TestComponent>) {
...
+    fixture.detectChanges();
+    expect(nativeElement.querySelector('clr-control-error')).toBeNull();
+  });
+
+  it('empty MIME type falls back to extension checks only', () => {
+    fixture.componentInstance.accept = '.gz';
+    fixture.detectChanges();
+
+    // No MIME provided, should be accepted by extension
+    selectFiles(fileInputElement, [new File(['+'.repeat(100)], 'readme.gz', { type: '' })]);
```
valentin-mladenov, line 302: Add some more tests for testing in file name extensions. 

Exmaple: `readme.gz.tar` with accept `.gz` should fail.

FYI logic is working as expected just the test is missing for such a edge case.

valentin-mladenov (commented):
Check my comments

projects/angular/src/forms/file-input/file-input-validator.ts
dtsanevmw: Fixed in [1c4489b](https://github.com/vmware-clarity/ng-clarity/pull/2014/commits/1c4489b5b6594e7565c18eb57494d2493814a84f)

projects/angular/src/forms/file-input/file-input-validator.spec.ts
dtsanevmw: Fixed in [1c4489b](https://github.com/vmware-clarity/ng-clarity/pull/2014/commits/1c4489b5b6594e7565c18eb57494d2493814a84f)

PR #2102, opened 2025-11-12
fix(datagrid): virtual scroll missing last row bottom border

valentin-mladenov:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Last row in virtual scroll have missing bottom border.

Issue Number: CDE-2989

## What is the new behavior?
Last row in virtual scroll have bottom border.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/src/data/datagrid/datagrid.ts
+++ b/projects/angular/src/data/datagrid/datagrid.ts
@@ -553,7 +553,7 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
 
     const scrollClass = 'datagrid-scrollbar-visible';
 
-    if (this.cachedRowsHeight - this.cachedContentHeight >= 1) {
+    if (this.cachedRowsHeight > this.cachedContentHeight) {
       this.renderer.addClass(rowsWrapper, scrollClass);
     } else {
       this.renderer.removeClass(rowsWrapper, scrollClass);
--- a/projects/angular/src/data/datagrid/_datagrid.clarity.scss
+++ b/projects/angular/src/data/datagrid/_datagrid.clarity.scss
@@ -153,7 +153,7 @@
         min-height: 100%;
 
         &.datagrid-scrollbar-visible .datagrid-placeholder-container {
-          border-top: none;
+          border-top-color: tokens.$cds-alias-object-opacity-0;
         }
       }
     }


projects/angular/src/data/datagrid/datagrid.ts
```diff
@@ -432,6 +434,16 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
...
+
+    if (
+      this.virtualScroll &&
+      this.contentWrapper.nativeElement &&
+      this.isScrollbarVisible !==
+        this.contentWrapper.nativeElement.scrollHeight > this.contentWrapper.nativeElement.clientHeight
+    ) {
+      this.isScrollbarVisible =
+        this.contentWrapper.nativeElement.scrollHeight > this.contentWrapper.nativeElement.clientHeight;
+    }
```
dtsanevmw: This content should be in main-renderer ViewChecked as we need to track view changes

dtsanevmw: Not sure why we need to check the same thing in the `if` condition and then assign the value with the same. Just assigning it when we have virtual scroll should be enough to put the class.
```suggestion

    if (
      this.virtualScroll &&
      this.contentWrapper.nativeElement
    ) {
      this.isScrollbarVisible =
        this.contentWrapper.nativeElement.scrollHeight > this.contentWrapper.nativeElement.clientHeight;
    }
```

projects/angular/src/data/datagrid/datagrid.html
```diff
@@ -90,7 +90,7 @@
             @if (virtualScroll) {
             <div class="datagrid-content-virtual-spacer" [style.height]="virtualScroll?.totalContentHeight"></div>
             }
-            <div role="presentation" class="datagrid-rows">
+            <div role="presentation" class="datagrid-rows" [class.datagrid-no-visible-scroller]="!isScrollbarVisible">
```
dtsanevmw: Sounds better to set it if we have visible scroll or not, it's shorter and easier to understand
```suggestion
            <div role="presentation" class="datagrid-rows" [class.datagrid-visible-scroller]="isScrollbarVisible">
```

projects/angular/src/data/datagrid/datagrid.ts
valentin-mladenov: Moving to `main-renderer ViewChecked` actually forces another cycle, because it's after view checked.
While testing the change a ` ClrDatagrid virtual scroller` `Moves focus on PageDown and PageUp` test is failing with.

```
Error: NG0100: ExpressionChangedAfterItHasBeenCheckedError: 
  Expression has changed after it was checked. 
  Previous value: 'false'. Current value: 'true'. Expression location: ClrDatagrid2 component. ....
```

projects/angular/src/data/datagrid/datagrid.html
valentin-mladenov: Class name changed in commit e6ba8ac.

projects/angular/src/data/datagrid/datagrid.ts
valentin-mladenov: Corrected in commit e6ba8ac.

projects/angular/src/data/datagrid/datagrid.ts
```diff
@@ -523,4 +537,21 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
       this._virtualScrollSubscriptions = [];
     }
   }
+
+  private handleResizeChanges(entries: ResizeObserverEntry[]) {
+    if (!this.virtualScroll) {
+      return;
+    }
```
dtsanevmw:
```suggestion
    // @TODO use for all variants of datagrid
    if (!this.virtualScroll) {
      return;
    }
```

projects/angular/src/data/datagrid/datagrid.ts
```diff
@@ -335,6 +346,9 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
   ngAfterViewInit() {
     this.keyNavigation.initializeKeyGrid(this.el.nativeElement);
 
+    this.resizeObserver.observe(this.contentWrapper.nativeElement);
+    this.resizeObserver.observe(this.rowsWrapper.nativeElement);
```
dtsanevmw: Add `this.resizeObserver.disconnect()` in ngOnDestroy and also we can try to check if there is a virtualScroll and not observing at all. For example we have `    this._virtualScroll.changes.subscribe(() => {
        this.toggleVirtualScrollSubscriptions();
      }),` where we track when the virtual scroll appear and enable subscriptions, we can skip observing on ngAfterViewInit if virtual scroll is not available and start observing if virtual scroll appear dynamically ( like in the demo app with empty rows toggle ).

valentin-mladenov: Corrected with commit d79aa3e.

projects/angular/src/data/datagrid/datagrid.ts
valentin-mladenov: Corrected with commit d79aa3e.

PR #2163, opened 2026-01-20
feat(badge, label) add outline and solid variants

valentin-mladenov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Labels are always outlined.
Badges are always solid.

Issue Number: CDE-2973

## What is the new behavior?
Labels heve solid variant
Badges have outlined vairant.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/clarity.api.md
+++ b/projects/angular/clarity.api.md
@@ -894,10 +894,6 @@ export class ClrAccordionPanel implements OnInit, OnChanges {
     // (undocumented)
     getPanelStateClasses(panel: AccordionPanelModel): string;
     // (undocumented)
-    headingEnabled: boolean;
-    // (undocumented)
-    get headingLevel(): HeadingLevel;
-    // (undocumented)
     get id(): string;
     set id(value: string);
     // (undocumented)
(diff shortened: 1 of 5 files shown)


projects/angular/src/emphasis/badge/badge.ts
```diff
@@ -25,11 +25,13 @@ export enum ClrBadgeColors {
   template: `<ng-content></ng-content>`,
   host: {
     class: 'badge',
+    '[class.badge-outlined]': '!solid',
     '[class]': 'colorClass',
   },
 })
 export class ClrBadge {
   @Input('clrColor') color: ClrBadgeColors | string = ClrBadgeColors.None;
+  @Input('clrSolid') solid = true;
```
dtsanevmw: If the new feature is the `outline` type, then we should have outlined as property and a transformer to be able to put static attribute:
```suggestion
  @Input({ alias: 'clrOutlined', transform: booleanAttribute }) outlined = true;
```

projects/angular/src/emphasis/label/label.ts
```diff
@@ -47,6 +48,7 @@ export class ClrLabel {
   @Input('clrText') textContent = '';
   @Input('clrClickable') clickable = false;
   @Input('clrDisabled') disabled = false;
+  @Input('clrSolid') solid = false;
```
dtsanevmw: Support static attributes
```suggestion
  @Input({ alias: 'clrSolid', transform: booleanAttribute}) solid = false;
```

projects/angular/src/emphasis/label/label.ts
```diff
@@ -37,6 +37,7 @@ export enum ClrLabelColors {
     class: 'label',
     '[class.clickable]': 'clickable',
     '[class.disabled]': 'disabled',
+    '[class.label-solid]': 'solid',
```
dtsanevmw: Maybe it will be better to be consistent with our other components like buttons where we have `btn-$color-$type`, like `btn-info-outline`. So we are going to have `label label-info-solid` or `badge badge-info-outline`

valentin-mladenov: Coupling the `color` with `type` wouldn't lead to simplification.

projects/angular/src/emphasis/label/label.ts
valentin-mladenov: `clrSolid` is repalced with `clrType` in commit 4842ca7

projects/angular/src/emphasis/badge/badge.ts
valentin-mladenov: `clrSolid` is repalced with `clrType` in commit 4842ca7

PR #2168, opened 2026-01-26
feat(combobox): add item identity function

idonchev-vmw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-1026

## What is the new behavior?
Item identity function added. Now combobox can update selected items if underlying data gets changed.

- [ ] Yes
- [x] No


projects/angular/src/forms/combobox/providers/option-selection.service.ts
```diff
@@ -115,4 +120,33 @@ export class OptionSelectionService<T> {
...
+        }
+      });
+    } else {
+      const selectedItem = this.selectionModel.model as T;
+      for (let index = 0; index < this.rawItems.length; index++) {
+        const item = this.rawItems[index];
+        if (this.identifyBy(item) === this.identifyBy(selectedItem)) {
+          // We need this call to be able to refresh the selection in single selection model.
+          // This prevents us from moving the current method in the respective models.
+          this.setSelectionValue(item);
```
dtsanevmw: This would trigger `selectionChanged` event to users both on initial load and also each time items change and even if there is not really a change to the item that is selected or other items, it could be 1 new item added or a new reference.

projects/angular/src/forms/combobox/option-items.directive.ts
```diff
@@ -83,28 +88,32 @@ export class ClrOptionItems<T> implements DoCheck, OnDestroy {
...
-    if (!this._rawItems || this.filter === undefined || this.filter === null) {
+    if (!this.optionService.rawItems || this.filter === undefined || this.filter === null) {
       return;
     }
 
     const normalizedFilterValue = normalizeValue(this.filter);
 
     if (this.optionService.showAllOptions) {
-      this.filteredItems = this._rawItems;
+      this.filteredItems = this.optionService.rawItems;
```
dtsanevmw: We can continue using `this.rawItems` as the getter points to `this.optionService.rawItems`
```suggestion
      this.filteredItems = this.rawItems;
```

idonchev-vmw: Done, other references too.

projects/angular/src/forms/combobox/providers/option-selection.service.ts
idonchev-vmw: Added check for the change, now we trigger selection change only if the display name used for the single selection is changed. Performing deep comparison is performance heavy and will not have any practical value.

projects/angular/forms/combobox/option-items.directive.ts
```diff
@@ -83,28 +88,32 @@ export class ClrOptionItems<T> implements DoCheck, OnDestroy {
         this.iterableProxy.ngDoCheck();
       }
     }
+    const changes = this.modelDiffer.diff(this.rawItems);
+    if (changes) {
+      this.optionService.updateSelection();
```
dtsanevmw, line 93: If we are going with this approach then any selection triggered manually to the control and is not by reference then the item will not be selected. The model will be changed but as it checks by `this.model === item` or `this.model.includes(item)` then the item will not be selected in the list of items. I think this is the con of trying to sync the selection and options all the time and expect everything is the same.

idonchev-vmw: You are right, but doing deep comparisons out of the box is quite expensive. There should be some ballance between utility and performance. The user can always be providing immutable chunks of data if updating on the field level.
This function in the combobox demo is exactly aiming to achieve the above, and we may want to add more visibility to the immutable approach on our website.
```
  private updateName(item: any, newName: string) {
    return { ...item, name: newName };
  }
  ```

idonchev-vmw:
Closing in favor of https://github.com/vmware-clarity/ng-clarity/pull/2211 where we only handle the identification, without synchronisation. The latter will be added if user requests arrive.

PR #2180, opened 2026-01-30
fix(header): remove obsolete header variants

idonchev-vmw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: Remove obsolete styles

Issue Number: CDE-1475

- [x] Yes
- [ ] No

BREAKING CHANGES:
The header styles from header-4 to header-7 have been removed, as they are obsolete and currently are pointing to some of styles between header-1 and header-3. Also header-8 has been removed as it was a hard-coded exception for the dark theme used by specific team in the past.

The following replacements should be done if any of the removed styles is used:
- header-4 -> header-3
- header-5 -> header-3
- header-6 -> header-1
- header-7 -> header-2
- header-8 -> use your own style override, or choose one of the supported styles above


projects/demo/src/app/nav/header-types-old.demo.html
dtsanevmw, line 1: Do we need this whole file if it's just for old headers? `header-types-old`?

projects/demo/src/app/app.component.html
```diff
@@ -6,7 +6,7 @@
 -->
 
 <clr-main-container>
-  <clr-header class="header-6">
+  <clr-header class="header-1">
```
dtsanevmw: If `header` only is equal to header-1, could we just use that everywhere where we have single class `header-1`?
```suggestion
  <clr-header class="header">
```

.storybook/stories/header/header-static.stories.ts
```diff
@@ -31,7 +31,7 @@ export default {
 
 const HeaderStaticTemplate: StoryFn = args => ({
   template: `
-    <header class="header-6">
+    <header class="header-1">
```
dtsanevmw: If header only is equal to header-1, could we just use that everywhere where we have single class header-1?
```suggestion
    <header class="header">
```

projects/demo/src/app/nav/header-types-old.demo.html
idonchev-vmw: Absolutely. Removing. Good catch

.storybook/stories/header/header-static.stories.ts
idonchev-vmw: If we want the original behavior, we should have had it as "header header-6", which can then get changed to either "header header-1", or just "header", which will yield the same results.
I don't know why the original author skipped the "header" class. There is slight difference in spacing between just "header" and "header-1". 
Preferred to keep the original look, so I don't need to research for the original ideas behind this, or change it without understanding it. We have many places where we have the dual class "header header-X", no pain in keeping the shorthand at one place.

valentin-mladenov: `There is slight difference in spacing between just "header" and "header-1".`
Where did you observe that?

`header-X` classes (should) only change the background color and nothing else. On short `header-1` class is redundant since `header` have the same background color.
We can still keep `header-1` for backward compatibility.

idonchev-vmw: @valentin-mladenov Exactly. I don't know if your comment is addressed at me, or Daniel's comment, but this difference was the reason I left the header-1.
It does not matter too much though, I don't mind any approach. Decided to keep the header-1 for the sake of variety in our demo cases.

projects/demo/src/app/app.component.html
idonchev-vmw: Same as below.

.storybook/stories/header/header-static.stories.ts
valentin-mladenov: @idonchev-vmw the first part of the comment is addressed to you. You stated: 
`There is slight difference in spacing between just "header" and "header-1".` Where did you observe that? 
There should not be any other changed but the background color.

The second part is to all of us. 
Class `header` have no differences than `header-1`. We can 

1. Remove the `header-1` class. Casses will be `header`, `header-2` and `header-3`
2. Keep the class and have all 4, but `header` and `header-1`  are the same.

EDIT: Using `header` HTML tag removes the need of using `header` and `header-1` classes al thogether.

projects/angular/src/layout/main-container/STYLES.md
```diff
@@ -0,0 +1,58 @@
...
+| --clr-header-search-border-active          | Border of active/focused search field                  |
+| --clr-header-search-border-hover           | Border of hovered search field                         |
+| --clr-header-search-placeholder-text-color | Text color of the header search placeholder            |
+
+## CSS Classes
+
+| Class name               | Description                          |
+| ------------------------ | ------------------------------------ |
+| clr-header               | Application header element           |
+| clr-header > search      | Header search field                  |
```
valentin-mladenov: The class is just `header`
.
```suggestion
| header               | Application header element           |
| header > search      | Header search field                  |
```

valentin-mladenov: `header` class names are mentioned 2 times.

PR #2203, opened 2026-02-11
feat(datagrid): simplify selection logic

valentin-mladenov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The datagrid has two separate selection APIs:
- **Multi selection**: `[(clrDgSelected)]` binding with `T[] | undefined` — implicitly sets `SelectionType.Multi`
- **Single selection**: `[(clrDgSingleSelected)]` binding with `T` — implicitly sets `SelectionType.Single`

There is no explicit `clrDgSelectionType` input. The selection type is determined implicitly by which input binding is used. The `SelectionType` enum uses numeric values (`None = 0`, `Single = 1`, `Multi = 2`) and is not directly exposed to templates. The `Selection` provider maintains separate `current` (array) and `currentSingle` (single item) properties, and the `change` Observable emits `T | T[]` depending on the selection mode. The addons package maintains a separate duplicate `SelectionType` enum in `@clr/addons/datagrid/interfaces/selection-type.ts`.

Issue Number: CDE-3030

### Unified Selection API

Both single and multi selection are now controlled through a single input/output pair plus an explicit selection type input:

- **`clrDgSelectionType`** — new input that explicitly sets the selection mode (`'none'`, `'single'`, `'multi'`)
- **`[clrDgSelected]`** — accepts `T[]` for both single and multi selection modes
- **`(clrDgSelectedChange)`** — emits `T[]` for both modes

For single selection, wrap the item in an array: `[clrDgSelected]="[selectedItem]"`.

The `clrDgSelectionType` input uses an Angular `transform` function (`selectionTypeAttribute`) that validates string values at runtime, following Angular's `booleanAttribute` / `numberAttribute` convention. This enables clean attribute binding:

```html
<clr-datagrid clrDgSelectionType="multi">
```

### SelectionType enum changes

`SelectionType` values changed from implicit numeric (`0`, `1`, `2`) to explicit string literals (`'none'`, `'single'`, `'multi'`).

### Addons simplification

- Removed the duplicate `SelectionType` enum from `@clr/addons/datagrid` — consumers now import from `@clr/angular/data/datagrid`.
- Removed ~150 lines of manual selection wiring in the addons `DatagridComponent` by delegating to the Clarity datagrid via template bindings (`[clrDgSelected]`, `[clrDgSelectionType]`, `(clrDgSelectedChange)`, `[clrDgItemsIdentityFn]`).
- Removed `setSelectedItems()`, `selectGridItems()`, `subscribeToSelectionChange()`, `preserveSelectionWhenThereIsFilterDefine()` methods.

### Selection provider cleanup

- Removed `currentSingle` setter — all selection mutations go through `current` / `updateCurrent()`.
- `currentSingle` is now a read-only getter that returns `_current[0]`.
- `selection.change` Observable type narrowed from `Observable<T[] | T>` to `Observable<T[]>`.
- `selectable` is now a public getter (was private `_selectable`).
- `selected` input setter normalizes `null`/`undefined` to `[]` for consistent internal state.
- Removed the `_items.smart && !newSingle` guard in the Single selection items-update path — stale selections are no longer auto-cleared when items update.

### Demo page

Added a "Selection Sandbox" demo page (`/datagrid/selection-sandbox`) — an interactive playground for testing datagrid selection behavior across both `clr-datagrid` and `appfx-datagrid`. 

- [x] Yes
- [ ] No

### Migration Guide

| Old | New | Entry Point |
|-----|-----|-------------|
| `[(clrDgSingleSelected)]="item"` | `[(clrDgSelected)]="items" clrDgSelectionType="single"` (wrap item in array) | `@clr/angular/data/datagrid` |
| `(clrDgSingleSelectedChange)` | `(clrDgSelectedChange)` (emits `T[]`, use `$event[0]` for the single item) | `@clr/angular/data/datagrid` |
| Implicit selection type via input used | Explicit `clrDgSelectionType="none" \| "single" \| "multi"` | `@clr/angular/data/datagrid` |
| `SelectionType` (from `@clr/addons/datagrid`) | `SelectionType` (from `@clr/angular/data/datagrid`) | `@clr/angular/data/datagrid` |
| `SelectionType.None = 0` / `Single = 1` / `Multi = 2` | `SelectionType.None = 'none'` / `Single = 'single'` / `Multi = 'multi'` | `@clr/angular/data/datagrid` |
| `selection.currentSingle = item` | `selection.current = [item]` | Internal / `Selection` provider |
| `selection.change` → `Observable<T \| T[]>` | `selection.change` → `Observable<T[]>` | Internal / `Selection` provider |

### Example migration

```html
<!-- Before: Single selection (implicit type) -->
<clr-datagrid [(clrDgSingleSelected)]="selectedUser">

<!-- After: Single selection (explicit type) -->
<clr-datagrid [(clrDgSelected)]="selectedUsers" clrDgSelectionType="single">
```

```html
<!-- Before: Multi selection (implicit type) -->
<clr-datagrid [(clrDgSelected)]="selectedUsers">

<!-- After: Multi selection (explicit type) -->
<clr-datagrid [(clrDgSelected)]="selectedUsers" clrDgSelectionType="multi">
```

```typescript
// Before
selectedUser: User;

// After
selectedUsers: User[] = [];

// To access the single selected item:
get selectedUser(): User {
  return this.selectedUsers[0];
}
```

BREAKING CHANGES: The datagrid selection API has been unified.

- `clrDgSingleSelected` input and `clrDgSingleSelectedChange` output have been removed — use `clrDgSelected` with `clrDgSelectionType="single"` instead, wrapping the selected item in an array.
- A new `clrDgSelectionType` input is now required to explicitly set the selection mode (previously determined implicitly by which input was used).
- `SelectionType` enum values changed from numeric (0, 1, 2) to string literals ('none', 'single', 'multi').
- The addons `SelectionType` enum has been removed — import from `@clr/angular/data/datagrid`.
- `selection.change` Observable now always emits `T[]` (previously `T | T[]` for single selection).
- `selection.currentSingle` setter has been removed — use `selection.current = [item]` instead.

### DIFF
--- a/projects/angular/data/datagrid/providers/selection.ts
+++ b/projects/angular/data/datagrid/providers/selection.ts
@@ -177,7 +177,7 @@ export class Selection<T = any> {
 
     this._selectionType = value ?? SelectionType.None;
 
-    if (!this._current && value !== SelectionType.None) {
+    if (!this._current && this._selectionType !== SelectionType.None) {
       this._current = [];
     }
   }


projects/angular/data/datagrid/providers/selection.ts
```diff
@@ -187,11 +182,14 @@ export class Selection<T = any> {
     if (value === this.selectionType) {
       return;
     }
-    this._selectionType = value;
-    if ([SelectionType.None, SelectionType.Single].includes(value)) {
-      delete this.current;
-    } else {
-      this.updateCurrent([], false);
+
+    this._selectionType = value ? value : SelectionType.None;
```
dtsanevmw:
```suggestion
    this._selectionType = value ?? SelectionType.None;
```

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -728,6 +729,8 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
 
   ngAfterViewInit() {
     this.#areRowsDisabled = !!this.gridLayoutModel.disabled;
+
+    // identity func
```
dtsanevmw: Not needed?

projects/angular/data/datagrid/utils/key-navigation-grid.controller.spec.ts
```diff
@@ -14,7 +14,12 @@ import { Selection } from '../providers/selection';
 
 @Component({
   template: `
-    <clr-datagrid class="datagrid-compact" [style.height.px]="185" [(clrDgSelected)]="selected">
+    <clr-datagrid
+      class="datagrid-compact"
+      [style.height.px]="185"
+      [(clrDgSelected)]="selected"
+      [clrDgSelectionType]="selected ? 2 : 0"
```
idonchev-vmw: I guess we are mocking the old behavior with this, so we don't need to rewrite much code in the tests.
It will work, but I consider tests as one of the examples of how to use our components. I would not like to see a user mimicking this technique.

For the very same reason, I don't much like the "magic numbers" values of 1, 2, 3 instead of the specific enums.

projects/angular/data/datagrid/providers/selection.ts
```diff
@@ -187,11 +182,14 @@ export class Selection<T = any> {
...
+
+    // TODO: do we really need to remove the selection data
```
idonchev-vmw: Empty array should be fine. It causes unnecessary changes now, like the one in the datagrid-footer.ts

projects/angular/data/datagrid/datagrid.ts
```diff
@@ -207,30 +206,32 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
...
+    // in SINGLE SELECTION mode if the value is the same as currently selected do nothing
+    if (
+      this.selection.selectionType === SelectionType.Single &&
+      // if value is NULL compare it directly to current single if NOT take first index of the array
+      ((value && value[0] === this.selection.currentSingle) || value === this.selection.currentSingle)
+    ) {
+      return;
     }
+
+    // TODO: it make no sense!!! SINGLE emits changes, MULTI doesn't
```
idonchev-vmw: Yes, it makes no sense. Shouldn't we unify this? But it will only work when selection array is immutable. If anything changes inside the array - it skips detection.

projects/angular/data/datagrid/datagrid.ts
```diff
@@ -207,30 +206,32 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
...
-      this.selection.currentSingle = value;
-    } else if (this.selection.currentSingle) {
-      this.selection.currentSingle = null;
+  @Input('clrDgSelected')
+  set selected(value: T[]) {
+    // in SINGLE SELECTION mode if the value is the same as currently selected do nothing
+    if (
+      this.selection.selectionType === SelectionType.Single &&
+      // if value is NULL compare it directly to current single if NOT take first index of the array
+      ((value && value[0] === this.selection.currentSingle) || value === this.selection.currentSingle)
```
idonchev-vmw: I'd much like to prune the extra values, if more than one is supplied in Single selection mode.
We may even add the safeguard in the getter too.

projects/angular/data/datagrid/datagrid.ts
```diff
@@ -350,10 +351,11 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
       this.stickyHeaders.changes.subscribe(() => this.resize()),
       this.stateProvider.change.subscribe(state => this.refresh.emit(state)),
       this.selection.change.subscribe(s => {
-        if (this.selection.selectionType === SelectionType.Single) {
-          this.singleSelectedChanged.emit(s as T);
-        } else if (this.selection.selectionType === SelectionType.Multi) {
-          this.selectedChanged.emit(s as T[]);
+        if (
+          this.selection.selectionType === SelectionType.Single ||
```
idonchev-vmw: `selection.selectionType !== SelectionType.None` is one line shorter.

projects/addons/datagrid/datagrid.component.spec.ts
```diff
@@ -1019,38 +1023,43 @@ describe('DatagridComponent', () => {
...
       gridHelper.getRows()[2].click();
       this.fixture.detectChanges();
+      await delay();
 
       expect(gridHelper.getRows()[2].isSelected()).toBeTruthy();
       expect(gridHelper.getSelectedRows().length).toBe(1);
     });
 
-    it('let the user to select only one item', fakeAsync(function (this: any) {
+    it('let the user to select only one item', async function (this: any) {
```
idonchev-vmw: Why do we drop the fakeAsync + tick in favor of async + await + setTimeout?

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -63,11 +63,11 @@ import { ActionBarLayout, ActionDefinition } from './shared/action/action-defini
 import { ActionClickEvent, SingleRowActionOpen } from './shared/action/actions-event-types';
 import { ColumnDefinition } from './shared/column/column-definitions';
 
-const resources = {
-  selection: {
-    singleDefaultEntity: {},
-  },
-};
+// const resources = {
```
idonchev-vmw: Don't leave commented, please.

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -604,12 +604,13 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
...
   }
-
   set selectedItems(items: T[]) {
     this.#selectedItems = Array.isArray(items) ? items.filter((item: T) => item !== undefined) : [];
 
+    this.cdr.detectChanges();
+
     if (this.clrDatagrid && items) {
-      this.selectGridItems(this.#selectedItems, false);
+      // this.selectGridItems(this.#selectedItems, false);
```
idonchev-vmw: Don't leave as comment.

projects/addons/datagrid/datagrid.component.spec.ts
valentin-mladenov: fakeAsync + tick didn't work as expected.

projects/angular/data/datagrid/datagrid.ts
valentin-mladenov: In Single selection mode we care for the 0 indexed value. 
The user provided array should be seen as unmutable for us and therefore removing any data by some secret logic from it is forbidden.

idonchev-vmw: As a user I would expect a single value for single selection mode.
Even the fact that we're supplying array for single selection is a "trade off" solution.
The user may supply various types of invalid inputs. We should be able to defend our "trade off" solution from the side effects of the compromise we made with it.

projects/addons/datagrid/datagrid.component.spec.ts
idonchev-vmw: I see we have numerous detectChanges(false) calls here. (And in some more datagrid-related specs)
I don't know the reason it was originally done this way (it suppresses ExpressionChangedAfterItHasBeenCheckedError errors), but do we still need it with the async/await approach?

projects/angular/data/datagrid/datagrid.ts
valentin-mladenov: > The user may supply various types of invalid inputs. 

Correct! The error is theirs to make, not ours to fix.

> We should be able to defend our "trade off" solution from the side effects of the compromise we made with it.

There are no side effects. The single selection will always work with the 0 indexed item, others are not our consern.

In some edge cases when Selection is changed dynamically between multi and single, keping the selection array as is will not lead to lose of selection.

valentin-mladenov: Also regarding removing data from the pre-selectiuon array. 

Angular does not guarantee input evaluation order.

```html
[clrDgSelectionType]="clrDgSelectionType"
[clrDgSelected]="selectedItems"
```

If `selected` setter runs first (while `selectionType` is still `None`), the multi-selection logic will behave incorrectly.

projects/addons/datagrid/datagrid.component.ts
valentin-mladenov: Corrected in commit 75ce000

projects/addons/datagrid/datagrid.component.ts
valentin-mladenov: Corrected in commit 75ce000

projects/addons/datagrid/datagrid.component.spec.ts
valentin-mladenov: reverted to `fakeAsync`

projects/angular/data/datagrid/datagrid.ts
valentin-mladenov: Corrected in commit 75ce000

projects/angular/data/datagrid/datagrid.ts
valentin-mladenov: Corrected in commit 75ce000

projects/angular/data/datagrid/providers/selection.ts
valentin-mladenov: Corrected in commit 75ce000

projects/addons/datagrid/datagrid.component.ts
valentin-mladenov: Corrected in commit 75ce000

projects/angular/data/datagrid/providers/selection.ts
valentin-mladenov: Corrected in commit 75ce000

projects/angular/data/datagrid/datagrid.ts
```diff
@@ -207,30 +206,29 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
   /**
    * Array of all selected items
    */
```
dtsanevmw, line 213:
```suggestion
  /**
   * Selection type
   * - `None`: No rows are selectable.
   * - `Single`: Only one row can be selected at a time.
   * - `Multi`: Multiple rows can be selected.
   *
   * Defaults to `None`.
   */
```

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -1200,85 +1174,23 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
...
-        .subscribe((selected: T) =>
-          this.onSelectedItemsChange(this.isNotObjectOrEmptyObject(selected) ? [] : [selected])
-        );
-    } else if (this.#selectionType === SelectionType.Multi && this.clrDatagrid.selectedChanged) {
+    if (this.#selectionType !== SelectionType.None && this.clrDatagrid.selectedChanged) {
       this.#gridSelectionChangedSub = this.clrDatagrid.selectedChanged
         .pipe(takeUntil(this.#unsubscribeSubject))
         .subscribe((selected: T[]) => this.onSelectedItemsChange(selected));
     }
   }
```
dtsanevmw: Replace with binding in the template `(clrDgSelectedChange)="onSelectedItemsChange($event)"`

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -1165,27 +1154,12 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
...
 
-  private initGridSelection(changedGridSelectionType?: SelectionType): void {
+  private initGridSelection(): void {
     if (this.clrDatagrid) {
       this.clrDatagrid.identityFn = this.trackByGridItemFn;
     }
```
dtsanevmw, line 1146: Not needed as we have template binding with:
`[clrDgItemsIdentityFn]="trackByGridItemFn"`

projects/angular/data/datagrid/enums/selection-type.ts
```diff
@@ -6,7 +6,18 @@
...
+
+  /**
+   * User can select only one row at a time in the Datagrid.
+   */
+  Single = 1,
+
+  /**
+   * User can select any number of rows in the Datagrid.
+   */
+  Multi = 2,
```
dtsanevmw: The current approach where people should either set 0/1/2 or need to import the enum and set it like that ( creates more boiler plate code ) is not very efficient. Reviewing the PR it was strange for me where I saw inputs set to ="2" ="1". What is 1? What is 2? I don't thing that is the approach we want to go. Consider using string values for the enum.
```suggestion
  /**
   * User cannot select any row in the Datagrid
   */
  None = 'none',

  /**
   * User can select only one row at a time in the Datagrid.
   */
  Single = 'single',

  /**
   * User can select any number of rows in the Datagrid.
   */
  Multi = 'multi',
```

projects/demo/src/app/datagrid/virtual-scroll-client-side/virtual-scroll-client-side.ts
```diff
@@ -67,7 +67,7 @@ export class DatagridVirtualScrollClientSideDemo implements OnInit, AfterViewChe
     };
   }
 
-  trackItemById: ClrDatagridItemsIdentityFunction<User> = item => item?.id;
+  trackItemById: ClrDatagridItemsIdentityFunction<any> = item => item?.id;
```
dtsanevmw: Why should this become any?

projects/demo/src/app/datagrid/test-cases/test-cases.ts
```diff
@@ -18,7 +18,7 @@ import { User } from '../inventory/user';
   standalone: false,
 })
 export class DatagridTestCasesDemo {
-  selectedUser: User;
+  selectedUser: User[];
```
dtsanevmw: The naming selectedUser look strange for array.
```suggestion
  selectedUsers: User[];
```

projects/demo/src/app/datagrid/selection-single/selection-single.ts
```diff
@@ -69,9 +69,9 @@ export class DatagridSelectionSingleDemo {
...
   selectItems() {
     const selectedUser = { ...this.users[1] };
-    this.singleSelected = selectedUser;
-    this.trackByIndexSingleSelected = selectedUser;
-    this.trackByIdSingleSelected = selectedUser;
-    this.trackByIdServerSingleSelected = selectedUser;
+    this.singleSelected = [selectedUser];
+    this.trackByIndexSingleSelected = [selectedUser];
+    this.trackByIdSingleSelected = [selectedUser];
+    this.trackByIdServerSingleSelected = [selectedUser];
```
dtsanevmw, line 75: Same here. Having namings in single pronoun for array values sounds strange and wrong.

projects/angular/data/datagrid/datagrid.ts
```diff
@@ -350,10 +348,8 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
       this.stickyHeaders.changes.subscribe(() => this.resize()),
       this.stateProvider.change.subscribe(state => this.refresh.emit(state)),
       this.selection.change.subscribe(s => {
-        if (this.selection.selectionType === SelectionType.Single) {
-          this.singleSelectedChanged.emit(s as T);
-        } else if (this.selection.selectionType === SelectionType.Multi) {
-          this.selectedChanged.emit(s as T[]);
+        if (this.selection.selectable) {
+          this.selectedChanged.emit(s);
```
dtsanevmw: Can we change the parameter 's' to `selection` or something like that. I don't see why it's a `s` only.

projects/angular/data/datagrid/providers/selection.ts
```diff
@@ -204,35 +201,36 @@ export class Selection<T = any> {
   }
 
   get currentSingle(): T {
-    return this._currentSingle;
+    return this._current?.length ? this._current[0] : null;
```
dtsanevmw: I see we changed a lot of tests to go from undefined to null because of the fallback here. Do we gain something with changing it to null? With undefined we will minimise the changes and we are going to keep it was expected both for us and the users. It's strange to have it null if you don't have a set selection be default.  Not a big issue but a consideration.

projects/angular/data/datagrid/render/main-renderer.spec.ts
```diff
@@ -405,11 +406,11 @@ class RenderWidthTest {
   expandable = false;
   hasActions = false;
   selected: any[] = [];
-  singleSelect;
   @ViewChild('dgContainer', { read: ElementRef }) container: ElementRef<HTMLElement>;
   @ViewChild('datagridDefault', { read: ElementRef }) datagridDefault: ElementRef<HTMLElement>;
   @ViewChild('datagridSingleSelect', { read: ElementRef }) datagridSingleSelect: ElementRef<HTMLElement>;
   @ViewChild('datagridMultiSelect', { read: ElementRef }) datagridMultiSelect: ElementRef<HTMLElement>;
+  protected readonly SelectionType = SelectionType;
```
dtsanevmw, line 413: Not used?

projects/angular/data/datagrid/providers/selection.ts
```diff
@@ -270,11 +267,12 @@ export class Selection<T = any> {
...
     const ref = this._items.identifyBy(item);
-    if (this._selectionType === SelectionType.Single) {
-      return this.currentSingleSelectionRef === ref;
-    } else if (this._selectionType === SelectionType.Multi) {
+    if (this._selectionType === SelectionType.Multi) {
       return this.currentSelectionRefs.indexOf(ref) >= 0;
+    } else if (this._selectionType === SelectionType.Single) {
+      return this.currentSelectionRefs.indexOf(ref) === 0;
     }
+
```
dtsanevmw: Do we need the change here? Just changing order?

projects/angular/data/datagrid/providers/selection.ts
```diff
@@ -204,35 +201,36 @@ export class Selection<T = any> {
   }
 
   get currentSingle(): T {
-    return this._currentSingle;
+    return this._current?.length ? this._current[0] : null;
   }
   set currentSingle(value: T) {
```
dtsanevmw: Do we really need a setter for currentSingle after we are doing unification? In this way were are going to still have setting a single item on every place where `currentSingle = something` is used?

projects/angular/data/datagrid/providers/selection.ts
```diff
@@ -187,11 +182,13 @@ export class Selection<T = any> {
...
-    this._selectionType = value;
-    if ([SelectionType.None, SelectionType.Single].includes(value)) {
-      delete this.current;
-    } else {
-      this.updateCurrent([], false);
+
+    this._selectionType = value ?? SelectionType.None;
+
+    if (this._selectionType === SelectionType.None) {
+      this.current = null;
```
dtsanevmw: I know we talked about this but having `null` here will create by default without anything set for selection to have `null` instead of `undefined` which crates a change in user expectations without much benefit.

projects/angular/data/datagrid/providers/selection.ts
```diff
@@ -99,7 +94,7 @@ export class Selection<T = any> {
             updatedItems.forEach(item => {
               const ref = _items.identifyBy(item);
               // If one of the updated items is the previously selectedSingle, set it as the new one
-              if (this.currentSingleSelectionRef === ref) {
+              if (this.currentSelectionRefs.indexOf(ref) >= 0) {
```
dtsanevmw: If we know we are in a single selection mode, why iterate trough the array?
```suggestion
              if (this.currentSelectionRefs[0] === ref) {
```

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -593,7 +587,7 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
       // likely that the embedded Clarity grid doesn't yet exist, will manually set the `selectionType` later
       if (this.clrDatagrid && this.#gridSelectionChangedSub) {
         this.#gridSelectionChangedSub.unsubscribe();
-        this.initGridSelection(type);
+        this.initGridSelection();
       }
```
dtsanevmw: Likely won't be needed due to the template binding,.

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -851,7 +842,6 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
   }
 
   setSelectedItems(items: T[]) {
-    this.selectGridItems(items, false);
     this.selectedItemsChange.emit(items);
   }
```
dtsanevmw: The functionality of this function changes entirely to just emitting. Please delete it and directly emit change where is needed.

projects/angular/data/datagrid/datagrid.ts
```diff
@@ -207,30 +206,29 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
...
+    // in SINGLE SELECTION mode if the value is the same as currently selected do nothing
+    if (
+      this.selection.selectionType === SelectionType.Single &&
+      // if value is NULL compare it directly to current single if NOT take first index of the array
+      ((value && value[0] === this.selection.currentSingle) || value === this.selection.currentSingle)
+    ) {
+      return;
     }
+
+    this.selection.updateCurrent(value, true);
```
dtsanevmw: We need to set it to false as this is creating a feedback loop when we have two-way binding. There are going to be multiple events for a single action.
```suggestion
    this.selection.updateCurrent(value, false);
```

dtsanevmw: Feel free to cherry-pick the examples and tests that I added to see the issue: 3eda42865680bba41b1de2e6467750bb33ee547a

projects/angular/data/datagrid/datagrid.ts
```diff
@@ -207,30 +206,29 @@ export class ClrDatagrid<T = any> implements AfterContentInit, AfterViewInit, On
...
-      this.selection.currentSingle = value;
-    } else if (this.selection.currentSingle) {
-      this.selection.currentSingle = null;
+  @Input('clrDgSelected')
+  set selected(value: T[]) {
+    // in SINGLE SELECTION mode if the value is the same as currently selected do nothing
+    if (
+      this.selection.selectionType === SelectionType.Single &&
+      // if value is NULL compare it directly to current single if NOT take first index of the array
+      ((value && value[0] === this.selection.currentSingle) || value === this.selection.currentSingle)
```
dtsanevmw: When we set the emit to false the guard here doesn't make much sense and we can just compare it like this to check if the references matches. 
```suggestion
  if (value === this.selection.current) {
    return;
  }
```

dtsanevmw (commented):
Check comments

projects/angular/data/datagrid/datagrid.ts
dtsanevmw: Added more examples with manual assignment and one way binding for testing:
https://github.com/vmware-clarity/ng-clarity/commit/dd73cd2c9b4d0e9c649a7b98a185e2d5da79e714

projects/demo/src/app/datagrid/selection-single/selection-single.ts
valentin-mladenov: It's for single selection demo page.

projects/angular/data/datagrid/datagrid.ts
idonchev-vmw: We're changing one sub-optimal solution for another sub-optimal solution. This is a vicious cycle for a library product, which must be focused on both stable APIs and stable (final) solutions.
Exposing single selection in an array, which may contain more than one element, and this being "their error to make" does not make me feel positive that we're not coming back to redesigning this feature again any time soon.

projects/addons/datagrid/datagrid.component.ts
valentin-mladenov: `this.initGridSelection();` is needed. There is a special `firsItemPreSelected` functionality in that function.
Also `identityFn` is ste there, because the binding might not come before the selection.

projects/angular/data/datagrid/datagrid.ts
valentin-mladenov: Corrected in commit 956ba28.

projects/angular/data/datagrid/datagrid.ts
valentin-mladenov: Corrected in commit 956ba28.

projects/addons/datagrid/datagrid.component.ts
valentin-mladenov: I prefer to keep the assignent through datagrid API `onInit` as well, because there will be an issue with the pre-selection.

projects/addons/datagrid/datagrid.component.ts
valentin-mladenov: Corrected in commit 956ba28.

projects/angular/data/datagrid/datagrid.ts
valentin-mladenov: Corrected in commit 956ba28.

projects/addons/datagrid/datagrid.component.ts
valentin-mladenov: Corrected in commit 956ba28.

projects/angular/data/datagrid/providers/selection.ts
valentin-mladenov: Corrected in commit 62f80d4.

Current is set to `undefined`.

projects/angular/data/datagrid/providers/selection.ts
valentin-mladenov: Corrected in commit 62f80d4

projects/angular/data/datagrid/providers/selection.ts
valentin-mladenov: Corrected in commit 62f80d4.

projects/angular/data/datagrid/datagrid.ts
valentin-mladenov: Corrected in commit 62f80d4

projects/angular/data/datagrid/providers/selection.ts
valentin-mladenov: Corrected in commit 93cb304.

projects/demo/src/app/datagrid/test-cases/test-cases.ts
valentin-mladenov: Corrected in commit 0146a94

projects/demo/src/app/datagrid/virtual-scroll-client-side/virtual-scroll-client-side.ts
valentin-mladenov: Corrected in commit 0146a94

projects/angular/data/datagrid/datagrid.ts
valentin-mladenov: Corrected in commit 0146a94

projects/angular/data/datagrid/render/main-renderer.spec.ts
valentin-mladenov: Corrected in commit 14d850c

projects/angular/data/datagrid/enums/selection-type.ts
valentin-mladenov: Corrected in commit 14d850c

projects/angular/data/datagrid/utils/key-navigation-grid.controller.spec.ts
valentin-mladenov: Corrected in commit 14d850c

dtsanevmw (dismissed):

PR #2211, opened 2026-02-18
feat(combobox): add identity function

idonchev-vmw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-1026

- [ ] Yes
- [x] No

### DIFF
--- a/projects/angular/forms/combobox/combobox.ts
+++ b/projects/angular/forms/combobox/combobox.ts
@@ -124,9 +124,7 @@ export class ClrCombobox<T>
     }
 
     // default to SingleSelectComboboxModel, in case the optional input [ClrMulti] isn't used
-    optionSelectionService.selectionModel = new SingleSelectComboboxModel<T>();
-    optionSelectionService.selectionModel.identityFn = optionSelectionService.identityFn;
-    this.updateControlValue();
+    this.multiSelect = false;
   }
 
   @Input('clrEditable')
--- a/projects/angular/forms/combobox/providers/option-selection.service.ts
+++ b/projects/angular/forms/combobox/providers/option-selection.service.ts
@@ -132,13 +132,22 @@ export class OptionSelectionService<T> {
     if (current === value) {
       return true;
     }
-    // Check if both are null or undefined.
-    if ((current === null || current === undefined) && (value === null || value === undefined)) {
+    // Check if both are null or undefined or empty string.
+    if (
+      (current === null || current === undefined || current === '') &&
+      (value === null || value === undefined || value === '')
+    ) {
       return true;
     }
-    // Check if one is null or undefined and the other is not.
-    // We already checked for both being null or undefined.
-    if (current === null || current === undefined || value === null || value === undefined) {
+    // Check if one is null or undefined or empty string and the other is not.
+    if (
+      current === null ||
+      current === undefined ||
+      current === '' ||
+      value === null ||
+      value === undefined ||
+      value === ''
+    ) {
       return false;
     }
 


projects/angular/forms/combobox/model/single-select-combobox.model.ts
```diff
@@ -7,10 +7,13 @@
...
 import { ComboboxModel } from './combobox.model';
 
-export class SingleSelectComboboxModel<T> implements ComboboxModel<T> {
-  model: T;
+export class SingleSelectComboboxModel<T> extends ComboboxModel<T> {
+  override model: T;
 
   containsItem(item: T): boolean {
+    if (this.identityFn) {
+      return this.model !== null && this.model !== undefined && this.identityFn(this.model) === this.identityFn(item);
```
valentin-mladenov:
```suggestion
      return this.model ?? this.identityFn(this.model) === this.identityFn(item);
```

projects/angular/forms/combobox/providers/option-selection.service.ts
```diff
@@ -92,14 +104,16 @@ export class OptionSelectionService<T> {
...
-    // this later
+    if (!this.selectionModel) {
+      return;
+    }
 
-    // if selection is undefined, or its value hasn't changed, or changing from null <-> undefined, that's not really changing so we return
-    if (!this.selectionModel || this.selectionModel.model === value || (!this.selectionModel.model && !value)) {
+    const current = this.selectionModel.model;
+    const noChange =
+      current === value || (!current && !value) || (this._identityFn && this.valuesEqualByIdentity(current, value));
```
valentin-mladenov: Checks `current === value || (!current && !value)` are run inside `valuesEqualByIdentity` as well.
Run them at single place.

```suggestion
      // TODO : Run 'current === value || (!current && !value)' at one place 
      current === value || (!current && !value) || (this._identityFn && this.valuesEqualByIdentity(current, value));
```

projects/angular/forms/combobox/providers/option-selection.service.ts
```diff
@@ -115,4 +129,26 @@ export class OptionSelectionService<T> {
...
     return value as T;
   }
+
+  private valuesEqualByIdentity(current: T | T[], value: T | T[]): boolean {
+    if (!current !== !value) {
+      return false;
+    }
+    if (!current && !value) {
+      return true;
+    }
```
valentin-mladenov: Related to comment https://github.com/vmware-clarity/ng-clarity/pull/2211/changes#r2858304507
Run checks at single place.

```suggestion
// TODO: It's the same as 'current === value || (!current && !value)'
    if (!current !== !value) {
      return false;
    }
    if (!current && !value) {
      return true;
    }
```

projects/angular/forms/combobox/providers/option-selection.service.ts
```diff
@@ -115,4 +129,26 @@ export class OptionSelectionService<T> {
...
+      if (cur.length !== val.length) {
+        return false;
+      }
+      const curIds = cur.map(fn).sort();
+      const valIds = val.map(fn).sort();
+      return curIds.length === valIds.length && curIds.every((id, i) => id === valIds[i]);
+    } else {
+      return fn(current as T) === fn(value as T);
+    }
+  }
```
valentin-mladenov:
```suggestion
    if (!this.multiselectable) {
      return this._identityFn(current as T) === this._identityFn(value as T);
    }
    
    const cur = current as T[];
    const val = value as T[];
    if (cur.length !== val.length) {
      return false;
    }
    const curIds = cur.map(this._identityFn).sort();
    const valIds = val.map(this._identityFn).sort();
    return curIds.length === valIds.length && curIds.every((id, i) => id === valIds[i]);
  }
```

projects/angular/forms/combobox/providers/option-selection.service.ts
```diff
@@ -115,4 +129,26 @@ export class OptionSelectionService<T> {
...
+    }
+    const fn = this._identityFn;
+    if (this.multiselectable) {
+      const cur = current as T[];
+      const val = value as T[];
+      if (cur.length !== val.length) {
+        return false;
+      }
+      const curIds = cur.map(fn).sort();
+      const valIds = val.map(fn).sort();
```
valentin-mladenov: Why sorting? The order of the input values is important. The same array with different possitions of the same elements is different input.

projects/angular/forms/combobox/providers/option-selection.service.ts
```diff
@@ -115,4 +129,26 @@ export class OptionSelectionService<T> {
...
+      return curIds.length === valIds.length && curIds.every((id, i) => id === valIds[i]);
```
valentin-mladenov: `length` chack again, why?
```suggestion
      return curIds.every((id, i) => id === valIds[i]);
```

projects/angular/forms/combobox/combobox.ts
```diff
@@ -125,6 +125,7 @@ export class ClrCombobox<T>
 
     // default to SingleSelectComboboxModel, in case the optional input [ClrMulti] isn't used
     optionSelectionService.selectionModel = new SingleSelectComboboxModel<T>();
+    optionSelectionService.selectionModel.identityFn = optionSelectionService.identityFn;
     this.updateControlValue();
```
valentin-mladenov: Code duplicate. Run through `multiSelect` setter.

```suggestion
    // default to SingleSelectComboboxModel, in case the optional input [ClrMulti] isn't used
    this.multiSelect = false;
```

projects/angular/forms/combobox/providers/option-selection.service.ts
```diff
@@ -25,6 +25,7 @@ export class OptionSelectionService<T> {
 
   private _currentInput = '';
   private _displayField: string;
+  private _identityFn: ClrComboboxIdentityFunction<T>;
```
valentin-mladenov: Default reference identity function is better istead of constant checking for it. Just use it always.

 
```suggestion
  private _identityFn: ClrComboboxIdentityFunction<T> = item => item;
```

projects/angular/forms/combobox/providers/option-selection.service.ts
```diff
@@ -115,4 +129,26 @@ export class OptionSelectionService<T> {
     }
     return value as T;
   }
+
+  private valuesEqualByIdentity(current: T | T[], value: T | T[]): boolean {
```
valentin-mladenov, line 131: To me `valuesEqualByIdentity` is not really needed. The models `Single` and `Multi` should leverage the identification function themselfs.

projects/angular/forms/combobox/model/combobox.model.ts
```diff
@@ -5,9 +5,12 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
+export type ClrComboboxIdentityFunction<T> = (item: T) => any;
+
 export abstract class ComboboxModel<T> {
   model: T | T[];
   displayField?: string;
+  identityFn?: ClrComboboxIdentityFunction<T>;
```
valentin-mladenov: Can have default.

```suggestion
  identityFn?: ClrComboboxIdentityFunction<T> = item => item;
```

projects/angular/forms/combobox/model/single-select-combobox.model.ts
idonchev-vmw: Thanks for the suggestion! However, containsItem must return a boolean, and this.model is of type T. Using the nullish coalescing operator ?? here would return this.model if it's defined, which would result in a type mismatch.

projects/angular/forms/combobox/providers/option-selection.service.ts
idonchev-vmw: Great suggestion. I've refactored setSelectionValue to delegate all equality checks to valuesEqualByIdentity, making it the single source of truth for comparison logic. I also updated valuesEqualByIdentity to handle the null/undefined checks gracefully, removing the duplication.

projects/angular/forms/combobox/providers/option-selection.service.ts
idonchev-vmw: Same as above, thanks for the suggestion.

projects/angular/forms/combobox/model/combobox.model.ts
idonchev-vmw: Agreed. I've updated both ComboboxModel and OptionSelectionService to initialize identityFn with a default item => item function. This significantly simplifies the code by removing the need for null checks on identityFn.

projects/angular/forms/combobox/providers/option-selection.service.ts
idonchev-vmw: Agreed. I've updated both ComboboxModel and OptionSelectionService to initialize identityFn with a default item => item function. This significantly simplifies the code by removing the need for null checks on identityFn.

projects/angular/forms/combobox/providers/option-selection.service.ts
idonchev-vmw: This may be a valid architectural point. Ideally, the model should encapsulate all equality logic. However, the service is handling details around setting selection now and this logic is specific for the selection setting workflow. I also feel uneasy about propagading the identity functions yet another level down, so I've prefered to keep these checks in the service for now.

projects/angular/forms/combobox/providers/option-selection.service.ts
idonchev-vmw: On Input level, yes, but semantically, as selection, it's just a collection of elements. Why do you think the order in which you selected them is significant? Without the sort, I would have made it check for each element of one array if the other array contains the same item.

projects/angular/forms/combobox/providers/option-selection.service.ts
idonchev-vmw: Thanks, fixed.

projects/angular/forms/combobox/combobox.ts
idonchev-vmw: Good catch.

projects/angular/forms/combobox/providers/option-selection.service.ts
idonchev-vmw: Done

projects/angular/forms/combobox/combobox.ts
valentin-mladenov: Yuo can keep the commnet `// default to SingleSelectComboboxModel, in case the optional input [ClrMulti] isn't used`, but is up to you.

Resolve this comment, if it make no sense to have the comment.

projects/angular/forms/combobox/providers/option-selection.service.ts
valentin-mladenov: IN my suggestion the order is changed as well.

First we have 
```
    if (!this.multiselectable) {
      return this._identityFn(current as T) === this._identityFn(value as T);
    }
```

And then the other logic. Else is not needed because of the IFs `return`.

```
    const cur = current as T[];
    const val = value as T[];
    if (cur.length !== val.length) {
      return false;
    }
    const curIds = cur.map(this._identityFn).sort();
    const valIds = val.map(this._identityFn).sort();
    return curIds.length === valIds.length && curIds.every((id, i) => id === valIds[i]);
```

projects/angular/forms/combobox/providers/option-selection.service.ts
```diff
@@ -115,4 +125,31 @@ export class OptionSelectionService<T> {
...
     return value as T;
   }
+
+  private _identityFn: ClrComboboxIdentityFunction<T> = (item: T) => item;
+
+  private valuesEqualByIdentity(current: T | T[], value: T | T[]): boolean {
+    if (current === value) {
+      return true;
+    }
+    if (!current && !value) {
```
valentin-mladenov: IF `current = 0` and `value = null` the check will return `true`, which is not.

projects/angular/forms/combobox/providers/option-selection.service.ts
```diff
@@ -115,4 +125,31 @@ export class OptionSelectionService<T> {
...
+      return true;
+    }
+    if (!current || !value) {
+      return false;
+    }
```
valentin-mladenov, line 152: Can you add some comments about those checks use cases?

projects/angular/forms/combobox/providers/option-selection.service.ts
valentin-mladenov: It's an edge case, but still an important one.

projects/angular/forms/combobox/providers/option-selection.service.ts
idonchev-vmw: Well, this This case, the order is insignificant :)
And a matter of personal preference.

projects/angular/forms/combobox/providers/option-selection.service.ts
idonchev-vmw: Good catch, replaced with a more explicit one.

projects/angular/forms/combobox/providers/option-selection.service.ts
idonchev-vmw: Replaced with more readable versions and commented.

projects/angular/forms/combobox/providers/option-selection.service.ts
idonchev-vmw: Fell out, as we agreed to keep the order as significant.

projects/angular/forms/combobox/combobox.ts
idonchev-vmw: Yeah, the comment is stating the obvious now. I'll remove it.

idonchev-vmw: Oh, you meant take it back... I realized when I went in the code to remove it.
I'll spare a rebuild. It's an obvious default, IMO.

PR #2296, opened 2026-03-24
fix(datagrid): forced refresh emit on pagination empty state

valentin-mladenov:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

In `ngAfterViewInit`, the datagrid unconditionally emits a `clrDgRefresh` event with the current state (`this.stateProvider.state`). This means that even when the state is an empty object (i.e. no pagination, sorting, or filtering has been configured yet), a refresh event is emitted. This causes an unnecessary initial data fetch in server-driven datagrids that have no meaningful state to send.

Issue Number: CDE-3066

## What is the new behavior?
The `clrDgRefresh` event is now only emitted during `ngAfterViewInit` when the state object is **non-empty** (i.e. pagination, sorting, or filtering is already configured). This prevents the redundant initial refresh in cases where the datagrid state has no actionable information. Server-driven datagrids with pagination still receive the initial refresh because pagination populates the state before `ngAfterViewInit` runs, while datagrids without any pre-configured state skip the unnecessary emit.

- [ ] Yes
- [x] No

## Other information

Consumers that relied on the unconditional initial `clrDgRefresh` emit (with an empty state) to trigger a first data load should ensure their pagination (or other state) is configured before the datagrid initializes so that the state is non-empty and the refresh fires as expected. Server-driven datagrids that set `[clrDgTotalItems]` and page size will continue to work without changes.


idonchev-vmw (approved):
It's an ancient one. Thanks! :)

dtsanevmw (approved):

PR #2019, opened 2025-09-25
feat(combobox): add editable state

dtsanevmw:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

Allows the combobox to be editable like in https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/ and it can replace datalist in some cases.

- [ ] Bugfix
- [X] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: CDE-2567

## Does this PR introduce a breaking change?

- [ ] Yes
- [X] No

PR #2310, opened 2026-03-30
feat(combobox): add editable resolver and fix identity compatibility

dtsanevmw:
## Summary

- The identity function introduced in #2211 broke the editable combobox feature from #2019. When `identityFn` is used with editable mode, `parseStringToModel` creates objects lacking identity-relevant fields, causing `containsItem` to malfunction
- Adds `[clrEditableResolver]` input so consumers can provide a function `(input: string) => T` to construct proper model objects from free-text input
- Fixes duplicate entry prevention when editable mode creates objects via `displayField` without a custom `identityFn`
- Adds website documentation, demos, storybook stories, and unit tests for editable combobox

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature

Issue Number: CDE-2567

When using `[clrEditable]="true"` together with `[clrComboboxIdentityFn]`, the editable combobox is broken:
1. `parseStringToModel` creates objects with only the `displayField` set, but the `identityFn` may reference a different field (e.g. `item.id`), causing `containsItem` to return incorrect results
2. `SingleSelectComboboxModel.containsItem` guards against `null` but the initial model state is `undefined`, causing `identityFn(undefined)` crashes
3. Without a custom `identityFn`, editable mode with object values allows duplicate entries because the default `item => item` identity compares by reference

## What is the new behavior?

1. New `[clrEditableResolver]` input allows consumers to provide a `(input: string) => T` function that creates properly formed model objects from free-text
2. `containsItem` now guards against both `null` and `undefined`
3. When editable mode creates objects via `displayField` and no custom `identityFn` is provided, the identity function is auto-upgraded to use the `displayField` for comparison, preventing duplicates
4. Website documentation with two new demos (single-select editable, multi-select editable with resolver) and API table entries for `[clrEditable]` and `[clrEditableResolver]`

- [ ] Yes
- [x] No

## Test plan

- [x] Verify single-select editable combobox accepts free-text values
- [x] Verify multi-select editable combobox with resolver creates proper objects on Enter
- [x] Verify duplicate entries are blocked when typing the same value twice
- [x] Verify `identityFn` + `editableResolver` combination works correctly
- [x] Verify pre-existing combobox behavior (non-editable) is unaffected
- [x] Unit tests pass (`ng test clr-angular --include='**/combobox/**/*.spec.ts'`)

Made with [Cursor](https://cursor.com)


projects/angular/forms/combobox/model/single-select-combobox.model.ts
```diff
@@ -11,7 +11,10 @@ export class SingleSelectComboboxModel<T> extends ComboboxModel<T> {
   override model: T;
 
   containsItem(item: T): boolean {
-    return this.model !== null && this.identityFn(this.model) === this.identityFn(item);
+    if (!this.model) {
```
valentin-mladenov: If `T` is number and `model= 0` then `!this.model` will return false, but in fact it may be not.

We need a test about that case for single selection.

projects/angular/forms/combobox/combobox.ts
```diff
@@ -135,6 +135,11 @@ export class ClrCombobox<T>
     this.optionSelectionService.editable = value;
   }
 
+  @Input('clrEditableResolver')
+  set editableResolver(value: ((input: string) => T) | undefined) {
+    this.optionSelectionService.editableResolver = value;
```
valentin-mladenov: `editableResolver` should have `Fn` at the end. Similar to `indentityFn`.
```suggestion
  @Input('clrEditableResolverFn')
  set editableResolverFn(value: ((input: string) => T) | undefined) {
    this.optionSelectionService.editableResolverFn = value;
```

projects/angular/forms/combobox/providers/option-selection.service.ts
```diff
@@ -18,6 +18,7 @@ export class OptionSelectionService<T> {
   filtering = true;
   selectionModel: ComboboxModel<T>;
   inputChanged: Observable<string>;
+  editableResolver: ((value: string) => T) | undefined;
```
valentin-mladenov: Will it make sense to have `ClrComboboxEditableResolverFunction` similar to `ClrComboboxIdentityFunction` in the combobox model.

```
export type ClrComboboxResolverFunction<T> = (value: string) => T;
```


```suggestion
  editableResolver: ClrComboboxResolverFunction<T> | undefined;
```

projects/angular/forms/combobox/providers/option-selection.service.ts
valentin-mladenov: SInce it's not cristal clear to me this comment is open for discussion. `parseStringToModel` is basivally a builder function for a string input that is not in the list.

We have the `editableResolver` funcion which may as a default to `return value as T;`, if `displayField` is introduced then it can be upgraded to ` return { [this.displayField]: value } as T;` and finally if it is introduced via `clrEditableResolver` will be used as is.

```suggestion
  parseStringToModel(value: string): T {
    return this.editableResolver(value);
  }
```

projects/angular/forms/combobox/model/single-select-combobox.model.ts
dtsanevmw: Fixed in [bfe6028](https://github.com/vmware-clarity/ng-clarity/pull/2310/commits/bfe6028393f0a19e763f49619b098655a754ce55). The idea was to exclude both null and undefined

projects/angular/forms/combobox/combobox.ts
dtsanevmw: [bfe6028](https://github.com/vmware-clarity/ng-clarity/pull/2310/commits/bfe6028393f0a19e763f49619b098655a754ce55)

projects/angular/forms/combobox/providers/option-selection.service.ts
dtsanevmw: Added in [bfe6028](https://github.com/vmware-clarity/ng-clarity/pull/2310/commits/bfe6028393f0a19e763f49619b098655a754ce55)

projects/angular/forms/combobox/providers/option-selection.service.ts
dtsanevmw: Fixed in [bfe6028](https://github.com/vmware-clarity/ng-clarity/pull/2310/commits/bfe6028393f0a19e763f49619b098655a754ce55). the displayField condition was missed from the old implementation. It should be either the resolver or directly the string value.

valentin-mladenov (dismissed):

valentin-mladenov (dismissed):

PR #2363, opened 2026-05-11
fix(appfx-datagrid): expanded state lost on refresh

idonchev-vmw:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When using `appfx-datagrid` with expandable rows, the expanded rows intermittently collapse when a data refresh occurs from the server. This happens specifically when the datagrid uses `trackByGridItemProperty` and the server returns new object references for the same data items. 
The root cause is that the internal `trackByFn` used `Object.prototype.hasOwnProperty.call()` to resolve the tracking property. This approach fails to find properties defined on the prototype chain and fails silently when encountering nested `null` or `undefined` values. When the check fails, the datagrid falls back to tracking by object reference. Consequently, Angular's `NgForOf` destroys the old `ClrDatagridRow` components and creates new ones, which causes the expansion state (held by the row's `DatagridIfExpandService`) to be lost.

Issue Number: VCFUI-15272

## What is the new behavior?
The `trackByFn` in `appfx-datagrid` has been updated to use the `in` operator instead of `hasOwnProperty`, allowing it to correctly resolve properties on the prototype chain. It also includes safe null-checking to traverse nested properties without failing. 
Additionally, the `preselectDetail` method in `DatagridComponent` and the `getItemUniquePropertyValue` method in `DatagridPreserveSelectionDirective` have been updated to use this corrected resolution logic, ensuring that row instances, selection state, and expansion state are properly preserved across data refreshes.

- [ ] Yes
- [x] No

## Other information
A systematic reproduction was created and verified locally using Playwright to ensure the expansion state is maintained across 10 consecutive simulated data refreshes with partial object reference updates.


projects/addons/datagrid/preserve-selection/datagrid-preserve-selection.directive.ts
```diff
@@ -188,6 +188,23 @@ export class DatagridPreserveSelectionDirective implements AfterViewInit, OnDest
...
+    let parseValid = true;
+    const value = propertyPath.split(separator).reduce((o: any, i: string) => {
+      if (o !== null && o !== undefined && typeof o === 'object' && i in o) {
+        return o[i];
+      }
+      parseValid = false;
+      return undefined;
+    }, item);
+
+    return parseValid ? value : '';
```
valentin-mladenov: Looks exactly like the reduce function in `trackByFn` in `projects/addons/datagrid/datagrid.component.ts`.

It will make sense to extract it in common place instead of dublicate it.

```suggestion
    return propertyPath.split(separator).reduce((o: any, i: string) => {
      if (o !== null && o !== undefined && typeof o === 'object' && i in o) {
        return o[i];
      }

      return '';
    }, item);
```

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -1111,13 +1113,11 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
...
     const isDetailEnabled: boolean = !!this.detailHeader || !!this.detailBody;
-    if (isDetailEnabled && this.detailState && this.trackByGridItemProperty) {
+    if (isDetailEnabled && this.detailState && (this.trackByGridItemProperty || this.trackByFunction)) {
       const gridItems: T[] = this.gridItems || [];
 
       const matchingItems: T[] = gridItems.filter((item: T): boolean => {
-        return (
-          (item as any)[this.trackByGridItemProperty] === (this.detailState as any)?.[this.trackByGridItemProperty]
-        );
+        return this.trackByFn(0, item) === this.trackByFn(0, this.detailState as T);
```
valentin-mladenov: Wouldn't this check work with first row only?

projects/addons/datagrid/preserve-selection/datagrid-preserve-selection.directive.ts
idonchev-vmw: If there was already a common utility, I would have extracted it now. But I decided not to introduce one especially for this case, and not since The Rule of Three -- https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming) -- is unsatisfied.

projects/addons/datagrid/datagrid.component.ts
idonchev-vmw: No, it will work for all rows. The 0 is just a dummy index required to satisfy the trackByFn signature.

It will only break in the case a custom trackBy function relies ONLY on the index. 
This is an unlikely, but possible scenario. Nevertheless, as we don't have the original index, there is nothing we can do to address this problem without further concealing it. The filter index is not the original index, so using it will be actually a harder to detect problem than a solution.

So, by passing the 0, we just state - "We can't provide a valid index here."

projects/addons/datagrid/preserve-selection/datagrid-preserve-selection.directive.ts
valentin-mladenov: But it violates the DRY principle: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself.

In this instance the Rule of Three doesn't apply since we already have well established application.

projects/addons/datagrid/datagrid.component.ts
valentin-mladenov: But `0` is valid index. If the requirement is to pass an invalid index then pass `-1`. 

Going deeper in that logic rabbit hole. 
Why do we use `trackByFn`, when clearly we don't want to use the `trackByFunction` part of it (by providing an invalid index)?

idonchev-vmw: Actually, we have an existing function trackByGridItemFn, which we should use instead.
By coincidence, it is resolving to this.trackByFn(0, item), but this seems to be the original approach in the package. Btw, I agree that -1 is better, I was even considering it myself, but as we already have an in-production implementation for trackByGridItemFn, I'd prefer to switch to it.

projects/addons/datagrid/preserve-selection/datagrid-preserve-selection.directive.ts
idonchev-vmw: I don't want to fight over principles. I explained my decision (yes, it was a deliberate decision). I will implement a common helper, as I don't have strong preference for either of the solutions. But I would be happy if we start respecting each others' engineering decisions more in future, after a suggestion is confirmed to not have been overlooked, but already considered.

idonchev-vmw: Also, the explicit, in-band error state (TBD) is important, because it makes difference between having a non-matching (invalid) selector, vs selector resolving to empty string "". So I will keep the more elaborate syntax.

projects/addons/datagrid/utils/property-resolver.util.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+    class Child extends Parent {
+      name = 'child';
+    }
+    const item = new Child();
+
+    expect(getNestedProperty(item, 'name')).toEqual({ value: 'child', isValid: true });
+    expect(getNestedProperty(item, 'id')).toEqual({ value: 'parent-id', isValid: true });
+  });
+});
```
dtsanevmw, line 85: **Coverage gap: the bug itself isn't covered by an integration test.**

This PR adds 85 lines of unit tests for the new utility — great — but the actual reported regression ("expanded rows collapse on refresh when the server returns new object references") isn't verified anywhere. If someone refactors `trackByFn` again, the unit tests on `getNestedProperty` will keep passing while the real bug silently returns.

Please consider adding at least one test in `datagrid.component.spec.ts` that:

1. Renders an `appfx-datagrid` with `trackByGridItemProperty` and `[detailState]`.
2. Expands a row.
3. Reassigns `gridItems` to a new array of new object references with the same tracking property values.
4. Asserts `detailState` is preserved and the same `ClrDatagridRow` instance is reused (not re-created).

A similar test would be valuable for the `preserve-selection` directive's selection-survival path. The Playwright reproduction mentioned in the PR description should ideally be codified as a Karma spec so it runs in CI.

projects/addons/datagrid/utils/property-resolver.util.ts
```diff
@@ -0,0 +1,38 @@
...
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+/**
+ * Result of a nested property resolution.
+ */
+export interface PropertyResolutionResult<T = any> {
+  value: T | undefined;
+  isValid: boolean;
+}
```
dtsanevmw: **Type-safety nit on `PropertyResolutionResult`.**

The current shape lets you construct `{ value: 'something', isValid: false }`, and the `T | undefined` on `value` forces every caller to either re-narrow or cast. A discriminated union expresses intent better and gives callers free narrowing:

```ts
export type PropertyResolutionResult<T = unknown> =
  | { isValid: true; value: T }
  | { isValid: false };
```

Then at the call sites:

```ts
const result = getNestedProperty(gridItem, this.trackByGridItemProperty);
if (result.isValid) {
  return result.value; // value is T, no cast needed
}
```

Minor — but since this is becoming a shared util, worth getting the API right up front. Also, since the util is currently internal to `datagrid/utils`, consider not exporting `PropertyResolutionResult` from a barrel until you actually need it externally; keeps the public surface of the addon small.

projects/addons/datagrid/utils/property-resolver.util.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @returns An object containing the resolved value and a boolean indicating if the path was valid.
+ */
+export function getNestedProperty<T = any>(item: any, propertyPath: string): PropertyResolutionResult<T> {
+  if (!item || !propertyPath) {
+    return { value: undefined, isValid: false };
+  }
+
+  let isValid = true;
+  const value = propertyPath.split('.').reduce((o: any, i: string) => {
+    if (o !== null && o !== undefined && typeof o === 'object' && i in o) {
```
dtsanevmw, line 27: **Lost context: the `// rethink constant when properties have "."` comment was removed.**

The original `trackByFn` carried this comment for a reason — a property literally named `"foo.bar"` (or any key containing a dot) cannot be expressed in this dot-path syntax. By moving the logic into a generic `getNestedProperty` util, that limitation is now baked into the util with no warning.

Two options:

1. Re-add the warning as a JSDoc note on `getNestedProperty`:
   ```ts
   /**
    * @remarks Property names containing `.` are not supported (always treated as a path separator).
    */
   ```
2. Better: accept an optional `separator` parameter (defaulting to `'.'`) so consumers can opt out:
   ```ts
   export function getNestedProperty<T = unknown>(
     item: unknown,
     propertyPath: string,
     separator = '.',
   ): PropertyResolutionResult<T> { ... }
   ```
   This costs nothing and future-proofs the util.

projects/addons/datagrid/preserve-selection/datagrid-preserve-selection.directive.ts
```diff
@@ -188,6 +190,12 @@ export class DatagridPreserveSelectionDirective implements AfterViewInit, OnDest
     if (typeof this.selectBy === 'function') {
       return (<any>this.selectBy)(0, item || {});
     }
-    return item ? item[<any>this.selectBy] : '';
+
+    if (!item || !this.selectBy) {
+      return '';
```
dtsanevmw: Small redundancy — `getNestedProperty` already short-circuits on falsy `item` and empty `propertyPath` and returns `{ isValid: false }`. The early `if (!item || !this.selectBy)` check above is duplicating that guard. You can drop it and let the helper handle both cases:

```ts
private getItemUniquePropertyValue(item: any): string {
  if (typeof this.selectBy === 'function') {
    return (this.selectBy as any)(0, item || {});
  }

  const result = getNestedProperty(item, this.selectBy as string);
  return result.isValid ? result.value : '';
}
```

Also: the function is typed `: string`, but `result.value` can be a `number`/`boolean`/object. Pre-existing issue (the previous `item[selectBy]` had the same problem), so feel free to leave it — but if you're touching it anyway, a `String(result.value)` coercion would make the contract honest.

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -1111,13 +1103,11 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
 
   private preselectDetail(): void {
     const isDetailEnabled: boolean = !!this.detailHeader || !!this.detailBody;
-    if (isDetailEnabled && this.detailState && this.trackByGridItemProperty) {
+    if (isDetailEnabled && this.detailState && (this.trackByGridItemProperty || this.trackByFunction)) {
```
dtsanevmw, line 1110: Good call switching to `trackByGridItemFn` on line 1110 — and adding `|| this.trackByFunction` here is a real behavioral improvement that wasn't called out in the PR description: `preselectDetail` now works for consumers who only provide `trackByFunction`, which previously silently did nothing. Worth mentioning in the commit message / changelog so it's not surprising in a `fix(...)` patch.

Secondary note (not blocking, since you already discussed this with @valentin-mladenov): `trackByGridItemFn` calls `trackByFn(0, item)`. For a custom `trackByFunction` that genuinely uses the index, both items get index `0` here, so the equality check still works (both sides degenerate the same way) — but a user-defined `trackByFunction` that returned `index + '_' + item.id` would silently mis-match. A brief JSDoc on `trackByGridItemFn` clarifying *"intended for identity comparison only; index is unstable"* would help future maintainers avoid that footgun.

dtsanevmw (commented):
Nice extraction of `getNestedProperty` into a tested utility — that part is clean. A few suggestions focused on test coverage of the actual bug, a small type-safety improvement on the result shape, one piece of dropped context, and a minor redundancy in the directive. Inline notes below.

projects/addons/datagrid/utils/property-resolver.util.spec.ts
idonchev-vmw: This one was handled.
- An integration test was added right after origin/main.
- It was proven to fail.
- The rest of the code was interactively rebased on top.
- Tests were run again and passing.

projects/addons/datagrid/utils/property-resolver.util.ts
valentin-mladenov: Great points.

projects/addons/datagrid/utils/property-resolver.util.ts
```diff
@@ -0,0 +1,38 @@
...
+      return o[i];
+    }
+    isValid = false;
+    return undefined;
+  }, item);
+
+  return { value, isValid };
+}
```
valentin-mladenov: As the function is written, I don't see real need to have `value` and `isValid`.
Just return the value if found on `undefined` if not found.

```suggestion
export function getNestedProperty<T = any>(item: any, propertyPath: string): T {
  if (!item || !propertyPath) {
    return undefined;
  }

  return propertyPath.split('.').reduce((o: any, i: string) => {
    if (o !== null && o !== undefined && typeof o === 'object' && i in o) {
      return o[i];
    }
    return undefined;
  }, item);
}
```

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -1070,19 +1071,10 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
...
-        .reduce(
-          (o, i) => (parseValid = Object.prototype.hasOwnProperty.call(o as object, i)) && (o as any)[i],
-          gridItem
-        );
-
-      if (parseValid) {
-        return observedPropertyValue;
+      const result = getNestedProperty(gridItem, this.trackByGridItemProperty);
+      if (result.isValid) {
+        return result.value;
```
valentin-mladenov: Based on my other comment https://github.com/vmware-clarity/ng-clarity/pull/2363/changes#r3248678757.

```suggestion
      const result = getNestedProperty(gridItem, this.trackByGridItemProperty);
      if (result !== undefined) {
        return result;
```

projects/addons/datagrid/preserve-selection/datagrid-preserve-selection.directive.ts
```diff
@@ -188,6 +190,12 @@ export class DatagridPreserveSelectionDirective implements AfterViewInit, OnDest
...
+    }
+
+    const result = getNestedProperty(item, this.selectBy as string);
+    return result.isValid ? result.value : '';
```
valentin-mladenov: Based on my other comment https://github.com/vmware-clarity/ng-clarity/pull/2363/changes#r3248678757.


```suggestion
    const result = getNestedProperty(item, this.selectBy as string);
    return result === undefined ? '' :  result;
```

valentin-mladenov (commented):
Great work. I wrote 3 connected extra comments.

projects/addons/datagrid/utils/property-resolver.util.ts
valentin-mladenov: So my bad. The key might have an `undefined` vlaue. `isValid` is needed.

projects/addons/datagrid/utils/property-resolver.util.ts
idonchev-vmw: handled with 54a7615e6f13cbdc13cc80adc63898e6228a2199

projects/addons/datagrid/utils/property-resolver.util.ts
idonchev-vmw: I do not think this is something we should support. 
my.foo.bar is okay and we support it.
my["foo.bar"] is an overkill.

projects/addons/datagrid/preserve-selection/datagrid-preserve-selection.directive.ts
idonchev-vmw: handled with https://github.com/vmware-clarity/ng-clarity/commit/54a7615e6f13cbdc13cc80adc63898e6228a2199

projects/addons/datagrid/datagrid.component.ts
idonchev-vmw: The PR description will remain concise. 
The rest is handled in 17d288dacd632cbf106441625db6a4da31278c37

idonchev-vmw: Actually, the description is not concise. I will add the requested note.

projects/addons/datagrid/utils/property-resolver.util.ts
idonchev-vmw: Even in the very rare case, where an automated tool may extract key like "foo.bar" from a database, the better pattern is that this key is process before adding it to the data structure, so instead of the weird
my["foo.bar"]
the user would work with
my.foo_bar 
and handle the parsing client side, instead of supplying the library with unintelligible properties like "my#foo.bar"

dtsanevmw (approved):

PR #2376, opened 2026-05-12
chore(website): update density and theme selectors

valentin-mladenov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


valentin-mladenov:
Merge after #2381

projects/website/src/app/shared/density-toggle/density-toggle.component.ts
```diff
@@ -16,32 +15,24 @@ const densityLocalStorageKey = 'density';
...
-      <clr-dropdown-menu clrPosition="bottom-left" *clrIfOpen>
-        @for (option of densityOptions; track option) {
-          <div (click)="setTheme(option)" clrDropdownItem>
-            {{ !!option === false ? 'Default' : (option | titlecase) }}
-          </div>
-        }
-      </clr-dropdown-menu>
-    </clr-dropdown>
+    <a
+      href="javascript://"
```
dtsanevmw: **Use `<button>` instead of `<a href="javascript://">` for consistency with `theme-toggle.component.ts`.**

`theme-toggle` uses `<button class="btn btn-link nav-link">` for the exact same kind of action (toggle a state, no navigation). The two toggles sit next to each other in the header and should render/behave identically.

Additionally, `href="javascript://"` is an anti-pattern:
- ESLint `no-script-url` flags it.
- It's bad for accessibility (screen readers announce a link, not a control).
- It's a CSP/XSS smell.

Suggested replacement (mirrors `theme-toggle`):

```html
<button
  class="btn btn-link nav-link toggle-density"
  [class.toggle-density--compact]="density === 'compact'"
  [attr.aria-label]="'toggle to ' + (density === 'compact' ? 'default' : 'compact') + ' density'"
  (click)="toggleDensity()"
  (mouseenter)="iconInverse = true"
  (mouseleave)="iconInverse = false"
>
```

projects/website/src/app/shared/density-toggle/density-toggle.component.ts
```diff
@@ -16,32 +15,24 @@ const densityLocalStorageKey = 'density';
...
+      class="nav-link nav-text toggle-density"
+      [class.toggle-density-compacted]="density === 'compact'"
```
dtsanevmw: **Class name `toggle-density-compacted` reads wrong.**

"Compacted" isn't the right adjective — the state, the `DensityType` value, and the `clr-density` body attribute are all `compact`. The class should match.

Also prefer a BEM-style modifier so the relationship to the base class is explicit:

```ts
[class.toggle-density--compact]="density === 'compact'"
```

or simply `is-compact`. Same change applies to the SCSS selector below.

projects/website/src/app/shared/density-toggle/density-toggle.component.ts
```diff
@@ -16,32 +15,24 @@ const densityLocalStorageKey = 'density';
...
+      [attr.aria-label]="'toggle to ' + (density === 'compact' ? 'Compact' : 'Default') + ' density'"
```
dtsanevmw: **Bug: the `aria-label` is inverted.**

Current expression:

```ts
'toggle to ' + (density === 'compact' ? 'Compact' : 'Default') + ' density'
```

When `density === 'compact'`, this announces *"toggle to Compact density"* — but clicking will switch it to default. Compare with `theme-toggle.component.ts`, which correctly announces the *target* state:

```ts
'toggle to ' + (theme === 'light' ? 'dark' : 'light') + ' theme'
```

Fix:

```ts
[attr.aria-label]="'toggle to ' + (density === 'compact' ? 'default' : 'compact') + ' density'"
```

Alternatively (and arguably better, since this is a binary on/off control): drop the dynamic `aria-label` entirely and use the switch pattern:

```html
<button
  role="switch"
  [attr.aria-checked]="density === 'compact'"
  aria-label="compact density"
  ...
>
```

That's both simpler and more semantically correct than a link-with-label.

projects/website/src/app/shared/density-toggle/density-toggle.component.ts
```diff
@@ -50,48 +41,52 @@ const densityLocalStorageKey = 'density';
...
 })
 export class DensityToggleComponent implements OnInit {
   protected density = getPreferredDensity();
-  protected themeIconInverse = true;
-  protected densityOptions: DensityType[] = ['', 'compact'];
+  protected iconInverse = true;
 
   constructor() {
     ClarityIcons.addIcons(landscapeIcon);
   }
```
dtsanevmw, line 66: Minor consistency nit: `theme-toggle` initialises `themeIconInverse = false` and uses `mouseenter → true / mouseleave → false`. Here it's the opposite (`iconInverse = true`, `mouseenter → false`). Same visual effect, but flipping the semantics so both components match would make them easier to reason about together.

projects/website/src/app/shared/density-toggle/density-toggle.component.ts
```diff
@@ -50,48 +41,52 @@ const densityLocalStorageKey = 'density';
...
+    const notInCompactModeWithHover = this.density !== 'compact' && this.iconInverse;
+
+    return inCompactModeWithoutHover || notInCompactModeWithHover;
+  }
+
   ngOnInit() {
-    this.setTheme(this.density);
+    this.setDensity(this.density);
+  }
+
```
dtsanevmw, line 77: This getter can be collapsed — the two branches are just an XOR between "is compact" and "icon is inverse":

```ts
protected get solidDensityIcon() {
  return (this.density === 'compact') !== this.iconInverse;
}
```

Much easier to read than the two named intermediate booleans, and removes the need for the verbose variable names that don't really clarify intent.

(If you flip `iconInverse` semantics per the other comment, this becomes `(this.density === 'compact') === this.iconInverse` — pick whichever direction matches the design.)

dtsanevmw (commented):
Nice cleanup — dropping the dropdown for a switch is the right call. A few concerns, mainly around consistency with `theme-toggle.component.ts` (which is doing the exact same thing) and one functional bug in the `aria-label`. Detailed notes inline.

projects/website/src/app/shared/density-toggle/density-toggle.component.ts
valentin-mladenov: I made the rules explicitly verbose to make the code self documenting insted for writing comments why the rule is written like `(this.density === 'compact') !== this.iconInverse`, because it is not immediately obvious.

projects/website/src/app/shared/density-toggle/density-toggle.component.ts
valentin-mladenov: That's because the landscape icon had SVG mixed up. Check PR #2381

valentin-mladenov: Corrected in commit 264841a

projects/website/src/app/shared/density-toggle/density-toggle.component.ts
valentin-mladenov: Corrected in commit 264841a

projects/website/src/app/shared/density-toggle/density-toggle.component.ts
valentin-mladenov: Corrected in commit 264841a

projects/website/src/app/shared/density-toggle/density-toggle.component.ts
valentin-mladenov: Corrected in commit 264841a

dtsanevmw (approved):

PR #2391, opened 2026-05-18
feat(ng): add Angular 20-21 support

dtsanevmw:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [x] Build related changes
- [x] CI related changes

Angular 20 and 21 are not accepted by Clarity's peer dependencies, and the Datagrid virtual scroll fails at runtime under CDK 20+ due to internal CDK API changes (the `_VIEW_REPEATER_STRATEGY` injection token was removed, `CdkVirtualForOf` now injects its parent viewport via the new `CDK_VIRTUAL_SCROLL_VIEWPORT` token rather than the `CdkVirtualScrollViewport` class, and `_totalContentHeight` was turned into a `Signal<string>`). Several internal Angular 21 changes (zoneless TestBed default, refreshed CLDR data, control-flow template migration, stricter projection resolver, stricter host-binding type-check) also break a handful of shipped templates and the test suite when consumers run on v20/v21.

Issue Number: CDE-3098

## What is the new behavior?

**Peer dependencies and CI**

- `@angular/core | @angular/common | @angular/cdk` peer ranges extended to `15 || 16 || 17 || 18 || 19 || 20 || 21`.
- `pr-build.yml` matrix gains an `Angular v20` and `Angular v21` row; Node 20 is provisioned for the v20 row and Node 22 for the v21 row (matching Angular's official supported Node ranges). `actions/setup-node@v3` bumped to `@v4`.

**Virtual scroll directive (`ClrDatagridVirtualScrollDirective`)**

- `totalContentHeight` getter handles both the CDK 15-19 plain `string` form and the CDK 20+ `Signal<string>` form, while keeping the public return type as `string`.
- `_VIEW_REPEATER_STRATEGY` and `CDK_VIRTUAL_SCROLL_VIEWPORT` are resolved at module load via indexed access against the namespace imports, so the same build loads cleanly against CDK 15 (no viewport-token export) and CDK 21 (no view-repeater-strategy export).
- For CDK ≥ 19 the viewport is provided under both the `CdkVirtualScrollViewport` class token (used by `CdkVirtualForOf` in CDK 15-20) and the `CDK_VIRTUAL_SCROLL_VIEWPORT` token (used in CDK 21+). Selection is feature-detected, so mixed `@angular/core` / `@angular/cdk` major versions allowed by the relaxed peer range still resolve correctly.
- The `_VIEW_REPEATER_STRATEGY` provider is included only when the token still exists (CDK 19-20); CDK 21+ skips it because `CdkVirtualForOf` constructs its own strategy internally.
- New unit test asserts `totalContentHeight` returns `string` for the plain-string form, the `Signal<string>` form, and the missing-viewport fallback.

**Template fixes for v21 projection / migration**

- `wizard.html`: replaces `*ngIf="cond; then T; else T2"` with two independent `*ngIf` + `ngTemplateOutlet`. The Angular 17+ control-flow migration inlines both branches' content, producing duplicate `<ng-content>` slots whose projection chain resolves to the wrong slot under v21. The footer-button branch keeps the `navService.currentPage &&` guard so the no-buttons branch does not flash while `currentPage` is briefly undefined.
- `modal.html`: switches the `modalContentTemplate` branch to a direct `[ngTemplateOutlet]`, avoiding the same v21 projection-resolver mismatch when a content template is provided.

**Test-suite fixes for v21**

- `projects/angular/src/test.ts` and `projects/addons/test.ts`: install `provideZoneChangeDetection` (runtime-feature-detected so v15/v16 are unaffected), restoring Zone.js-based change detection under v21's zoneless TestBed default. Prevents `NG0100` in components that mutate state during `ngAfterViewInit` / `ngAfterContentInit`.
- `outside-click.spec.ts` and `abstract-popover.spec.ts`: zoneless TestBed in v21 makes `ApplicationRef.tick` callable for runtime-driven reasons. The strict CD-count assertion is kept for v15-v20 and falls back to a behavioral assertion for v21+. Spy target remains `ApplicationRef.tick`.
- `datagrid.component.spec.ts`: a `.then(...)` chained inside an `it()` is replaced with `await fixture.whenStable()` so the assertion runs before `afterEach` nullifies `this.component` (otherwise Jasmine surfaces an unhandled error in v21).
- `date-io.service.spec.ts`: locale `ak`'s short-date format changed in v21's CLDR data and became indistinguishable from `en-US`; replaced with `hr`, which is stable across versions. Dead `localeAk` import/register removed.
- `date-time.service.spec.ts`: relaxes the `GMT` vs. `GMT+00:00` assertion regex to accept both Chrome/Intl outputs across versions.
- `datagrid-content-no-wrap.directive.spec.ts`: removes the duplicate declaration of `DatagridContentNoWrapDirective` (already declared in `AppfxDatagridModule`); under v21 bundling, both module compilation contexts collide with "directive is part of 2 modules".

**Build configuration**

- `tsconfig.json` sets `typeCheckHostBindings: false`. Angular 17+ enabled this strict check by default; several legacy `host` bindings in the library fail strict typing under v20+. Disabling preserves v15-v19 behavior for this patch release; removal is tracked as a follow-up structural cleanup.

**Public API**

- `clarity.api.md` / `testing.api.md` / `datagrid.api.md` / `datagrid-filters.api.md`: the only diff is the `ListRange` import being expressed as `cdkCollections.ListRange` (semantically identical type, different module path) to keep the API report stable across CDK majors. `totalContentHeight: string` is retained. `MockDatagridActionBarComponent` typings keep `unknown`.

- [ ] Yes
- [x] No

The public API surface is preserved (return type, mock typings, namespace-qualified `ListRange`). The template restructurings (`wizard.html`, `modal.html`) are semantically equivalent across versions and are only required for v21 projection resolution.

## Other information

VRT (visual regression) for `<clr-modal>` and `<clr-wizard>` should pass given the template restructurings are semantically equivalent.


dtsanevmw:
Preview project with affected components for validation - https://dainty-youtiao-e084fe.netlify.app/
Our demo app with Angular 21 and preview build from this branch - https://gilded-twilight-6e0f09.netlify.app/?cds-theme=light

projects/addons/test.ts
```diff
@@ -22,3 +22,20 @@ const testConfig: TestEnvironmentOptions = {
...
+// check so tests still pass for Angular 15–16 where the function does not exist.
+// eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-explicit-any, @typescript-eslint/no-var-requires
+const _ngCore: Record<string, any> = ngCore;
+const _provideZoneCD: (() => unknown) | undefined = _ngCore['provideZoneChangeDetection'];
+if (typeof _provideZoneCD === 'function') {
+  beforeEach(() => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    TestBed.configureTestingModule({ providers: [_provideZoneCD() as any] });
+  });
+}
```
valentin-mladenov, line 41: SInce this only affect Angular v21.


```suggestion
if (+ANGULAR_VERSION.major >= 21) {
  beforeEach(() => {
    TestBed.configureTestingModule({ providers: [ngCore.provideZoneChangeDetection] });
  });
}

```

projects/angular/src/data/datagrid/datagrid-virtual-scroll.directive.ts
```diff
@@ -150,8 +170,14 @@ export class ClrDatagridVirtualScrollDirective<T> implements AfterViewInit, DoCh
...
-  get totalContentHeight() {
-    return this.virtualScrollViewport?._totalContentHeight || '';
+  get totalContentHeight(): string {
+    // CDK 15-19: `_totalContentHeight` is a `string`.
+    // CDK 20+:   `_totalContentHeight` is a `Signal<string>`. Calling it returns the value.
+    const h = this.virtualScrollViewport?._totalContentHeight as unknown;
+    if (typeof h === 'function') {
+      return (h as () => string)() || '';
+    }
+    return (h as string) || '';
```
valentin-mladenov, line 180: Use noremal nameiong convention `height` or `totalHeight`


```suggestion
    const totalHeight = this.virtualScrollViewport?._totalContentHeight as unknown;
    if (typeof totalHeight === 'function') {
      return (totalHeight as () => string)() || '';
    }
    return (totalHeight as string) || '';
```

projects/angular/src/forms/file-input/file-input-container.spec.ts
```diff
@@ -18,6 +19,12 @@ import { ClrFileInputContainer } from './file-input-container';
 import { ClrFileInputValueAccessor } from './file-input-value-accessor';
 import { selectFiles } from './file-input.helpers';
 
+// provideZoneChangeDetection was introduced in Angular 17; guard with a runtime feature
+// check so tests still pass for Angular 15–16 where the function does not exist.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const _ngCore: Record<string, any> = ngCore;
+const _provideZoneCD: (() => unknown) | undefined = _ngCore['provideZoneChangeDetection'];
```
valentin-mladenov, line 26: Check my other cvomment for `provideZoneChangeDetection` in addons.

You can still store `provideZoneChangeDetection` in const based on angular version. 


```suggestion
// provideZoneChangeDetection was introduced in Angular 17; guard with a runtime feature
// check so tests still pass for Angular 15–16 where the function does not exist.
const _provideZoneCD = +ANGULAR_VERSION.major >17 ? ngCore.provideZoneChangeDetection : null;
```

projects/angular/package.json
```diff
@@ -18,9 +18,9 @@
     "url": "https://github.com/vmware-clarity/ng-clarity/issues"
   },
   "peerDependencies": {
-    "@angular/cdk": "15 || 16 || 17 || 18 || 19",
-    "@angular/common": "15 || 16 || 17 || 18 || 19",
-    "@angular/core": "15 || 16 || 17 || 18 || 19",
+    "@angular/cdk": "15 || 16 || 17 || 18 || 19 || 20 || 21",
+    "@angular/common": "15 || 16 || 17 || 18 || 19 || 20 || 21",
+    "@angular/core": "15 || 16 || 17 || 18 || 19 || 20 || 21",
```
valentin-mladenov, line 23:
```suggestion
    "@angular/cdk": "15 || 16 || 17 || 18 || 19 || 20 || >=21.1.0",
    "@angular/common": "15 || 16 || 17 || 18 || 19 || 20 || >=21.1.0",
    "@angular/core": "15 || 16 || 17 || 18 || 19 || 20 || >=21.1.0",
```

PR #2394, opened 2026-05-18
feat(date-picker): add support for custom first day of week

valentin-mladenov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-3070

- [ ] Yes
- [x] No


projects/angular/forms/datepicker/providers/locale-helper.service.ts
```diff
@@ -58,6 +58,24 @@ export class LocaleHelperService {
...
+  overrideFirstDayOfWeek(day: ClrWeekday | null): void {
+    if (day !== null && (day < ClrWeekday.Sunday || day > ClrWeekday.Saturday)) {
+      //
+      if (this._firstDayOfWeek !== ClrWeekday.Sunday) {
+        this._firstDayOfWeek = ClrWeekday.Sunday;
+        this.initializeLocaleDays();
+      }
+      return;
+    }
+    this._firstDayOfWeek = day === null ? ClrWeekday.Sunday : day;
```
dtsanevmw: When `day` is `null` we set `_firstDayOfWeek = Sunday`, but the doc above (and the container's input doc) says this reverts to the locale default. For `fr`/`pl`/etc. the locale default is Monday, so clearing the override silently switches them to Sunday.

Probably want to call `initializeFirstDayOfWeek()` here so we go back to the CLDR value.

projects/angular/forms/datepicker/providers/locale-helper.service.ts
```diff
@@ -58,6 +58,24 @@ export class LocaleHelperService {
     return this._localeDateFormat;
   }
 
+  /**
+   * Overrides the first day of the week regardless of locale.
+   * Accepts a `ClrWeekday` value (Sunday=0 through Saturday=6), or null to revert to locale default.
+   * Incorrect values will revert to default value (Sunday).
+   */
+  overrideFirstDayOfWeek(day: ClrWeekday | null): void {
+    if (day !== null && (day < ClrWeekday.Sunday || day > ClrWeekday.Saturday)) {
```
dtsanevmw: Coercing out-of-range values to Sunday feels surprising — if the consumer passes `7` by mistake and the locale was Monday, we'd silently flip them to Sunday. Could we just treat invalid input the same as `null` (revert to locale default), or no-op and warn in dev?

projects/angular/forms/datepicker/providers/locale-helper.service.ts
```diff
@@ -58,6 +58,24 @@ export class LocaleHelperService {
...
+      //
```
dtsanevmw: Leftover empty `//` comment.

projects/angular/forms/datepicker/interfaces/day-of-week.interface.ts
```diff
@@ -9,3 +9,17 @@ export interface ClrDayOfWeek {
...
+/**
+ * Index of the day of the week, matching JavaScript's Date.getDay() convention.
+ * Used to override the locale-derived first day of the week in the date picker.
+ */
+export enum ClrWeekday {
+  Sunday = 0,
+  Monday = 1,
+  Tuesday = 2,
+  Wednesday = 3,
+  Thursday = 4,
```
dtsanevmw: Nit: `ClrWeekday` is an enum, not an interface — would be cleaner to put it in its own file (`day-of-week.ts` or `enums/clr-weekday.enum.ts`).

dtsanevmw (commented):
Thanks for picking this up! Left some comments — most of them are around the override semantics and how this hooks into the calendar. Main concern is that clearing the override doesn't actually go back to the locale default for non-US locales.

projects/angular/forms/datepicker/providers/date-navigation.service.ts
```diff
@@ -184,6 +184,14 @@ export class DateNavigationService {
     this._focusOnCalendarChange.next();
   }
 
+  /**
+   * Forces a full calendar view regeneration without changing the displayed month.
+   * Used when layout-affecting settings (e.g. first day of week) change.
+   */
+  refreshDisplayedCalendar(): void {
```
dtsanevmw: I'd push back on re-using `_displayedCalendarChange` here. Its contract is "the displayed month/year changed" — everything else relies on `setDisplayedCalendar` to be the only emitter. Firing it for an unrelated reason (a layout setting change) blurs that contract.

Two cleaner options:
- Expose a `firstDayOfWeekChange` observable on `LocaleHelperService` and let `ClrCalendar` subscribe directly. Then nothing needs to change in this service.
- Add a dedicated subject here (e.g. `_calendarLayoutChange`) instead of piggy-backing on the month-change one.

The first one also removes the need for the container to orchestrate two services.

projects/angular/forms/datepicker/date-container.ts
```diff
@@ -147,6 +149,18 @@ export class ClrDateContainer extends ClrAbstractContainer implements AfterViewI
     }
   }
 
+  /**
+   * Overrides the locale-derived first day of the week for the calendar.
+   * Accepts a `ClrWeekday` value (Sunday=0 through Saturday=6).
+   * When not set, the first day of the week is determined by the Angular locale.
+   */
+  @Input('clrFirstDayOfWeek')
+  set firstDayOfWeek(value: ClrWeekday) {
```
dtsanevmw: The type should allow `ClrWeekday | null` — both the demo and the tests pass `null` through this input and also users should be able to clear the `custom` first day and revert to the locale one.

projects/angular/forms/datepicker/date-container.ts
```diff
@@ -147,6 +149,18 @@ export class ClrDateContainer extends ClrAbstractContainer implements AfterViewI
...
+    this.localeHelperService.overrideFirstDayOfWeek(value ?? null);
+
+    this.dateNavigationService.refreshDisplayedCalendar();
```
dtsanevmw: If `LocaleHelperService` emitted a change event (see comment on `date-navigation.service.ts`), the container wouldn't need to know about `refreshDisplayedCalendar` at all. Right now this setter has to coordinate two services and assumes a calendar is alive to consume the event, which is a bit fragile.

projects/angular/forms/datepicker/date-container.spec.ts
```diff
@@ -231,6 +232,91 @@ export default function () {
...
+    });
+
+    it('reverts to locale default when firstDayOfWeek is cleared', () => {
+      context.testComponent.firstDayOfWeek = ClrWeekday.Wednesday;
+      context.detectChanges();
+      expect(localeHelperService.firstDayOfWeek).toBe(ClrWeekday.Wednesday);
+
+      context.testComponent.firstDayOfWeek = null;
+      context.detectChanges();
+      expect(localeHelperService.firstDayOfWeek).toBe(ClrWeekday.Sunday);
```
dtsanevmw: This test only catches the broken case because `en-US` happens to default to Sunday. Could you add an equivalent assertion against a non-Sunday locale (e.g. `fr` → Monday) so the locale revert is actually verified? Right now the locale-helper bug above passes this spec.

projects/angular/forms/datepicker/date-container.spec.ts
```diff
@@ -231,6 +232,91 @@ export default function () {
...
+
+      context = this.create(ClrDateContainer, TestComponentWithFirstDay, DATEPICKER_PROVIDERS);
+      localeHelperService = context.getClarityProvider(LocaleHelperService);
+      dateNavigationService = context.getClarityProvider(DateNavigationService);
+    });
+
+    afterEach(() => {
+      const viewManager = document.querySelector('clr-datepicker-view-manager');
+      if (viewManager) {
+        viewManager.remove();
```
dtsanevmw: Reaching into `document` to clean up the view-manager looks like a workaround for TestBed teardown. None of the new tests look at the calendar DOM — do we actually need the popover open here? If not, we can drop this `afterEach`.

projects/angular/forms/datepicker/date-container.spec.ts
```diff
@@ -231,6 +232,91 @@ export default function () {
...
+      context.detectChanges();
+      expect(dateNavigationService.refreshDisplayedCalendar).not.toHaveBeenCalled();
+    });
+
+    it('supports setting first day of week to ClrWeekday.Saturday', () => {
+      context.testComponent.firstDayOfWeek = ClrWeekday.Saturday;
+      context.detectChanges();
+      expect(localeHelperService.firstDayOfWeek).toBe(ClrWeekday.Saturday);
+      expect(localeHelperService.localeDaysNarrow[0]).toBe('S');
+      expect(localeHelperService.localeDaysNarrow[1]).toBe('S');
```
dtsanevmw: Both narrow letters are `S` for Sunday and Saturday in en-US, so this would still pass with the wrong value. Maybe assert against `localeDays[0].day === 'Saturday'` or check `localeDaysNarrow[6]` so it's unambiguous.

valentin-mladenov: Corrected in commit b2d93f6

projects/angular/forms/datepicker/date-container.spec.ts
valentin-mladenov: Corrected in commit b2d93f6

projects/angular/forms/datepicker/date-container.spec.ts
valentin-mladenov: Corrected in commit b2d93f6

projects/angular/forms/datepicker/date-container.ts
valentin-mladenov: Corrected in commit b2d93f6

projects/angular/forms/datepicker/providers/date-navigation.service.ts
valentin-mladenov: Corrected in commit b2d93f6

projects/angular/forms/datepicker/interfaces/day-of-week.interface.ts
valentin-mladenov: Corrected in commit b2d93f6

projects/angular/forms/datepicker/providers/locale-helper.service.ts
valentin-mladenov: Corrected in commit b2d93f6

projects/angular/forms/datepicker/providers/locale-helper.service.ts
valentin-mladenov: Corrected in commit b2d93f6

projects/angular/forms/datepicker/providers/locale-helper.service.ts
valentin-mladenov: Corrected in commit b2d93f6

projects/angular/forms/datepicker/date-container.ts
valentin-mladenov: Corrected in commit f0d312e

projects/website/src/app/documentation/demos/datepicker/demos/datepicker-api.demo.html
```diff
@@ -114,3 +114,44 @@ <h5 cds-text="subsection" class="clr-mt-24px">Confirmation Action Buttons</h5>
...
+    <input
+      clrNumberInput
+      type="number"
+      min="0"
+      max="6"
+      [value]="firstDayOfWeek ?? ''"
+      (input)="onFirstDayChange($any($event.target).value)"
+      placeholder="Locale default"
+      style="min-width: 150px"
+    />
```
dtsanevmw: Consider using `Select` component with labels Monday-Sunday and their values being 0-6, it will be easier for users playing in website to understand what they are choosing, number input is harder choice.

projects/angular/forms/datepicker/providers/locale-helper.service.ts
```diff
@@ -58,12 +59,30 @@ export class LocaleHelperService {
     return this._localeDateFormat;
   }
 
+  /**
+   * Overrides the first day of the week regardless of locale.
+   * Accepts a `ClrWeekday` value (Sunday=0 through Saturday=6), or null to revert to locale default.
+   * Incorrect values will revert to default value (Sunday).
+   */
+  overrideFirstDayOfWeek(day: ClrWeekday | null): void {
```
dtsanevmw: Last suggestion, `override` sounds strange to me for this function as if this is supported functionality now, it's simply updating the value, consider different name that fits more and doesn't feel like you are doing something unsupported?
```suggestion
  updateFirstDayOfWeek(day: ClrWeekday | null): void {
```

dtsanevmw (approved):

PR #2439, opened 2026-06-05
chore(figma): tokens single source of truth

valentin-mladenov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [x] CI related changes

Issue Number: CDE-3113

- [ ] Yes
- [x] No


.github/workflows/figma-publish.yml
```diff
@@ -0,0 +1,109 @@
...
+          |---|---|
+          | Branch | `${{ github.ref_name }}` |
+          | Production push | `${{ steps.branch.outputs.is_main == 'true' && 'yes' || 'no (isolated)' }}` |
+          | Isolation mode | `${{ inputs.branch_isolation || 'collection' }}` |
+          | Dry run | `${{ inputs.dry_run || false }}` |
+          | Triggered by | `${{ github.event_name }}` |
+
+          Design tokens from `dist/clr-ui/clr-ui.css` were published to the
+          [Figma file](https://www.figma.com/file/${{ secrets.FIGMA_FILE_KEY }}).
+          EOF
```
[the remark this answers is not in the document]
valentin-mladenov: Resolved in commit 688ea54

.env.figma
dtsanevmw, line 1: Do we need a second .env.figma file when we already have a .env file where these can be added?

valentin-mladenov: Not really. Creating `.env.figma` file was more like isolation practice.

dtsanevmw: So we can safely remove it and if someone is working locally can just add these to their `.env` file?

figma-tokens.config.json
```diff
@@ -0,0 +1,269 @@
...
+          "--cds-alias-object-opacity-",
+          "--cds-alias-typography-color-",
+          "--cds-alias-typography-link-",
+          "--cds-alias-typography-info-",
+          "--cds-alias-typography-success-",
+          "--cds-alias-typography-warning-",
+          "--cds-alias-typography-danger-",
+          "--cds-alias-typography-neutral-",
+          "--cds-alias-typography-disabled"
+        ]
```
dtsanevmw: If these are going to be always the same as the one above, could the be re-used somehow instead of manually keeping them in sync?

figma/figma-tokens.config.json
```diff
@@ -0,0 +1,269 @@
...
+    {
+      "_comment": "Border radius tokens",
+      "pattern": "*border-radius*",
+      "scopes": ["CORNER_RADIUS"]
+    },
+    {
+      "_comment": "Border radius short name (clr-global-borderradius, etc.)",
+      "pattern": "*borderradius*",
+      "scopes": ["CORNER_RADIUS"]
+    },
```
dtsanevmw, line 239: Base on the ones above, would this work too?
```suggestion
      "_comment": "Border radius tokens",
      "pattern": "*(border-radius|borderradius)*",
      "scopes": ["CORNER_RADIUS"]
    },
```

figma-tokens.config.json
```diff
@@ -0,0 +1,269 @@
...
+    {
+      "_comment": "Stroke / border color tokens",
+      "pattern": "*-border-color*",
+      "scopes": ["STROKE_COLOR"]
+    },
+    {
+      "_comment": "Stroke / border color tokens",
+      "pattern": "*-viz-border*",
+      "scopes": ["STROKE_COLOR"]
+    },
```
dtsanevmw, line 264: Same simplification as above
```suggestion
      "_comment": "Stroke / border color tokens",
      "pattern": "*(-border-color|-viz-border)*",
      "scopes": ["STROKE_COLOR"]
    },
```

figma-tokens.config.json
```diff
@@ -0,0 +1,269 @@
...
+    {
+      "_comment": "Border-specific token names",
+      "pattern": "*-border-width*",
+      "scopes": ["STROKE_FLOAT"]
+    },
+    {
+      "_comment": "Border-specific token names",
+      "pattern": "*-stroke*",
+      "scopes": ["STROKE_FLOAT"]
+    },
```
dtsanevmw, line 274:
```suggestion
      "_comment": "Border-specific token names",
      "pattern": "*(-border-width|-stroke)*",
      "patterns":  ["*-border-width*", "*-stroke*"] // Alternative
      "scopes": ["STROKE_FLOAT"]
    },
```

figma/figma-tokens.config.schema.json
```diff
@@ -0,0 +1,134 @@
...
+    "scopeRule": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["pattern", "scopes"],
+      "properties": {
+        "_comment": { "type": "string" },
+        "pattern": {
+          "type": "string",
+          "description": "Glob-like pattern matched against token names (* → .*, ? → .). Matched case-insensitively."
+        },
```
dtsanevmw, line 155: It would be way easier to define one scope with multiple patterns that apply to it, instead of creating a separate object for each of them with the same name
```suggestion
       "patterns": {
          "type": "array",
          "description": "Glob-like pattern matched against token names (* → .*, ? → .). Matched case-insensitively."
          "items": { "type": "string" }
        },
```

scripts/figma/index.mjs
```diff
@@ -0,0 +1,202 @@
...
+
+  // ── Fetch existing Figma variables ──────────────────────────────────────────
+  let existingCollections = [];
+  let existingVars = [];
+  let existingModes = [];
+  let effectiveFileKey = figmaFileKey;
+
+  if (!dryRun && !extractMode) {
+    console.log('⬇️   Fetching existing Figma variables…');
+    const existing = await figma.get(`/files/${figmaFileKey}/variables/local`);
```
dtsanevmw: Would it be better to abstract these in the client and have less endpoint writing here?
```suggestion
    const existing = await figma.getVariables();
```

scripts/figma/index.mjs
```diff
@@ -0,0 +1,202 @@
...
+    ({ collections: existingCollections, vars: existingVars, modes: existingModes } = parseFigmaVarsResponse(existing));
+  }
+
+  // ── Branch isolation ────────────────────────────────────────────────────────
+  let collectionSuffix = '';
+  if (branchName) {
+    if (figmaBranchMode === 'branch' && !dryRun) {
+      // Target the Figma branch file key
+      const branches = await figma.get(`/files/${figmaFileKey}/branches`);
```
dtsanevmw:
```suggestion
      const branches = await figma.getBranches();
```

scripts/figma/index.mjs
```diff
@@ -0,0 +1,202 @@
...
+const CONFIG_PATH = path.join(ROOT, 'figma-tokens.config.json');
+const CSS_FILE = path.join(ROOT, 'dist', 'clr-ui', 'clr-ui.css');
+
+async function main() {
+  // ── CLI + environment ─────────────────────────────────────────────────────
+  const { dryRun, extractMode, extractFile, branchName } = parseCliArgs();
+  const { figmaToken, figmaFileKey, figmaBranchMode } = loadEnv({ extractMode });
+  const config = loadConfig(CONFIG_PATH);
+  const rules = createTokenRules(config);
+  const figma = createFigmaClient(figmaToken);
```
dtsanevmw:
```suggestion
  const figma = createFigmaClient(figmaToken, figmaFileKey);
```

scripts/figma/index.mjs
```diff
@@ -0,0 +1,202 @@
...
+    } else {
+      collectionSuffix = ` [${branchName}]`;
+    }
+  }
+
+  // ── Parse CSS ──────────────────────────────────────────────────────────────
+  if (!fs.existsSync(CSS_FILE)) {
+    console.error(`❌  ${CSS_FILE} not found. Run: npm run _build:ui`);
+    process.exit(1);
+  }
```
dtsanevmw: I think we should put this into the top, before doing anything, we don't have to waste time for the functions above if we don't have the required step ready.

scripts/figma/index.mjs
```diff
@@ -0,0 +1,202 @@
...
+
+  // ── Dry-run / extract: build full plan and report, then exit ───────────────
+  if (dryRun || extractMode) {
+    const idMap = createIdMap();
+    const plan = buildPushPlan({
+      collectionDefs,
+      collectionSuffix,
+      existingCollections,
+      existingModes,
+      existingVars,
```
dtsanevmw: How we load all these if they are received only when `if (!dryRun && !extractMode) {`

scripts/figma/setup/css-parser.mjs
```diff
@@ -0,0 +1,150 @@
...
+
+// Canonical selectors we care about.
+// The "bare" DARK / COMPACT forms (without :where(:root,:host) prefix) are the
+// primary sources for mode overrides in the compiled Clarity stylesheet.
+export const SELECTORS = {
+  ROOT: ':where(:root, :host)',
+  DARK: '[cds-theme~=dark]',
+  COMPACT: '[clr-density=compact]',
+  BOTH_THEME: ':where(:root, :host), :where(:root, :host) [cds-theme]',
+  BOTH_DENSITY: ':where(:root, :host), :where(:root, :host) [clr-density]',
```
dtsanevmw:
```suggestion
  BOTH_THEME: '${SELECTORS.ROOT}, ${SELECTORS.ROOT} [cds-theme]',
  BOTH_DENSITY: '${SELECTORS.ROOT}, ${SELECTORS.ROOT} [clr-density]',
```

scripts/figma/setup/config.mjs
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Configuration loading for the Figma token publisher.
+ *
+ * Reads figma-tokens.config.json and derives the normalized lookup structures
+ * (exclusions, scope rules, code-syntax templates) consumed by the token rules.
+ */
+
+import fs from 'node:fs';
+
+const DEFAULT_CODE_SYNTAX = { WEB: 'var(${name})', ANDROID: '@clr/${kebab}', iOS: 'Clarity.${camel}' };
```
dtsanevmw: We don't ened the others
```suggestion
const DEFAULT_CODE_SYNTAX ='var(${name})';
```

scripts/figma/api/diff-printer.mjs
```diff
@@ -0,0 +1,109 @@
...
+ * Build the human-readable diff string (see {@link printDiff}) without emitting it.
+ * Kept side-effect-free so the formatting can be unit-tested.
+ *
+ * @param {Array<{ collectionName: string, diff: import('./planner.mjs').DiffEntry[] }>} diffReport
+ * @returns {string}
+ */
+export function formatDiff(diffReport) {
+  const lines = ['\n📋  Diff:\n'];
+
+  let anyChanges = false;
```
dtsanevmw:
```suggestion
  const hasChanges = diffReport.some(report => report.diff?.length);
  if (!hasChanges) {
    return '\n📋  Diff:\n\n  No changes.\n';
  }
  const lines = ['\n📋  Diff:\n'];
```

figma/api/diff-printer.mjs
```diff
@@ -0,0 +1,109 @@
...
+  let anyChanges = false;
+
+  for (const { collectionName, diff } of diffReport) {
+    const retyped = diff.filter(e => e.action === 'retype');
+    const created = diff.filter(e => e.action === 'create');
+    const updated = diff.filter(e => e.action === 'update');
+
+    if (diff.length === 0) {
+      continue;
+    }
```
dtsanevmw, line 76:
```suggestion
```

figma/api/diff-printer.mjs
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @param {Array<{ collectionName: string, diff: import('./planner.mjs').DiffEntry[] }>} diffReport
+ * @returns {string}
+ */
+export function formatDiff(diffReport) {
+  const lines = ['\n📋  Diff:\n'];
+
+  let anyChanges = false;
+
+  for (const { collectionName, diff } of diffReport) {
```
dtsanevmw, line 73: Move to the beginning, no need to filter if empty.
```suggestion
  for (const { collectionName, diff } of diffReport) {
    if (diff.length === 0) {
      continue;
    }
```

scripts/figma/api/diff-printer.mjs
```diff
@@ -0,0 +1,109 @@
...
+    for (const e of created) {
+      lines.push(`    +  ${e.name.padEnd(60)}  ${e.type}`);
+    }
+
+    for (const e of updated) {
+      lines.push(`    ~  ${e.name.padEnd(60)}  ${e.type}`);
+      for (const { field, from, to } of e.changes ?? []) {
+        lines.push(`         ${field.padEnd(22)}  ${from}  →  ${to}`);
+      }
+    }
```
dtsanevmw: Re-usable helper for the formatting on each type
```suggestion
const addLine = (symbol, e, suffix) => {
      lines.push(`    ${symbol}  ${e.name.padEnd(60)}  ${suffix}`);
      
      e.changes?.forEach(c => {
        lines.push(`         ${c.field.padEnd(22)}  ${c.from}  →  ${c.to}`);
      });
    };

    retyped.forEach(e => addLine('↻', e, `${e.prevType} → ${e.type}`));
    created.forEach(e => addLine('+', e, e.type));
    updated.forEach(e => addLine('~', e, e.type));
```

scripts/figma/api/extract-view.mjs
```diff
@@ -0,0 +1,102 @@
...
+
+  // Build a human-readable per-collection view alongside the raw Figma payload
+  const modeIdToName = new Map();
+  for (const m of payloadModes) {
+    modeIdToName.set(m.id, m.name);
+  }
+  // For existing modes (not in payloadModes) we need a lookup from existingModes
+  for (const em of existingModes) {
+    modeIdToName.set(em.modeId, em.name);
+  }
```
dtsanevmw: These are not used, can be removed?

figma/api/extract-view.mjs
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright (c) 2016-2026 Broadcom. All Rights Reserved.
+ * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
+
+import { figmaNameToCssName } from '../util/naming.mjs';
+
+/**
```
dtsanevmw, line 16: Create a constant for that prefix as it's used in more than one place and easily could break.
```suggestion
import { figmaNameToCssName } from '../util/naming.mjs';

/**
 * Description prefix written by planner.mjs for human-readable alias variables.
 * Kept here as a constant so a format change in planner.mjs produces an obvious
 * compile-time (or grep-time) failure rather than a silent wrong cssVar output.
 */
export const ALIAS_DESCRIPTION_PREFIX = 'Alias of: ';

/**
```

scripts/figma/api/extract-view.mjs
```diff
@@ -0,0 +1,102 @@
...
+        }
+        // Human-readable alias variables store the source CSS var in the description
+        // ("Alias of: --cds-…"); regular variables derive it from the Figma path.
+        const cssName = v.description?.startsWith('Alias of: ')
+          ? v.description.slice('Alias of: '.length)
+          : figmaNameToCssName(v.name);
+        return {
+          id: v.id,
+          figmaPath: v.name,
+          cssVar: cssName,
```
dtsanevmw:
```suggestion
          cssVar,
```

scripts/figma/api/extract-view.mjs
```diff
@@ -0,0 +1,102 @@
...
+          const mv = varModeValues.find(mv => mv.modeId === mode.id);
+          if (mv) {
+            valuesByMode[mode.name] = mv.value;
+          }
+        }
+        // Human-readable alias variables store the source CSS var in the description
+        // ("Alias of: --cds-…"); regular variables derive it from the Figma path.
+        const cssName = v.description?.startsWith('Alias of: ')
+          ? v.description.slice('Alias of: '.length)
+          : figmaNameToCssName(v.name);
```
dtsanevmw:
```suggestion
      const cssVar = v.description?.startsWith(ALIAS_DESCRIPTION_PREFIX)
        ? v.description.slice(ALIAS_DESCRIPTION_PREFIX.length)
        : figmaNameToCssName(v.name);
```

scripts/figma/api/extract-view.mjs
```diff
@@ -0,0 +1,102 @@
...
+  // For existing modes (not in payloadModes) we need a lookup from existingModes
+  for (const em of existingModes) {
+    modeIdToName.set(em.modeId, em.name);
+  }
+
+  const collectionView = collectionDefs.map(colDef => {
+    const existingCol = existingCollByName.get(colDef.name + collectionSuffix);
+    const colId = existingCol
+      ? existingCol.id
+      : (payloadCollections.find(c => c.name === colDef.name + collectionSuffix)?.id ?? '?');
```
dtsanevmw: Create and re-use colName.
```suggestion
    const colName = colDef.name + collectionSuffix;
    const existingCol = existingCollByName.get(colName);
    const colId = existingCol
      ? existingCol.id
      : (payloadCollections.find(c => c.name === colName)?.id ?? '?');
```

scripts/figma/api/extract-view.mjs
```diff
@@ -0,0 +1,102 @@
...
+  // Build a human-readable per-collection view alongside the raw Figma payload
+  const modeIdToName = new Map();
+  for (const m of payloadModes) {
+    modeIdToName.set(m.id, m.name);
+  }
+  // For existing modes (not in payloadModes) we need a lookup from existingModes
+  for (const em of existingModes) {
+    modeIdToName.set(em.modeId, em.name);
+  }
+
```
dtsanevmw, line 45:
```suggestion
  // Pre-index all lookups so the per-collection/per-variable loops below are O(1)
  // instead of O(n) scans repeated across every iteration.
  const payloadColByName = new Map(payloadCollections.map(c => [c.name, c]));
  // Composite key: "<collectionId>:<modeName>" → payload mode entry
  const payloadModeByKey = new Map(payloadModes.map(m => [`${m.variableCollectionId}:${m.name}`, m]));
  // Composite key: "<collectionId>:<modeName>" → existing mode entry
  const existingModeByKey = new Map(existingModes.map(m => [`${m.collectionId}:${m.name}`, m]));
  // Group non-deleted payload vars by their collection ID
  const varsByColId = new Map();
  for (const v of payloadVars) {
    if (v.action === 'DELETE') continue;
    const bucket = varsByColId.get(v.variableCollectionId);
    if (bucket) {
      bucket.push(v);
    } else {
      varsByColId.set(v.variableCollectionId, [v]);
    }
    
      const modeValuesByVarId = new Map();
  for (const mv of payloadModeValues) {
    const bucket = modeValuesByVarId.get(mv.variableId);
    if (bucket) {
      bucket.push(mv);
    } else {
      modeValuesByVarId.set(mv.variableId, [mv]);
    }
  }

```

scripts/figma/api/extract-view.mjs
```diff
@@ -0,0 +1,102 @@
...
+    const existingCol = existingCollByName.get(colDef.name + collectionSuffix);
+    const colId = existingCol
+      ? existingCol.id
+      : (payloadCollections.find(c => c.name === colDef.name + collectionSuffix)?.id ?? '?');
+
+    // Resolve mode IDs once at the collection level so variable loops can reuse them.
+    const modes = colDef.modes.map(modeName => {
+      const modeEntry = payloadModes.find(m => m.variableCollectionId === colId && m.name === modeName);
+      const existingModeObj = existingModes.find(m => m.collectionId === colId && m.name === modeName);
+      return { id: modeEntry?.id ?? existingModeObj?.modeId ?? null, name: modeName };
```
dtsanevmw:
```suggestion
      const key = `${colId}:${modeName}`;
      const id = payloadModeByKey.get(key)?.id ?? existingModeByKey.get(key)?.modeId ?? null;
      return { id, name: modeName };

```

scripts/figma/api/extract-view.mjs
```diff
@@ -0,0 +1,102 @@
...
+      .filter(v => v.variableCollectionId === colId && v.action !== 'DELETE')
+      .map(v => {
+        const varModeValues = payloadModeValues.filter(mv => mv.variableId === v.id);
+        const valuesByMode = {};
+        for (const mode of modes) {
+          const mv = varModeValues.find(mv => mv.modeId === mode.id);
+          if (mv) {
+            valuesByMode[mode.name] = mv.value;
+          }
+        }
```
dtsanevmw:
```suggestion
   const modeById = new Map(modes.map(m => [m.id, m.name]));
   const variables = (varsByColId.get(colId) ?? []).map(v => {
      const values = {};
      for (const mv of modeValuesByVarId.get(v.id) ?? []) {
        const modeName = modeById.get(mv.modeId);
        if (modeName) values[modeName] = mv.value;
      }
```

scripts/figma/api/extract-view.mjs
```diff
@@ -0,0 +1,102 @@
...
+          scopes: v.scopes,
+          codeSyntax: v.codeSyntax,
+          description: v.description,
+          values: valuesByMode,
+        };
+      });
+
+    return {
+      id: colId,
+      name: colDef.name + collectionSuffix,
```
dtsanevmw:
```suggestion
      name: colName,
```

figma/api/push-executor.mjs
```diff
@@ -0,0 +1,186 @@
...
+  effectiveFileKey,
+  collectionDefs,
+  collectionSuffix,
+  existingCollections,
+  existingVars,
+  existingModes,
+  rules,
+  varLookup,
+}) {
+  const idMap = createIdMap();
```
dtsanevmw, line 65:
```suggestion
  const varsUrl = `/files/${effectiveFileKey}/variables`;
  const localVarsUrl = `${varsUrl}/local`;
  
  const idMap = createIdMap();
```

scripts/figma/api/push-executor.mjs
```diff
@@ -0,0 +1,186 @@
...
+  existingModes,
+  rules,
+  varLookup,
+}) {
+  const idMap = createIdMap();
+  populateIdMapFromExisting(existingVars, idMap);
+
+  // Mutable lookup Maps rebuilt from each GET response so subsequent
+  // collections always see the latest authoritative Figma state.
+  const { collByName: existingCollByName, varByName: existingVarByName } = buildLookupMaps(
```
dtsanevmw:
```suggestion
  let { collByName: existingCollByName, varByName: existingVarByName } = buildLookupMaps(
```

scripts/figma/api/push-executor.mjs
```diff
@@ -0,0 +1,186 @@
...
+    existingCollections,
+    existingVars
+  );
+
```
dtsanevmw:
```suggestion
  );

  // Local reference — never mutates the caller's array.
  let currentModes = existingModes;

```

scripts/figma/api/push-executor.mjs
```diff
@@ -0,0 +1,186 @@
...
+    totalSkipped += colPlan.stats.skipped;
+    totalDeleted += colPlan.deletedVarIds.size;
+    totalModeValues += colPlan.payloadModeValues.length;
+    diffReport.push({ collectionName: colDef.name + collectionSuffix, diff: colPlan.diff });
+
+    // ── Type-mismatch deletions (must precede recreations) ──────────────────
+    const deleteVars = colPlan.payloadVars.filter(v => v.action === 'DELETE');
+    if (deleteVars.length > 0) {
+      console.log(`    🗑️   Deleting ${deleteVars.length} type-mismatched variable(s)…`);
+      await figma.post(`/files/${effectiveFileKey}/variables`, {
```
dtsanevmw:
```suggestion
      await figma.post(varsUrl, {
```

scripts/figma/api/push-executor.mjs
```diff
@@ -0,0 +1,186 @@
...
+
+    totalNew += colPlan.stats.new;
+    totalUpdate += colPlan.stats.update;
+    totalSkipped += colPlan.stats.skipped;
+    totalDeleted += colPlan.deletedVarIds.size;
+    totalModeValues += colPlan.payloadModeValues.length;
+    diffReport.push({ collectionName: colDef.name + collectionSuffix, diff: colPlan.diff });
+
+    // ── Type-mismatch deletions (must precede recreations) ──────────────────
+    const deleteVars = colPlan.payloadVars.filter(v => v.action === 'DELETE');
```
dtsanevmw:
```suggestion
  // Partition in one pass — deletes and create/updates are complementary sets
    const deleteVars = [];
    const createUpdateVars = [];
    for (const v of colPlan.payloadVars) {
      (v.action === 'DELETE' ? deleteVars : createUpdateVars).push(v);
    }
```

scripts/figma/api/push-executor.mjs
```diff
@@ -0,0 +1,186 @@
...
+        variables: deleteVars,
+        variableModeValues: [],
+      });
+    }
+
+    // ── Push this collection in batches ─────────────────────────────────────
+    // Each collection is kept in a single POST whenever possible (< BATCH vars).
+    // If a collection exceeds BATCH variables it is split, but the collection
+    // and modes are always included in the first batch so Figma can resolve IDs.
+    const createUpdateVars = colPlan.payloadVars.filter(v => v.action !== 'DELETE');
```
dtsanevmw:
```suggestion
```

scripts/figma/api/push-executor.mjs
```diff
@@ -0,0 +1,186 @@
...
+        variableModeValues: sliceMV,
+      });
+    }
+
+    // ── Refresh idMap with real Figma IDs ───────────────────────────────────
+    // Any temp IDs assigned during this collection's plan are now resolved to
+    // real Figma IDs.  The next collection's VARIABLE_ALIAS mode values will
+    // therefore carry real IDs rather than stale temp IDs.
+    console.log(`    🔄  Syncing variable IDs…`);
+    const fresh = await figma.get(`/files/${effectiveFileKey}/variables/local`);
```
dtsanevmw:
```suggestion
    const fresh = await figma.get(localVarsUrl);
```

scripts/figma/api/push-executor.mjs
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  // ── Remove Figma's auto-created "Mode 1" from every new collection ─────────
+  // Figma silently appends a default "Mode 1" whenever a collection is created,
+  // even when we supply our own named modes in the same request.  Detect and
+  // delete every such stale mode now that all collections are live.
+  const defaultModes = existingModes.filter(m => m.name === 'Mode 1');
+  if (defaultModes.length > 0) {
+    console.log(`\n  🧹  Removing ${defaultModes.length} auto-created "Mode 1" mode(s)…`);
+    await figma.post(`/files/${effectiveFileKey}/variables`, {
```
dtsanevmw:
```suggestion
    await figma.post(varsUrl, {
```

scripts/figma/api/push-executor.mjs
```diff
@@ -0,0 +1,186 @@
...
+
+    existingModes.length = 0;
+    existingModes.push(...freshModes);
+  }
+
+  // ── Remove Figma's auto-created "Mode 1" from every new collection ─────────
+  // Figma silently appends a default "Mode 1" whenever a collection is created,
+  // even when we supply our own named modes in the same request.  Detect and
+  // delete every such stale mode now that all collections are live.
+  const defaultModes = existingModes.filter(m => m.name === 'Mode 1');
```
dtsanevmw:
```suggestion
  const defaultModes = currentModes.filter(m => m.name === 'Mode 1');
```

scripts/figma/api/push-executor.mjs
```diff
@@ -0,0 +1,186 @@
...
+      existingCollByName.set(c.name, c);
+    }
+
+    existingVarByName.clear();
+    for (const v of freshVars) {
+      existingVarByName.set(v.name, v);
+    }
+
+    existingModes.length = 0;
+    existingModes.push(...freshModes);
```
dtsanevmw:
```suggestion
   ({ collByName: existingCollByName, varByName: existingVarByName } = buildLookupMaps(freshCollections, freshVars));
    currentModes = freshModes;```

File: scripts/figma/api/push-executor.mjs [hunk 37]
```diff
@@ -0,0 +1,186 @@
...
+    // ── Push this collection in batches ─────────────────────────────────────
+    // Each collection is kept in a single POST whenever possible (< BATCH vars).
+    // If a collection exceeds BATCH variables it is split, but the collection
+    // and modes are always included in the first batch so Figma can resolve IDs.
+    const createUpdateVars = colPlan.payloadVars.filter(v => v.action !== 'DELETE');
+
+    for (let i = 0; i < Math.max(createUpdateVars.length, 1); i += BATCH) {
+      const slice = createUpdateVars.slice(i, i + BATCH);
+      const sliceIds = new Set(slice.map(v => v.id));
+      const sliceMV = colPlan.payloadModeValues.filter(mv => sliceIds.has(mv.variableId));
```
dtsanevmw:
```suggestion
    const modeValuesByVarId = new Map();
    for (const mv of colPlan.payloadModeValues) {
      const bucket = modeValuesByVarId.get(mv.variableId);
      if (bucket) bucket.push(mv);
      else modeValuesByVarId.set(mv.variableId, [mv]);
    }
    const batchCount = Math.ceil(Math.max(createUpdateVars.length, 1) / BATCH);
    for (let i = 0; i < batchCount; i += BATCH) {
      const slice = createUpdateVars.slice(i, i + BATCH);
      const sliceMV = slice.flatMap(v => modeValuesByVarId.get(v.id) ?? []);
```

scripts/figma/api/push-executor.mjs
```diff
@@ -0,0 +1,186 @@
...
+      console.log(
+        `    📤  Batch ${Math.floor(i / BATCH) + 1}: ${slice.length} var(s), ${sliceMV.length} mode value(s)`
+      );
+      await figma.post(`/files/${effectiveFileKey}/variables`, {
```
dtsanevmw:
```suggestion
      await figma.post(varsUrl, {
```

scripts/figma/core/collections.mjs
```diff
@@ -0,0 +1,73 @@
...
+ *   Present only on humanReadable collections. Maps CSS variable name → display name.
+ *   When present the planner iterates this map instead of the CSS source.
+ */
+
+/** @type {Record<import('./config.mjs').CssModeSource, keyof ModeVars>} */
+const SOURCE_KEY = { root: 'rootVars', dark: 'darkVars', compact: 'compactVars' };
+
+/**
+ * @typedef {{ rootVars: Map<string, string>, darkVars: Map<string, string>, compactVars: Map<string, string> }} ModeVars
+ */
```
dtsanevmw:
```suggestion
```

scripts/figma/core/collections.mjs
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * Each collection's `filter` uses prefix matching:
+ *   - A token is included if it starts with any entry in `filter.include` (OR).
+ *   - A token is excluded if it starts with any entry in `filter.exclude` (OR).
+ *
+ * Collections marked `humanReadable: true` in the config get a `humanReadableEntries`
+ * map attached to their def. Their filter always returns false (no CSS tokens are
+ * scanned directly); the planner populates them from the entries map instead.
+ *
+ * Each mode's `source` key maps to the matching parsed CSS variable map.
```
dtsanevmw:
```suggestion
 * Each mode's `source` string ('root' | 'dark' | 'compact') is used directly as a
 * key into the {@link import('../setup/css-parser.mjs').ModeVars} object returned by
 * `resolveModeVars`, so no translation table is needed.
```

scripts/figma/core/collections.mjs
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @param {import('./config.mjs').CollectionConfig[]} collectionConfigs
+ * @param {ModeVars} modeVars
```
dtsanevmw:
```suggestion
 * @param {import('../setup/css-parser.mjs').ModeVars} modeVars
```

scripts/figma/core/collections.mjs
```diff
@@ -0,0 +1,73 @@
...
+ */
+export function buildCollectionDefs(collectionConfigs, modeVars, humanReadableMap = {}) {
+  return collectionConfigs.map(({ name, filter, modes, humanReadable }) => {
+    /** @type {CollectionDef} */
+    const def = {
+      name,
+      filter: cssName =>
+        filter.include.some(p => cssName.startsWith(p)) && !filter.exclude.some(p => cssName.startsWith(p)),
+      modes: modes.map(m => m.name),
+      source: idx => modeVars[SOURCE_KEY[modes[idx].source]],
```
dtsanevmw:
```suggestion
      source: idx => modeVars[modes[idx].source],
```

scripts/figma/index.mjs
```diff
@@ -0,0 +1,202 @@
...
+
+  // ── Parse CSS ──────────────────────────────────────────────────────────────
+  if (!fs.existsSync(CSS_FILE)) {
+    console.error(`❌  ${CSS_FILE} not found. Run: npm run _build:ui`);
+    process.exit(1);
+  }
+  const cssText = fs.readFileSync(CSS_FILE, 'utf8');
+  const modeVars = resolveModeVars(parseCssBlocks(cssText));
+  const collectionDefs = buildCollectionDefs(config.collections, modeVars, config.humanReadable);
+
```
dtsanevmw:
```suggestion
  const varLookup = name => modeVars.root.get(name);
  const hrCount = Object.keys(config.humanReadable).length;

```

scripts/figma/index.mjs
```diff
@@ -0,0 +1,202 @@
...
+    const idMap = createIdMap();
+    const plan = buildPushPlan({
+      collectionDefs,
+      collectionSuffix,
+      existingCollections,
+      existingModes,
+      existingVars,
+      idMap,
+      rules,
+      varLookup: name => modeVars.rootVars.get(name),
```
dtsanevmw:
```suggestion
      varLookup,
```

scripts/figma/index.mjs
```diff
@@ -0,0 +1,202 @@
...
+    });
+
+    const { payloadModeValues, deletedVarIds, stats } = plan;
+    const hrCount = Object.keys(config.humanReadable).length;
```
dtsanevmw:
```suggestion
```

scripts/figma/index.mjs
```diff
@@ -0,0 +1,202 @@
...
+  const { totalNew, totalUpdate, totalSkipped, totalDeleted, totalModeValues, diffReport } = await executePush({
+    figma,
+    effectiveFileKey,
+    collectionDefs,
+    collectionSuffix,
+    existingCollections,
+    existingVars,
+    existingModes,
+    rules,
+    varLookup: name => modeVars.rootVars.get(name),
```
dtsanevmw:
```suggestion
    varLookup,
```

scripts/figma/index.mjs
```diff
@@ -0,0 +1,202 @@
...
+  });
+
+  // ── Diff ───────────────────────────────────────────────────────────────────
+  printDiff(diffReport);
+
+  // ── Summary ────────────────────────────────────────────────────────────────
+  const hrCount = Object.keys(config.humanReadable).length;
```
dtsanevmw:
```suggestion
```

figma/setup/css-parser.mjs
```diff
@@ -0,0 +1,150 @@
...
+    // Multi-line selector accumulation (comma-separated selectors across lines)
+    if (depth === 0 && line && !line.startsWith('@') && !line.startsWith('}')) {
+      selectorBuffer += (selectorBuffer ? ' ' : '') + line;
+    }
+  }
+
+  return blocks;
+}
+
+/**
```
dtsanevmw, line 80:
```suggestion
/**
 * Parsed CSS variable maps, keyed by mode name.
 * Keys match the `CssModeSource` union used in `config.mjs` and `figma-tokens.config.json`.
 *
 * @typedef {{ root: Map<string, string>, dark: Map<string, string>, compact: Map<string, string> }} ModeVars
 */

/**
```

scripts/figma/setup/css-parser.mjs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+/**
+ * Resolve the canonical per-mode variable maps from parsed CSS blocks.
+ *
+ * Merges the shorthand dual-selector blocks (which apply to both modes equally)
+ * into the base buckets without overwriting mode-specific declarations.
+ *
+ * @param {Map<string, Map<string, string>>} allBlocks
+ * @returns {{ rootVars: Map<string, string>, darkVars: Map<string, string>, compactVars: Map<string, string> }}
```
dtsanevmw:
```suggestion
 * @returns {ModeVars}
```

scripts/figma/setup/css-parser.mjs
```diff
@@ -0,0 +1,150 @@
...
+ */
+export function resolveModeVars(allBlocks) {
+  const rootVars = allBlocks.get(SELECTORS.ROOT) ?? new Map();
```
dtsanevmw:
```suggestion
  const root = allBlocks.get(SELECTORS.ROOT) ?? new Map();
```

scripts/figma/setup/css-parser.mjs
```diff
@@ -0,0 +1,150 @@
...
+  // Primary dark overrides live in the bare [cds-theme~=dark] block (235 props).
+  const darkVars = new Map(allBlocks.get(SELECTORS.DARK) ?? []);
```
dtsanevmw:
```suggestion
  const dark = new Map(allBlocks.get(SELECTORS.DARK) ?? []);
```

scripts/figma/setup/css-parser.mjs
```diff
@@ -0,0 +1,150 @@
...
+
+  // Compact overrides live in [clr-density=compact].
+  const compactVars = new Map(allBlocks.get(SELECTORS.COMPACT) ?? []);
```
dtsanevmw:
```suggestion
  const compact = new Map(allBlocks.get(SELECTORS.COMPACT) ?? []);
```

scripts/figma/setup/css-parser.mjs
```diff
@@ -0,0 +1,150 @@
...
+
+  // Merge shorthand dual-selector blocks (these apply to both modes equally).
+  // Only fills gaps — never overwrites mode-specific declarations.
+  const bothThemeVars = allBlocks.get(SELECTORS.BOTH_THEME) ?? new Map();
+  const bothDensityVars = allBlocks.get(SELECTORS.BOTH_DENSITY) ?? new Map();
```
dtsanevmw:
```suggestion
  const bothTheme = allBlocks.get(SELECTORS.BOTH_THEME) ?? new Map();
  const bothDensity = allBlocks.get(SELECTORS.BOTH_DENSITY) ?? new Map();
```

scripts/figma/setup/css-parser.mjs
```diff
@@ -0,0 +1,150 @@
...
+  for (const [k, v] of bothDensityVars) {
+    if (!rootVars.has(k)) {
+      rootVars.set(k, v);
+    }
+    if (!compactVars.has(k)) {
+      compactVars.set(k, v);
+    }
+  }
+
+  return { rootVars, darkVars, compactVars };
```
dtsanevmw:
```suggestion
  return { root, dark, compact };
```

dtsanevmw (commented):
Part 1 review

figma/figma-tokens.config.json
valentin-mladenov: Corrected in commit bde7069

figma-tokens.config.json
valentin-mladenov: Corrected in commit bde7069

figma-tokens.config.json
valentin-mladenov: Corrected in commit bde7069

figma/figma-tokens.config.schema.json
valentin-mladenov: I prefer to keep is as RegEx string until we really need the change to array. 
Using array will do multiple RegEx executions per Scope.

scripts/figma/index.mjs
valentin-mladenov: Corrected in commit 8aaa45e

scripts/figma/index.mjs
valentin-mladenov: Corrected in commit 8aaa45e

scripts/figma/index.mjs
valentin-mladenov: Corrected in commit 8aaa45e

scripts/figma/index.mjs
valentin-mladenov: Reading `CSS_FILE` is pushed to the top of the `buildRunContext` method. Corrected in commit ac9fa65.

scripts/figma/index.mjs
valentin-mladenov: For every avaliable action there is a separate controller now. That is not relevant any more.

scripts/figma/setup/css-parser.mjs
valentin-mladenov: Corrected in commit ac9fa65.

scripts/figma/setup/config.mjs
valentin-mladenov: Corrected in commit ac9fa65. Partially we still need the `WEB` key.

scripts/figma/api/push-executor.mjs
valentin-mladenov: Corrected in commit ac9fa65.

scripts/figma/api/push-executor.mjs
valentin-mladenov: `existingModes` is an array, reference it to another variable will still mutate the original array.

scripts/figma/api/push-executor.mjs
valentin-mladenov: Corrected.

scripts/figma/api/push-executor.mjs
valentin-mladenov: Corrected in erliar commit.

scripts/figma/api/push-executor.mjs
valentin-mladenov: Corrected in previous commit.

scripts/figma/api/push-executor.mjs
valentin-mladenov: Corrected in commit ac9fa65.

scripts/figma/api/push-executor.mjs
valentin-mladenov: Corrected in commit ac9fa65. Partially. The second part about `batchCount` doesnt look correct.

scripts/figma/api/push-executor.mjs
valentin-mladenov: Corrected.

scripts/figma/core/collections.mjs
valentin-mladenov: Corrected in commit ac9fa65.

scripts/figma/core/collections.mjs
valentin-mladenov: Corrected in commit ac9fa65.

scripts/figma/core/collections.mjs
valentin-mladenov: Corrected in commit ac9fa65.

scripts/figma/core/collections.mjs
valentin-mladenov: Corrected in commit ac9fa65.

scripts/figma/setup/css-parser.mjs
valentin-mladenov: Corrected in commit ac9fa65.

scripts/figma/setup/css-parser.mjs
valentin-mladenov: Corrected in commit ac9fa65.

scripts/figma/setup/css-parser.mjs
valentin-mladenov: Corrected in commit ac9fa65.

scripts/figma/setup/css-parser.mjs
valentin-mladenov: Corrected in commit ac9fa65.

scripts/figma/setup/css-parser.mjs
valentin-mladenov: Corrected in commit ac9fa65.

scripts/figma/setup/css-parser.mjs
valentin-mladenov: Corrected in commit ac9fa65.

figma/setup/css-parser.mjs
valentin-mladenov: Corrected in commit ac9fa65.

scripts/figma/index.mjs
valentin-mladenov: This is not needed any more `varLookup` is moved in saparate `context.mjs` file.

scripts/figma/index.mjs
valentin-mladenov: This is not needed any more `varLookup` is moved in saparate `context.mjs` file.

figma/api/push-executor.mjs
valentin-mladenov: This is corrected.

scripts/figma/api/extract-view.mjs
valentin-mladenov: It is corrected to `name: colDef.name,`

scripts/figma/index.mjs
valentin-mladenov: This is not needed any more varLookup is moved in saparate context.mjs file.

scripts/figma/api/push-executor.mjs
valentin-mladenov: Correctd in commit c3d6145

scripts/figma/api/push-executor.mjs
valentin-mladenov: Correctd in commit c3d6145

scripts/figma/api/diff-printer.mjs
valentin-mladenov: I prefer the original implementation, because with `diffReport.some` we iterate  `diffReport` twice.

scripts/figma/api/extract-view.mjs
valentin-mladenov: This make no sense.

scripts/figma/api/extract-view.mjs
valentin-mladenov: This make no sense.

scripts/figma/api/extract-view.mjs
valentin-mladenov: This make no sense.

scripts/figma/api/extract-view.mjs
valentin-mladenov: `colName` is not needed since the `collectionSuffix` is no longer relavant.

scripts/figma/api/extract-view.mjs
valentin-mladenov: Corrected in commit 3981e30

scripts/figma/api/extract-view.mjs
valentin-mladenov: Corrected in commit 3981e30

figma/api/extract-view.mjs
valentin-mladenov: Corrected in commit 3981e30

scripts/figma/api/extract-view.mjs
valentin-mladenov: Removed in commit 3981e30

scripts/figma/api/diff-printer.mjs
valentin-mladenov: Corrected in commit 3981e30

figma/api/diff-printer.mjs
valentin-mladenov: No changes for this.

figma/api/diff-printer.mjs
valentin-mladenov: No changes for this.

scripts/figma/api/push-executor.mjs
valentin-mladenov: NO changes for this.

scripts/figma/index.mjs
valentin-mladenov: Human readable count is removed in commit 3014948

scripts/figma/index.mjs
valentin-mladenov: Human readable count is removed in commit 3014948

.env.figma
valentin-mladenov: File is removed.

figma-tokens.config.json
valentin-mladenov: Those are not exacly the same.
With the new implementation they are fragmented even more.

Check commit e560d92 for more details.

dtsanevmw (commented):
## Part 2 review

Re-reviewed everything changed since my "Part 1 review" comment (`d2a43f6c9e` → current head). Scope this round: (1) what's new, (2) whether the "pr feedback" commits actually addressed Part 1, and (3) readability/reuse, since most of this tool is AI-generated and the goal is to keep it maintainable by a human.

I've pushed a branch with all the fixes below already implemented and verified: **[`dtsanevmw/cde-3113-review-suggestions`](https://github.com/vmware-clarity/ng-clarity/compare/vmadenov/token-sync-to-figma_CDE-3113...dtsanevmw/cde-3113-review-suggestions)** (branched off this PR's current head). Feel free to merge it in, cherry-pick pieces, or just use it as a reference — happy to open a PR against your branch instead if that's easier to review.

### Bugs/regressions introduced since Part 1

- **`core/variable-diff.mjs`** — the color-equality tolerance was loosened from `1e-5` to `1e-2` with no explanation (commit `bde7069eb`, "update some logic"). On Figma's 0–1 color scale that's ~2.5/255 per channel — enough to hide an intentional small color/alpha edit from ever reaching Figma. Reverted to `1e-5` in the branch; if there was a specific reason for loosening it (e.g. float rounding from a color-space conversion), worth a comment explaining exactly how loose it needs to be.
- **`controllers/dry-run.controller.mjs`** — logs `ctx.figmaFileKey`, but `RunContext` never sets that field (only `loadEnv()` does, which dry-run intentionally never calls so it works without credentials). `--dry-run` always printed `file: undefined`. Fixed by dropping the file key from the dry-run banner.
- **`core/planner.mjs`** (`resolveVariableId`) — when a token retypes, the old variable is renamed to `<name>_deprecated` with no check for whether that name is already taken. If a token retypes twice across separate runs without anyone cleaning up Figma in between, the second retype would try to reuse the same `_deprecated` name. Added a uniqueness guard (`_deprecated`, `_deprecated_2`, …).

### Part 1 feedback marked "addressed" that wasn't

Verified by reading the current code, not just the commit messages:

- **`api/push-executor.mjs`** — still did `existingModes.length = 0; existingModes.push(...)`, mutating the caller's array in place. Fixed to reassign instead.
- **`api/extract-view.mjs`** — still all `.find()`/`.filter()` linear scans (worst case O(variables × modeValues) inside a `.map()`). Pre-indexed into Maps.
- **`api/diff-printer.mjs`** — `created`/`updated`/`deprecated` each still hand-built their line independently; extracted a shared `formatEntry()` helper. Also fixed a real formatting bug this duplication caused: the deprecated badge rendered as `3⚠ ` instead of `⚠3` (icon-then-number everywhere else).
- **`figma-tokens.config.json`** — the "could these be reused instead of manually kept in sync?" comment on the duplicate exclude lists is now addressed for real: generalized `config.mjs`'s `resolveReference()` (previously dead code — its own comment admitted "there is no internal JSON references within the config file") to support `{ "$ref": "key" }` entries mixed into an array, and hoisted the shared prefix lists into `typographyColorPrefixes`/`objectColorPrefixes`. Verified live that `loadConfig()` produces byte-identical effective filters to before.

### Reuse — biggest one is a real bug, not just style

- **`setup/css-parser.mjs`** was a hand-rolled line-scanner with a documented assumption ("always true for output from sass/csso"). I tested it against the **real `dist/clr-ui/clr-ui.css`** (the exact file this tool reads) and found it silently corrupts data: any custom-property value wrapped across multiple lines by Sass's formatter gets truncated to its first line only (e.g. a `linear-gradient(...)` spanning several lines becomes the literal string `"linear-gradient("`), and dozens of variables (all the `--clr-badge-*` tokens, several `--clr-table-*`/`--clr-dropdown-*` tokens) were dropped entirely once the scanner's manual brace-depth tracking desynced. Rewrote it on `postcss` (already a transitive dependency via `postcss-cli`/`autoprefixer` — added as an explicit devDependency) and confirmed zero data loss against the same real file. **This means the tool as currently written would very likely push wrong/missing values for a meaningful number of real tokens** — worth prioritizing this one regardless of anything else in this review.
- **`setup/config.mjs`** hand-rolled the same validation rules already expressed in `figma-tokens.config.schema.json` (required fields, `mode.source` enum, the `humanReadable` `--` pattern), so the two could silently drift. Swapped to `ajv` validating against the existing schema; verified live that it both accepts the real config and rejects a broken one with a clear error path.
- **`util/temp-id.mjs`**'s module-level mutable counter (shared, unresettable state across the whole run) replaced with `crypto.randomUUID()` — a Node built-in, no new dependency, and no more global state to reason about.

### Controllers — cut 2 of the 6 new files

- `dry-run.controller.mjs` and `extract.controller.mjs` were near-duplicates by the code's own admission (`plan.mjs`'s doc comment: "both branches build the same full push plan... only their tail behavior differs"). Merged into `offline-plan.controller.mjs`.
- `preview.controller.mjs` re-implemented `buildPushPlan`'s collection-aggregation loop by hand instead of reusing it. `executePush()` now takes a `preview` flag so both push and preview share one aggregation implementation instead of two that can drift.

### CI

- `figma-publish.yml`'s `workflow_call` trigger is unused — `release.yml`/`release-preview.yml` push tokens via a plain `npm run figma:push` step instead, so they never got the `concurrency` guard that exists specifically to stop overlapping partial pushes. Added a matching `concurrency` group to both, and corrected the comment in `figma-publish.yml` that claimed it was already being called that way.

### Also fixed

- Dead `getBranches()` in `figma-client.mjs` (no callers since branch-mode was removed earlier in this PR) and its stale JSDoc type.
- Stale docs: README said "7 collections" (actual: 9), both README and `figma-tokens.config.json`'s `Colors`/`Spacings` comments still referenced a "top-level humanReadable map" that's been per-collection since this PR's own refactor, and the credential setup step read as garbled text after `.env.figma` was deleted.

### Two things I looked at but deliberately did **not** change

- **`push-executor.mjs` does a full `getVariables` GET after every collection's POST** purely to relearn IDs Figma just assigned — Figma's POST response is documented to include a `tempIdToRealId` map for exactly this, which would let this loop skip ~9 round-trips (one per collection) and roughly halve the push's wall-clock time. I didn't implement this: I can't verify the exact response shape or write-then-read consistency against the live API from here, and getting it wrong would risk silently corrupting variable IDs in the shared production Figma file with no rollback. Left a comment recommending it as a follow-up once someone can test it against the real API.
- **The `humanReadable` collections feature** (Colors/Spacings) is the single biggest source of the duplicated branch logic in `planner.mjs`'s `buildCollectionPlan` (still split it into `planHumanReadableCollection`/`planRegularCollection` for readability, but the feature itself adds real complexity). If a smaller initial merge is preferred, this is the one feature substantial enough to defer to a follow-up PR rather than a mechanical cleanup — your call, not something I'd unilaterally cut.

I ran the existing `_tests/value-converters.test.mjs` suite and eslint against everything in the suggestion branch. A few of the changes (`planner.mjs`'s collision guard, and the full push pipeline end-to-end) I could only verify by careful manual tracing rather than executing them against a live Figma file — worth a `npm run figma:push:dry-run` sanity check locally before merging either the fixes or this PR itself.

dtsanevmw (dismissed):

dtsanevmw (approved):

PR #2467, opened 2026-06-22
chore(dropdown): remove unsupported disabled attribute from nested trigger demos

dtsanevmw:
## PR Checklist

## PR Type

- [x] Chore

Two demo entries (`dropdown-angular-nested.demo.html`, `dropdown-angular-shadow-root.demo.html`) applied the native `disabled` attribute directly to a nested `clrDropdownTrigger` button, and one applied a `clrDisabled` attribute that `ClrDropdownTrigger` doesn't declare (so it was a no-op). Both were added just to preview disabled colors.

`ClrDropdownTrigger` has no disabled functionality of its own. Native `disabled` makes a button unfocusable per the HTML spec (confirmed empirically: re-applying `disabled` to a focused element blurs it back to `document.body` in the same tick, with no workaround), which broke arrow-key navigation past it. This looked like a keyboard-navigation bug in `FocusService`, but it's expected browser behavior triggered by an unsupported demo pattern, not a Clarity defect.

Issue Number: CDE-3003

## What is the new behavior?

Removed the unsupported `disabled`/`clrDisabled` attempts on `clrDropdownTrigger` from both demo files. No production code changes. If trigger disabling is wanted in the future, it should go through a `clrDisabled`-style API consistent with `clrDropdownItem`, not the native attribute — that would be a separate feature PR.

- [ ] Yes
- [x] No

## Other information

Closing CDE-3003 as not-a-bug; see the ticket for details.


idonchev-vmw (approved):
The obsolescence claim is technically verified — the removed attributes had no functional effect on ClrDropdownTrigger, and no other tests/demos depend on the removed markup.

PR #2470, opened 2026-06-22
fix(stepper): prevent step button from submitting parent form and fix navigation logic

dtsanevmw:
## PR Checklist

## PR Type

- [x] Bugfix

Two related bugs in `ClrStepButton`:

1. The button used `[type]="'button'"` (Angular property binding), which sets the DOM property but not the HTML attribute. Browsers default the `type` attribute to `submit`, causing the button to submit any parent `<form>` on click.

2. Navigation logic used `if (this.previousButton) ... else navigateToNextPanel()`, meaning any button type that was not `previous` — including custom string types — would unconditionally call `navigateToNextPanel()`.

Issue Number: CDE-3004

## What is the new behavior?

1. The button now uses `[attr.type]="'button'"` to set the HTML attribute directly.

2. Navigation is now gated on explicit enum checks: only `ClrStepButtonType.Next` and `ClrStepButtonType.Submit` navigate forward; only `ClrStepButtonType.Previous` navigates backward. Custom or unknown button types perform no navigation.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/projects/angular/stepper/step-button.spec.ts
+++ b/projects/angular/stepper/step-button.spec.ts
@@ -103,6 +103,34 @@ describe('ClrStepButton Next', () => {
   });
 });
 
+describe('ClrStepButton Custom type', () => {
+  let fixture: ComponentFixture<any>;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      declarations: [TestComponent],
+      providers: [StepperService, { provide: StepperService, useClass: MockStepperService }],
+      imports: [ReactiveFormsModule, NoopAnimationsModule, ClrStepperModule],
+    });
+
+    fixture = TestBed.createComponent(TestComponent);
+    fixture.componentInstance.buttonType = 'custom' as ClrStepButtonType;
+    fixture.detectChanges();
+  });
+
+  it('should not navigate when button type is not next, previous, or submit', () => {
+    const stepperService = fixture.debugElement.query(By.directive(ClrStepButton)).injector.get(StepperService);
+    spyOn(stepperService, 'navigateToNextPanel');
+    spyOn(stepperService, 'navigateToPreviousPanel');
+
+    fixture.nativeElement.querySelector('.clr-step-button').click();
+    fixture.detectChanges();
+
+    expect(stepperService.navigateToNextPanel).not.toHaveBeenCalled();
+    expect(stepperService.navigateToPreviousPanel).not.toHaveBeenCalled();
+  });
+});
+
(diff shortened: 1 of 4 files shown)


projects/angular/stepper/step-button.ts
```diff
@@ -40,13 +40,13 @@ export class ClrStepButton implements OnInit {
     this.previousButton = this.type === ClrStepButtonType.Previous;
   }
 
-  @HostListener('click')
-  navigateToNextPanel() {
-    if (this.previousButton) {
+  @HostListener('click', ['$event'])
+  navigateToNextPanel(event: Event) {
+    event.preventDefault();
```
valentin-mladenov, line 45: Why do we need `event.preventDefault();` the custom buttons should do their custom stuff?

```suggestion
```

projects/angular/stepper/step-button.spec.ts
```diff
@@ -88,6 +88,47 @@ describe('ClrStepButton Next', () => {
...
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      declarations: [TestComponent],
+      providers: [StepperService, { provide: StepperService, useClass: MockStepperService }],
+      imports: [ReactiveFormsModule, NoopAnimationsModule, ClrStepperModule],
+    });
+
+    fixture = TestBed.createComponent(TestComponent);
+    fixture.componentInstance.buttonType = 'custom' as ClrStepButtonType;
```
valentin-mladenov, line 117: Casting to `ClrStepButtonType` is not really needed since the `buttonType` input accept both `ClrStepButtonType` and `string`.


```suggestion
    fixture.componentInstance.buttonType = 'custom';
```

projects/angular/stepper/step-button.ts
```diff
@@ -21,7 +21,7 @@ export enum ClrStepButtonType {
   host: {
     '[class.clr-step-button]': 'true',
     '[class.btn]': 'true',
-    '[type]': "'button'",
+    '[attr.type]': "'button'",
```
valentin-mladenov, line 24: Is adding `attr.` really needed?

valentin-mladenov (changes requested):
Check my comments.

----------------------------------------

Things to be moved in to a separate JIRA/PR. `submitButton` and `previousButton` host bindings can only be changed upon `ngOnInit` cycle. If the `type` of the Button changes dynamically those bindings won't update.

```
  @HostBinding('class.btn-success') submitButton = false;
  @HostBinding('class.btn-link') previousButton = false;
```

PR #2473, opened 2026-06-22
fix(datagrid): prevent sticky columns from being obscured by expanded row detail

dtsanevmw:
## PR Checklist

## PR Type

- [x] Bugfix

When a datagrid has locked (sticky) columns and horizontal scroll, expanding a row causes the sticky column cells to be obscured by the expanded row detail content. The detail element was inadvertently creating a competing stacking context.

Issue Number: CDE-3127

## What is the new behavior?

`.datagrid-row-detail` is left without `position` or `z-index`, so it does not create a stacking context. Sticky column cells (z-index: 500) naturally render above the expanded detail content without interference.

- [ ] Yes
- [x] No

## Other information


valentin-mladenov (approved):
LGTM

PR #2476, opened 2026-06-22
fix(tree-view): prevent double-tab after clicking tree node

dtsanevmw:
## PR Checklist

## PR Type

- [x] Bugfix

After clicking a tree node, pressing Tab could require two presses to move focus outside the tree, or leave the tree completely unreachable by keyboard afterwards. This happened for a few different reasons, all rooted in the same roving-tabindex mechanism used by `clr-tree-node`:

- `broadcastFocusedNode()` was called asynchronously (via a timeout), so peer nodes still had `tabindex="0"` when focus arrived, creating multiple tabbable elements.
- Interactive content projected into a node (e.g. an `<a class="clr-treenode-link">`) is natively focusable in addition to the node's own roving-tabindex container, adding an extra Tab stop for every node that uses a link.
- Clicking a node whose content has no focusable descendant (e.g. plain text, such as a parent/root node) makes the browser focus the node's content container directly on `mousedown`, since it always carries a `tabindex` attribute. `focusTreeNode()`'s "already focused" guard then skipped `setTabIndex(0)` entirely, leaving that node's `tabindex` stuck at `-1` — making it permanently unreachable by Tab/Shift+Tab afterwards.

Issue Number: CDE-3007

## What is the new behavior?

- `broadcastFocusedNode()` is called synchronously at the start of `focusTreeNode()`, before `containerEl.focus()`, so all peer nodes reset to `tabindex="-1"` before the newly focused node receives focus.
- `ClrTreeNodeLink` now sets `tabindex="-1"` on its host anchor, matching the caret button and checkbox, which already excluded them from the native Tab sequence. The link remains fully clickable and keyboard-activatable (Enter/Space), it's just no longer a separate Tab stop.
- `focusTreeNode()` now unconditionally calls `setTabIndex(0)` on its own container (idempotent), instead of only doing so when it also needs to call `.focus()`. This keeps the container's `tabindex` correct even when the browser already moved DOM focus to it directly, so the node always remains reachable afterwards.

Together, these changes ensure only one node in the tree has `tabindex="0"` at any time, and that node is always the one that currently has focus — so a single Tab press reliably moves focus in or out of the tree.

- [ ] Yes
- [x] No


valentin-mladenov (changes requested):
This is not a complete solution.

The use case of the Routing tree is it's nodes to be navigated through `Tab` key.
Setting `tabindex` to `-1` removes that functionality and leaves only the arrows navigation.

valentin-mladenov (approved):

PR #2569, opened 2026-07-20
fix(datagrid): reset Items provider when *clrDgItems view is destroyed

dtsanevmw:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

`ClrDatagridItems` (`*clrDgItems`) unconditionally calls `items.smartenUp()` in its
constructor, which flips the shared `Items` provider into "smart" mode and wires up
live `_filtersSub` / `_sortSub` / `_pageSub` subscriptions on it. When the
`*clrDgItems` view is later destroyed, `ngOnDestroy()` only unsubscribes its own two
local subscriptions — it never calls `items.smartenDown()` / `items.destroy()`. The
`Items` provider is scoped to the whole `<clr-datagrid>` and can be shared by more
than one template branch (e.g. a datagrid with both a server-driven `*ngFor` row path
and a client-side `*clrDgItems` row path gated by an `@if`), so once that shared
provider is stuck `smart = true`, its stale `_pageSub` subscription keeps firing on
every page change.

Concretely, on a server-driven, paginated grid this means:
- `Items._all` / `_filtered` stop being refreshed with newly-fetched server pages,
  because the `!items.smart` guard in `ClrDatagrid` that normally applies new data
  no longer fires.
- The leaked `_pageSub` still triggers `_changePage()` on every page change, which
  slices the **stale** page-1 data using the **new** page's `firstItem`/`lastItem`
  (e.g. `10`/`19` against a 10-item array) — always out of range, so `displayed`
  becomes `[]`.
- `clr-dg-placeholder`'s `emptyDatagrid` check (`items.displayed.length === 0`) then
  incorrectly shows "No items found" over the grid, even though the correct page
  data was fetched and rendered underneath.

## What is the new behavior?

`ClrDatagridItems.ngOnDestroy()` now calls `this.items.smartenDown()` after
unsubscribing its own subscriptions, so tearing down a `*clrDgItems` view always
puts the shared `Items` provider back into "dumb" mode and clears out the
filter/sort/page subscriptions it created. This prevents the leaked pagination
subscription from slicing stale data against a new page's offsets.

Added regression tests in `datagrid-items.spec.ts` covering:
- The `Items` provider going back to `smart === false` once the `*clrDgItems` view
  is destroyed.
- `displayed` no longer reacting to `Page` changes after destroy, instead of
  slicing stale data with new offsets.

- [ ] Yes
- [x] No

## Other information

Repro path: a datagrid whose template briefly instantiates a `*clrDgItems` view
(directly or via a conditional template branch) alongside another row-rendering
path that manages `Items.all` itself (e.g. server-driven pagination). Once that
transient view is destroyed, the shared `Items` provider was left permanently
"smart" with a dangling page-change subscription, corrupting pagination on
subsequent page navigations.

### DIFF
--- a/projects/angular/data/datagrid/datagrid-items.spec.ts
+++ b/projects/angular/data/datagrid/datagrid-items.spec.ts
@@ -5,6 +5,7 @@
  * The full license information can be found in LICENSE in the root directory of this project.
  */
 
+import { CommonModule } from '@angular/common';
 import { Component, TrackByFunction, ViewChild } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 
@@ -30,6 +31,22 @@ class FullTest {
   numbers = [1, 2, 3, 4, 5];
 
   trackBy: (index: number, item: number) => any;
+}
+
+@Component({
+  template: `
+    <ul *ngIf="showItems">
+      <li *clrDgItems="let n of numbers; trackBy: trackBy">{{ n }}</li>
+    </ul>
+  `,
+  standalone: false,
+})
+class DestroyTest {
+  numbers = [1, 2, 3, 4, 5];
+
+  trackBy: (index: number, item: number) => any;
+
+  showItems = true;
 }
 
 @Component({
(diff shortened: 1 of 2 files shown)


valentin-mladenov (approved):
LGTM.

PR #2581, opened 2026-07-31
fix(popover): make popovers safe across ShadowRoot and cross-window origins

idonchev-vmw:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-3155

`ClrPopoverContent.getScrollableParents()` throws when a popover's origin crosses a window (realm) boundary:

```
TypeError: Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.
  at _ClrPopoverContent.getScrollableParents (clr-angular-popover-common.mjs:733)
  at _ClrPopoverContent.listenToScrollForElementOrigin (…:772)
  at _ClrPopoverContent.listenToScrollEvents (…:768)
```

Confirmed in production on `@clr/angular` 18.2.0 + Angular 21.2.16, in the iframe-based vSphere/VCF plugin architecture.

### A note on reproducing this

The StackBlitz attached to the ticket is **not a valid repro** and cost us some time — worth recording so the next person doesn't repeat it:

- It pins **`@clr/angular` 17.9.0** (the `18.2.10` visible throughout is `@angular/*`, not Clarity). Clarity 17 had two different, pre-CDK popover implementations; the crash it produces comes from the old `Popover` class (`Popover.scrolls` / `addScrollEventListeners` in that stack trace), which was **deleted in 18.0.0** by the CDK migration. That specific crash is long gone.
- It has **no iframe**, so it can't exercise the actual bug. A pure `ViewEncapsulation.ShadowDom` component on current versions does not throw — verified directly.

The real trigger is the realm boundary, which only the actual product architecture (or a deliberately-built iframe harness) reproduces.

### Root cause

`instanceof` is realm-bound: a node created by another window fails `instanceof` against this window's constructors. The walk relied on it in two places:

```ts
while (parent && !(parent instanceof HTMLHtmlElement)) {   // foreign <html> not recognised
  if (parent instanceof ShadowRoot) {                      // foreign ShadowRoot not recognised
    parent = parent.host;
  }
  const { overflowY, overflowX } = window.getComputedStyle(parent);   // ← throws
```

So the walk either calls `getComputedStyle` on an unrecognised foreign `ShadowRoot`, or steps past the foreign `<html>` into its `Document`. Same class of bug as [a CVE-level `instanceof` bypass in DOMPurify](https://github.com/cure53/DOMPurify/security/advisories/GHSA-hpcv-96wg-7vj8).

A sweep against a verbatim copy of the pre-fix code established exactly which structures throw:

| Ancestor structure | Pre-fix |
|---|---|
| `<body>` inside a **same-realm** ShadowRoot | no throw |
| Fully detached subtree | no throw |
| **Plain (non-shadow) `DocumentFragment`** in the chain | **throws** |
| **Cross-realm `instanceof` fall-through** | **throws** |

The `DocumentFragment` case was a second, realm-independent trigger found during that sweep — a plain fragment shares `ShadowRoot`'s `nodeType` but has no `host`, so `instanceof ShadowRoot` is false and the walk fell straight through.

### Three further bugs behind the exception

Patching only the throw would have converted a loud, obvious failure into three silent ones. Once the crash was suppressed, the same realm boundary broke:

1. **Overlay anchors to the wrong place** — CDK's `FlexibleConnectedPositionStrategy` reads `getBoundingClientRect()` relative to the origin's *own* window, not the overlay's. Known, unfixed CDK gap: [angular/components#21482](https://github.com/angular/components/issues/21482) (closed without a fix). `FlexibleConnectedPositionStrategyOrigin` has no "virtual element" escape hatch, unlike Popper v2 / Floating UI which support this natively.
2. **Outside clicks never close the popover** — CDK's dispatcher listens only on the overlay's own document, so clicks inside the origin's iframe are invisible to it. (Matches the reporter's "can't be closed except by clicking the trigger".)
3. **Scrolling neither repositions nor closes** — the walk stopped at the realm boundary, so no scroll container beyond it was ever found.

## What is the new behavior?

- **`getScrollableParents` is realm-safe**, not merely realm-stopping. All `instanceof` replaced with realm-independent shape checks (`nodeType`, `tagName`, duck-typed `ShadowRoot` via `host`), computed style resolved through each node's *own* window, and the walk now continues across iframe boundaries via `window.frameElement` up to the true top document — collecting scroll containers on both sides, and each embedded document it passes through.
- **`resolveCrossWindowOrigin()`** supplies CDK a `Point`-based origin (a shape it natively supports) translated by the cumulative ancestor-iframe offset, with live getters so it keeps tracking across repositions. Memoized per microtask, so one reposition costs one layout read instead of ~6. Falls back to the untranslated rect when the frame chain can't be resolved (cross-origin), which is numerically what CDK would have computed anyway.
- **Cross-window outside-click detection** via capture-phase `click`/`auxclick` on the origin's own document, mirroring CDK's own dispatcher.
- **Same-window popovers are provably untouched** — `getCrossWindowOriginContext()` gates all of the above; a same-window origin takes an early return in every path, enforced by a test asserting zero extra listeners.

Also adds a demo page (Popovers demo) that reproduces the whole thing end-to-end by relocating a live `clr-dropdown` into a ShadowRoot inside a same-origin iframe.

### Notes for reviewers

- **Capture phase is load-bearing.** A bubble-phase or `pointerdown` listener cannot suppress the trigger's own click handler, so a toggle re-click would close and instantly reopen. Since trigger-click is currently the *only* way affected users can dismiss a popover, getting this wrong would have removed their last working interaction. Caught in review; regression test added.
- **Live getters over a snapshot, deliberately.** `positionStrategy` is only read at overlay creation; ordinary scroll repositioning calls `overlayRef.updatePosition()` and reuses the same origin object. A frozen snapshot goes stale on the first scroll after opening. The microtask memo gets the performance win without giving that up.
- Every fix here was verified by reverting it and confirming the corresponding test fails, then restoring — including a mutation check on the same-window guard.

- [ ] Yes
- [x] No

No public API change (`public-api:check` clean — the new helpers are internal, not exported from the barrel).

## Other information

Testing done:
- Full `clr-angular` unit suite: **3483/3483**, coverage above thresholds.
- New coverage: realm-safe traversal across nested iframes, plain-`DocumentFragment` ancestor, cross-window outside-click close, toggle re-click suppression, positioning integration through `positionStrategy`, microtask memoization, iframe border+padding offset, and the same-window no-op contract.
- Manually verified via the demo page.

AI-Tool-Used: Claude Code
AI-Tool-Use-Level: 2
AI-Code-Category: 1


projects/angular/popover/common/utils/cross-window-origin.ts
```diff
@@ -0,0 +1,114 @@
...
+      const offset = getCumulativeFrameOffset(elementWindow) ?? { x: 0, y: 0 };
+      return element.getBoundingClientRect().top + offset.y;
+    },
+    get width(): number {
+      return element.getBoundingClientRect().width;
+    },
+    get height(): number {
+      return element.getBoundingClientRect().height;
+    },
+  };
```
dtsanevmw:
```suggestion
const offset = getCumulativeFrameOffset(elementWindow);

  if (offset === null) {
    return origin;
  }

  const rect = element.getBoundingClientRect();

  return {
    x: rect.left + offset.x,
    y: rect.top + offset.y,
    width: rect.width,
    height: rect.height,
  };
```

idonchev-vmw: Good catch on the redundant getCumulativeFrameOffset call, applied in d2ce3da0e (now called once per getter access instead of an extra time up front). I kept the live getters though: positionStrategy is only read at overlay creation, but ordinary scroll-triggered repositioning (overlayRef.updatePosition()) reuses that same origin object without re-invoking resolveCrossWindowOrigin(), so a frozen snapshot would go stale on the next scroll/resize after the popover opens — confirmed this by testing both versions live in the demo page.

idonchev-vmw:
Also pushed a follow-up (15b35f630, separate from the review feedback): CDK's outside-click detection only listens on this window's document, so clicking anywhere inside a cross-window origin's own iframe never closed the popover — added a parallel listener on that origin's document, reusing the existing toggle-button suppression logic. Related but unfixed: scroll-triggered repositioning can't reach this trigger either, since getScrollableParents stops at the ShadowRoot realm boundary by design (same guard that fixes the crash) — flagging as a known limitation rather than fixing, since discovering scrollable containers inside a foreign realm is a separate, larger problem.

projects/angular/popover/common/popover-content.ts
```diff
@@ -249,27 +252,65 @@ export class ClrPopoverContent implements OnDestroy, AfterViewInit {
...
+
+    if (!originEl || !originWindow || originWindow === window) {
+      return null;
+    }
+
+    return this.zone.runOutsideAngular(() =>
+      merge(fromEvent(originWindow.document, 'pointerdown'), fromEvent(originWindow.document, 'click')).subscribe(
+        event => this.zone.run(() => this.handleOutsideClick(event))
+      )
+    );
```
dtsanevmw, line 387: **Bug: a cross-window toggle re-click closes the popover and immediately reopens it.**

The same-window path only works because CDK's `OverlayOutsideClickDispatcher` registers **capture-phase** listeners and emits on **`click`** (pointerdown is only recorded as the press origin). That's why `handleOutsideClick`'s `stopPropagation()` can suppress the trigger's own bubble-phase click handler before it runs.

This listener breaks both assumptions — bubble phase, and closing on `pointerdown`. For a trigger living in the foreign document (the exact scenario this PR enables):

1. `pointerdown` bubbles to the foreign document → `isToggleButton` is true → `closePopover()` (the `stopPropagation()` is useless: wrong event, wrong phase).
2. The subsequent `click` fires on the trigger → `ClrDropdownTrigger`'s `@HostListener('click')` → `toggleWithEvent()` flips `open` back to `true`.

Net effect: the toggle button can never close a cross-window popover — it flickers and reopens. Secondary parity issue: any interaction in the foreign document (starting a drag/text selection) closes on mousedown, while same-window behavior closes on a completed click. The `pointerdown`+`click` merge also runs the handler twice per click whenever the first invocation doesn't close.

Mirroring CDK fixes all three:

```suggestion
    return this.zone.runOutsideAngular(() =>
      merge(
        fromEvent(originWindow.document, 'click', { capture: true }),
        fromEvent(originWindow.document, 'auxclick', { capture: true })
      ).subscribe(event => this.zone.run(() => this.handleOutsideClick(event)))
    );
```

The new spec in `popover-content.spec.ts` should then dispatch `click` instead of `pointerdown`, and a second spec covering the toggle-button-in-foreign-doc case (trigger with a click listener calling `toggleWithEvent`; assert the popover ends closed and the listener wasn't invoked) would have caught this.

Related gap worth a note or follow-up: this listens only on the origin's own document, so with nesting depth ≥ 2, clicks in an *intermediate* frame are invisible to both CDK and this listener. Walking the frame chain and listening on each same-origin document would cover it.

---

projects/angular/popover/common/popover-content.ts
```diff
@@ -249,27 +252,65 @@ export class ClrPopoverContent implements OnDestroy, AfterViewInit {
...
+    // web components (cds-icon) register as outside pointer events, so if the event target is inside the content panel return early
+    if (this.elementRef?.nativeElement?.contains(event.target as Node)) {
+      return;
+    }
+
+    // Check if the same element that opened the popover is the same element triggering the outside pointer events (toggle button)
+    const isToggleButton =
+      this.popoverService.openEvent &&
+      ((this.popoverService.openEvent.target as Element).contains(event.target as Element) ||
+        (this.popoverService.openEvent.target as Element).parentElement.contains(event.target as Element) ||
```
dtsanevmw: Pre-existing NPE that moved with the refactor — cheap to harden while touching it: if the open target has no `parentElement` (detached node, or the document root), this throws.

```suggestion
        (this.popoverService.openEvent.target as Element).parentElement?.contains(event.target as Element) ||
```

---

projects/angular/popover/common/popover-content.ts
```diff
@@ -374,7 +415,11 @@ export class ClrPopoverContent implements OnDestroy, AfterViewInit {
     const overflowScrollKeys = ['auto', 'scroll', 'clip'];
     const scrollableParents: (HTMLDocument | HTMLElement)[] = [window.document];
 
-    while (parent && !(parent instanceof HTMLHtmlElement)) {
+    while (
+      parent &&
+      (parent instanceof ShadowRoot || parent instanceof Element) &&
+      !(parent instanceof HTMLHtmlElement)
+    ) {
```
dtsanevmw: This guard works *because* `instanceof` fails cross-realm — the fix leans on the very pitfall the PR description explains, but nothing in the code says so. A future well-intentioned refactor to `nodeType` duck-typing would silently push the walk into the foreign realm. Please make the intent explicit:

```suggestion
    // Deliberately realm-sensitive: `instanceof` fails for nodes from another window
    // realm, so for a cross-window origin this walk stops at the realm boundary instead
    // of falling through to the foreign Document (where getComputedStyle would throw).
    // Trade-off: scroll containers inside the foreign document are never collected, so
    // scrolling there won't reposition/close the popover.
    while (
      parent &&
      (parent instanceof ShadowRoot || parent instanceof Element) &&
      !(parent instanceof HTMLHtmlElement)
    ) {
```

Worth being explicit about the consequence too: since the *trigger itself* is foreign, the loop exits on the first iteration for cross-window origins — no scroll containers inside the iframe are ever collected, which is why the demo has to poll `updatePosition()` every 200 ms. That's a real product gap; consider a follow-up ticket.

Alternative that closes the gap instead of documenting it: make the walk realm-safe (duck-type via `nodeType`, resolve styles through each element's own `elementWindow.getComputedStyle()`, and at an embedded document's root continue from `window.frameElement`). That fixes the same crash *and* keeps scroll-repositioning working across the whole frame chain, making the demo's polling workaround unnecessary. Happy to share a working patch (full suite passing) if you want to take that route.

On the plus side, this guard also fixes a latent same-realm edge: a walk topping out at a `Document`/`DocumentFragment` without passing `<html>` previously reached `getComputedStyle(document)` and threw; now it stops cleanly.

---

projects/angular/popover/common/utils/cross-window-origin.ts
```diff
@@ -0,0 +1,112 @@
...
```
dtsanevmw, line 98: **Performance: many forced layout reads per reposition.** CDK's `FlexibleConnectedPositionStrategy._getOriginRect()` builds the rect for a Point origin as `{top: y, bottom: y + height, left: x, right: x + width, …}` — one `apply()` reads `x` twice, `y` twice, plus `width` and `height`. As written that's 6 × `getBoundingClientRect()` on the element plus 2 full frame-chain walks (each with a `getBoundingClientRect()` per ancestor iframe) — ~10 layout reads per reposition, and repositions run per scroll event.

All six reads happen synchronously within one `apply()`, so a microtask-scoped memo collapses them to a single computation with no loss of liveness across repositions:

```suggestion
  let cached: { x: number; y: number; width: number; height: number } | null = null;

  // The position strategy reads x/y/width/height several times per apply(); memoize the
  // translated rect for the current microtask so one reposition costs one layout read.
  const read = () => {
    if (!cached) {
      const rect = element.getBoundingClientRect();
      const offset = getCumulativeFrameOffset(elementWindow) ?? { x: 0, y: 0 };
      cached = { x: rect.left + offset.x, y: rect.top + offset.y, width: rect.width, height: rect.height };
      queueMicrotask(() => (cached = null));
    }
    return cached;
  };

  return {
    get x(): number {
      return read().x;
    },
    get y(): number {
      return read().y;
    },
    get width(): number {
      return read().width;
    },
    get height(): number {
      return read().height;
    },
  };
```

Note for the spec: with memoization, the "recomputes live" test must `await Promise.resolve()` between the first read and the post-move assertion (the cache clears on the next microtask). At minimum, even without the memo, `x` and `y` should share one offset walk rather than each doing their own.

---

projects/angular/popover/common/utils/cross-window-origin.ts
```diff
@@ -0,0 +1,112 @@
...
+  if (origin instanceof ElementRef) {
+    return origin.nativeElement;
+  }
+
+  if (origin instanceof Element) {
+    return origin;
+  }
+
+  return null;
+}
```
dtsanevmw, line 113: `origin instanceof Element` is itself realm-sensitive — it returns false for a raw element belonging to another window, which is exactly the population this util targets. It works today only because Clarity components always store an `ElementRef` (a same-realm class), but the public input path has the same hole: the `contentOrigin` setter in `popover-content.ts` also does `origin instanceof Element`, so a consumer passing a raw foreign element via `clrPopoverContentOrigin` bypasses the `ElementRef` wrap *and* this fix. Duck-typing is realm-proof:

```suggestion
function getOriginElement(origin: FlexibleConnectedPositionStrategyOrigin): Element | null {
  if (origin instanceof ElementRef) {
    return origin.nativeElement;
  }

  // `instanceof Element` is realm-sensitive and misses raw elements belonging to
  // another window - exactly the population this util targets - so duck-type instead.
  if ((origin as Node)?.nodeType === Node.ELEMENT_NODE) {
    return origin as Element;
  }

  return null;
}
```

(Same change recommended for the `contentOrigin` setter, which is outside this diff.)

---

projects/angular/popover/common/utils/cross-window-origin.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    if (!frameElement) {
+      return null;
+    }
+
+    const frameRect = frameElement.getBoundingClientRect();
+    const frameEl = frameElement as HTMLElement;
+    offsetX += frameRect.left + frameEl.clientLeft;
+    offsetY += frameRect.top + frameEl.clientTop;
```
dtsanevmw: Minor accuracy nit: the embedded document renders inside the frame's **content box**, but `clientLeft`/`clientTop` only cover borders — an iframe with CSS padding produces a skewed anchor. Reading computed style (through the frame's own window, for realm-safety) covers both:

```suggestion
    const frameRect = frameElement.getBoundingClientRect();
    const frameWindow = frameElement.ownerDocument?.defaultView;
    const { borderLeftWidth, borderTopWidth, paddingLeft, paddingTop } = frameWindow.getComputedStyle(frameElement);
    offsetX += frameRect.left + (parseFloat(borderLeftWidth) || 0) + (parseFloat(paddingLeft) || 0);
    offsetY += frameRect.top + (parseFloat(borderTopWidth) || 0) + (parseFloat(paddingTop) || 0);
```

Take or leave — if you'd rather keep `clientLeft`, a one-line comment noting the padding limitation is enough.

The overall fallback contract here is exactly right, by the way: `null` → untranslated rect is precisely what CDK would compute if handed the element directly, and the `try/catch` around `frameElement` is necessary and correctly scoped.

---

projects/angular/popover/common/popover-content.spec.ts
```diff
@@ -160,6 +160,58 @@ export default function (): void {
...
+
+        expect(this.popoverService.open).toBe(true);
+
+        iframeDoc.dispatchEvent(new Event('pointerdown', { bubbles: true }));
+
+        expect(this.popoverService.open).toBe(false);
+
+        iframe.remove();
+      });
+    });
```
dtsanevmw, line 405: Three notes on the new specs:

1. **This test encodes the `pointerdown` behavior rather than catching the bug it introduces** (see my comment on `createCrossWindowOutsideClickSubscription`): with the capture-phase `click` fix, this should dispatch `new Event('click', { bubbles: true })` instead.
2. **Missing regression test for the toggle case** — the one that would have caught the close-then-reopen bug: give the foreign trigger a click listener that calls `popoverService.toggleWithEvent(event)` (mimicking `ClrDropdownTrigger`), open via a first click, then dispatch a second click on the trigger and assert the popover ends **closed** and the listener was **not** invoked (capture-phase `stopPropagation()` must suppress it).
3. **Cleanup leaks on failure**: `iframe.remove()` at the end of the `it` body doesn't run if an expectation throws, leaking the iframe into subsequent specs. `cross-window-origin.spec.ts` handles this correctly with `cleanupFns` + `afterEach` — same pattern applies here (and to the `getScrollableParents` spec above).

---

dtsanevmw (changes requested):
## Review summary

This PR correctly diagnoses and fixes a real crash (CDE-3155): the realm-sensitive `instanceof HTMLHtmlElement` check lets the ancestor walk fall through to a foreign `Document`, and `getComputedStyle(document)` throws. The analysis in the description is accurate, the new util is correctly kept out of the public barrel, the service's `origin` is never mutated (only a local positioning copy is wrapped), naming and file placement follow existing conventions, and the spec coverage of the positioning util is genuinely thorough. The doc comments explain *why*, not *what* — including the correct and non-obvious zone.js note.

**Requesting changes for one functional bug** (inline comment on `createCrossWindowOutsideClickSubscription`): the cross-window outside-click listener closes on bubble-phase `pointerdown`, so re-clicking the toggle button closes the popover and the trigger's own `click` handler immediately reopens it — the toggle can never close a cross-window popover. CDK's dispatcher uses capture-phase `click` listeners, which is what makes the same-window `stopPropagation()` dance work; mirroring that (suggestion inline) fixes it. The new spec currently encodes the `pointerdown` behavior rather than catching this.

Strongly recommended before merge (suggestions inline):
- **Performance**: the live Point getters cost ~10 forced layout reads per reposition (CDK reads `x`/`y` twice each per `apply()`); a microtask-scoped memo collapses this to one.
- **Robustness**: two spots in the fix itself rely on realm-sensitive `instanceof` (`getOriginElement`, and the `contentOrigin` input setter has the same pre-existing hole); duck-typing on `nodeType` is realm-proof. The new `getScrollableParents` guard deliberately leans on the cross-realm `instanceof` failure to stop the walk — that intent deserves a code comment, and the resulting loss of scroll tracking inside the iframe (which the demo papers over with 200 ms polling) deserves either a follow-up ticket or the realm-safe-walk alternative described inline.

Take-or-leave polish: iframe padding in the frame-offset math, a `parentElement?.` hardening, and spec cleanup via `afterEach`.

---

projects/angular/popover/common/popover-content.ts
```diff
@@ -453,3 +539,30 @@ export class ClrPopoverContent implements OnDestroy, AfterViewInit {
...
+
+// Same-origin-safe: returns null (rather than throwing) if `frameElement` can't be
+// read, which happens when `win` is embedded in a cross-origin parent window.
+function getFrameElement(win: Window): Element | null {
+  try {
+    return win.frameElement;
+  } catch {
+    return null;
+  }
+}
```
dtsanevmw, line 639: These can be moved into utils

idonchev-vmw:
Thanks — this was a genuinely valuable review. All points addressed:

**The capture-phase bug (blocking):** you were right, and it mattered more than it first looked. Fixed in b55d04a36's predecessor `3b83d4232` by mirroring CDK's dispatcher exactly (capture-phase `click`/`auxclick`). Worth noting why this was nearly shipped-broken: trigger-click is currently the *only* way affected users can dismiss a popover, so the flicker-reopen would have taken away their last working interaction. Added the toggle regression test you described, and verified it fails against the old bubble-phase implementation.

**Realm-sensitive `instanceof` in the fix itself:** took the deeper option rather than documenting the trade-off. `getScrollableParents` no longer uses `instanceof` at all — `nodeType`/`tagName` shape checks, computed style resolved through each node's own window, and the walk continues across the frame chain instead of stopping at the boundary. That removes the scroll-tracking gap entirely, so the demo's 200 ms polling workaround is gone too. `getOriginElement` and the `contentOrigin` setter got the same duck-typing treatment.

**Performance:** microtask memo applied as suggested — one layout read per reposition instead of ~6, still live across repositions. Nice suggestion; it resolved the earlier snapshot-vs-getters tension cleanly rather than trading one off against the other.

**Polish:** `parentElement?.`, iframe padding in the offset math, and `afterEach` cleanup all applied.

**One counter-proposal**, which Daniel and I agreed on offline: rather than adding `isCrossWindowOrigin()` at the two call sites, the check is now `getCrossWindowOriginContext()` used *inside* both functions. Both already early-returned for same-window origins, so a call-site guard would have re-run the identical condition and duplicated it in two places; returning the element+window pair instead of a boolean also avoids the outside-click path having to derive the window twice. Same intent, single definition. The contract is now enforced by a test asserting a same-window origin registers zero extra listeners (mutation-checked: removing the guard makes it fail).

---

Two findings from investigating this that changed the picture, both now in the PR description:

1. **The ticket's StackBlitz is invalid** — it pins `@clr/angular` **17**.9.0 (the `18.2.10` everywhere is `@angular/*`), so it reproduces the *old* pre-CDK `Popover` class crash that 18.0.0 already deleted, and it has no iframe. The real bug needs the realm boundary.
2. **A second, realm-independent crash trigger**: a plain `DocumentFragment` in the ancestor chain — it shares `ShadowRoot`'s `nodeType` but has no `host`, so the old `instanceof ShadowRoot` check missed it. Found by sweeping candidate structures against a verbatim copy of the pre-fix code; now covered by a test.

Suite is 3483/3483. Every fix was verified by reverting it and confirming the matching test fails.

---
_Generated by [Claude Code](https://claude.com/claude-code)_

idonchev-vmw:
Update for @dtsanevmw: pushed one more commit (`213a62b60`) since your last review, found while testing the demo manually.

**New: Escape doesn't close a cross-window popover.** Same root cause class as the outside-click fix, just for keyboard: CDK's `OverlayKeyboardDispatcher` attaches a single `keydown` listener on this window's `document.body` at bootstrap, so once a popover's origin (and focus) sits inside a relocated cross-realm iframe, an Escape press never bubbles out of that foreign document and CDK's handler never sees it. Fixed with `createCrossWindowEscapeSubscription()`, mirroring `createCrossWindowOutsideClickSubscription()` - listens for Escape on the origin's own document too when it differs from this window's.

Also added a Signpost example to the shadow-dom-crash-repro demo alongside the existing Dropdown one, since Signpost is one of the few popover consumers that leaves `scrollToClose` at its default `false` - useful for eyeballing that scroll-follow behavior survives the cross-realm relocation too.

Covered by 3 new mutation-tested specs; full suite still green (3486/3486). Should be ready for a fresh look whenever you and Fable have time.

projects/angular/popover/common/popover-content.ts
idonchev-vmw: Done in `f70476543` — moved into a new `utils/dom-realm.ts` alongside `cross-window-origin.ts`, same as the other cross-window helpers. Pure move, no behavior change; not in the public barrel.

idonchev-vmw:
Found and fixed another one via a live repro, cc @dtsanevmw: commit `717565409`. `handleOutsideClick`'s toggle-button check guarded `openEvent` but not `openEvent.target`, which can be `null` - the unguarded `.contains()` call threw before `closePopover()` ran, so outside clicks silently did nothing. Fixed with optional chaining, plus a regression test (mutation-tested against the exact error from the live repro). Suite still 3487/3487 green.

projects/angular/popover/common/utils/cross-window-origin.ts
```diff
@@ -0,0 +1,149 @@
...
+  let currentWindow: Window | null = originWindow;
+
+  while (currentWindow && currentWindow !== window) {
+    let frameElement: Element | null;
+
+    try {
+      frameElement = currentWindow.frameElement;
+    } catch {
+      return null;
+    }
```
dtsanevmw: You can use this util from dom-realm
```suggestion
    const frameElement: Element | null = getFrameElement(currentWindow);
```

projects/angular/popover/common/popover-content.ts
dtsanevmw: Merge both escape subscriptions into one

projects/angular/popover/common/popover-content.ts
```diff
@@ -224,6 +231,11 @@ export class ClrPopoverContent implements OnDestroy, AfterViewInit {
         ? this.createPointBasedOutsideClickSubscription()
```
dtsanevmw, line 226: Make sure point based outside click also works in cross window

projects/angular/popover/common/utils/cross-window-origin.ts
idonchev-vmw: Done in 96e432a31 - reuses getFrameElement from dom-realm now.

projects/angular/popover/common/popover-content.ts
idonchev-vmw: Done in 96e432a31 - merged into a single createEscapeSubscription() that merges the streams and handles Escape once.

projects/angular/popover/common/popover-content.ts
idonchev-vmw: Done in 96e432a31 - added createCrossWindowPointOutsideClickSubscription(), mirroring the element-based one but keyed off pointTargetElement, plus a regression test.

dtsanevmw (dismissed):

dtsanevmw (approved):

PR #2606, opened 2026-08-12
feat(typography): add break-word modifier to cds-text attribute

idonchev-vmw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The `[cds-text]` attribute has no modifier for word-breaking. A long word with no natural break point (no spaces/hyphens) never wraps, so it overflows its container once the container is narrower than the word.

Issue Number: [CDE-2952](https://vmw-jira.broadcom.net/browse/CDE-2952)

## What is the new behavior?

- Adds a `break-word` value for the `[cds-text]` attribute: `overflow-wrap: anywhere; word-break: normal;` — the same pairing already used for long-word wrapping in `combobox`, `wizard`, `stack-view`, and `timeline`.
- It only breaks a word mid-way as a last resort, when that single word alone is too long to fit on one line; ordinary multi-word content keeps wrapping at spaces exactly as before.
- Adds a Storybook example (`Layout/Typography` → `Transforms`) and a website docs table row demonstrating the new modifier, using a narrow, bordered demo box (Clarity design tokens for border/padding) so the wrap is actually visible rather than just described.

- [ ] Yes
- [x] No

## Other information

This was originally expected to also close [CDE-2953](https://vmw-jira.broadcom.net/browse/CDE-2953) (Datagrid cells overlapping on column resize), but `clr-dg-cell` never applies the `cds-text` attribute to its projected content, so this modifier alone doesn't reach default grid cell content. CDE-2953 is fixed directly in the Datagrid instead, as a separate PR: #2609.


projects/website/src/app/documentation/demos/typography/typography.demo.html
```diff
@@ -612,6 +612,18 @@ <h3 cds-text="section disable-lhe" cds-layout="m-t:xl">Transforms, Positions, Li
             </ul>
           </td>
         </tr>
+        <tr>
+          <td class="left" cds-layout="p-t:lg p-b:lg">
+            <p cds-text="body break-word" style="max-width: 180px; border: 1px dashed #999; padding: 4px">
```
valentin-mladenov: Let's use clarity tokens, wherever possible.

```suggestion
            <p cds-text="body break-word" style="max-width: 180px; border: var(--cds-alias-object-border-width-100) dashed var(--cds-alias-object-border-color); padding: var(--clr-base-vertical-offset-s) var(--clr-base-horizontal-offset-s);">
```

.storybook/stories/layout/typography.stories.ts
```diff
@@ -119,6 +119,9 @@ const TransformsTemplate: StoryFn = () => ({
       <p cds-text="body truncate" cds-layout="container:xs">
         Text Truncation: The quick brown fox jumps over the lazy dog. (truncate)
       </p>
+      <p cds-text="body break-word" style="max-width: 180px; border: 1px dashed #999; padding: 4px">
```
valentin-mladenov: Let's use clarity tokens, wherever possible.

```suggestion
            <p cds-text="body break-word" style="max-width: 180px; border: var(--cds-alias-object-border-width-100) dashed var(--cds-alias-object-border-color); padding: var(--clr-base-vertical-offset-s) var(--clr-base-horizontal-offset-s)">
```

projects/angular/styles/core/typography/_typography.scss
```diff
@@ -271,6 +271,13 @@ body[cds-text*='body'] {
   @include mixins.remove-line-height-erasers;
 }
 
+// word breaking
+
+[cds-text~='break-word'] {
+  overflow-wrap: anywhere !important;
+  word-break: normal !important;
```
valentin-mladenov, line 278: Why not using only [word-break: break-word](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/word-break#break-word)?


```suggestion
  word-break: break-word !important;
```

valentin-mladenov (commented):
Check my comments

projects/angular/styles/core/typography/_typography.scss
idonchev-vmw: It's deprecated.

https://www.w3.org/TR/css-text-3/#word-break-property

For compatibility with legacy content, the [word-break](https://www.w3.org/TR/css-text-3/#propdef-word-break) property also supports a deprecated break-word keyword. When specified, this has the same effect as word-break: normal and [overflow-wrap: anywhere](https://www.w3.org/TR/css-text-3/#propdef-overflow-wrap), regardless of the actual value of the overflow-wrap property.

projects/website/src/app/documentation/demos/typography/typography.demo.html
idonchev-vmw: done

.storybook/stories/layout/typography.stories.ts
idonchev-vmw: done

valentin-mladenov:
Use standard Clarity PR description template.

valentin-mladenov (approved):

PR #2609, opened 2026-08-13
fix(datagrid): break long unbreakable words in cells instead of overflowing

idonchev-vmw:
## Summary
- Fixes CDE-2953: a datagrid cell containing a long word with no natural break point (no spaces/hyphens) overflows into neighboring cells once its column is resized narrower than the word, when `datagrid-overflow-ellipsis` is not used.
- Root cause: `.datagrid-cell` sets no `overflow-wrap`/`word-break` of its own, so it falls back to the browser default, which never breaks mid-word.
- Fix: add `overflow-wrap: anywhere; word-break: normal;` to the base `.datagrid-cell` rule — the same pairing already used for long-word wrapping in `combobox`, `wizard`, `stack-view`, and `timeline`. It only breaks a word mid-way as a last resort, when that single word alone is wider than the cell; ordinary multi-word content keeps wrapping at spaces exactly as before.
- Scoped to `.datagrid-table .datagrid-cell`, confirmed this does not reach the separate, hidden `.datagrid-calculation-table` used for auto-sizing column widths (they're sibling elements, not nested).
- Remains mutually exclusive with the opt-in `.datagrid-overflow-ellipsis` class: that class's `white-space: nowrap` suppresses all wrapping (including this fallback), so single-line truncation still works as before when explicitly requested.

Note: this was originally going to depend on CDE-2952 (a `cds-text` `break-word` modifier, see #2606) — but `clr-dg-cell` never applies `cds-text` to its content, so that alone wouldn't have fixed default cell content. This PR fixes the datagrid cell directly instead; #2606 ships the `cds-text` utility separately for consumers who want it on their own markup.

## Test plan
- [x] Added a `LongWordWrap` Storybook story (`Datagrid/Datagrid`) with a narrow 100px column containing a long unbroken word, demonstrating it wraps inside the cell instead of overflowing
- [x] Ran the full `clr-angular` datagrid test suite: 635/635 passing
- [x] `stylelint`, `eslint`, `prettier --check`, and the story-template formatter all clean on changed files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.storybook/stories/datagrid/datagrid.stories.ts
+++ b/.storybook/stories/datagrid/datagrid.stories.ts
@@ -69,6 +69,7 @@ export default {
     expandable: false,
     compact: false,
     overflowEllipsis: false,
+    longWordWrap: false,
     hidableColumns: false,
     showActions: false,
     height: 0,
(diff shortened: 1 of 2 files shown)


idonchev-vmw:
Retargeted onto `next`

idonchev-vmw:
Closing on behalf of https://github.com/vmware-clarity/ng-clarity/pull/2606
We prefer to manually control this behavior by using a cds-text word breaking

PR #2625, opened 2026-08-19
fix(icon): avoid copying the full icon registry on every icon render

dtsanevmw:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`ClarityIcons.registry` (`icon.service.ts`) is a getter that spreads the entire icon registry into a brand-new object on every access:

```ts
static get registry(): Readonly<IconRegistry> {
  return { unknown: unknownIcon[1] as string, ...GlobalStateService.state.iconRegistry };
}
```

`ClrIcon` used this getter as a single-key lookup (`ClarityIcons.registry[this.shape]`) in its hot render paths: `updateIcon()` (called on init and every input change), the global icon-registry update subscription, and `getIconSVG()`. Each of these calls copies the *entire* registry just to read one key.

With many `<clr-icon>`/`<cds-icon>` instances on screen at once (e.g. a `clr-datagrid` with many rows), this results in a full-registry copy per icon per render — unnecessary O(registry size) allocation and CPU work that scales with both registry size and icon count, showing up as GC pressure and sluggish/frozen UI when icons render/re-render at scale.

Issue Number: N/A

## What is the new behavior?

Added `ClarityIcons.getIconShape(name)`, an O(1) lookup that reads a single key directly from the underlying registry state without copying it, preserving the exact same fallback semantics as `registry[name]` (including the built-in `unknown` icon fallback).

Switched the hot-path call sites to use it instead of `registry[name]`:
- `ClrIcon.updateIcon()`
- `ClrIcon`'s icon-registry update subscription (`ngOnInit`)
- `getIconSVG()` (`icon.svg-helpers.ts`)
- `ClarityIcons.addAliases()`

The public `ClarityIcons.registry` getter is unchanged and still available for consumers who want a full registry snapshot.

Added regression tests asserting the `registry` getter is not invoked during icon rendering/lookup, plus unit tests for `getIconShape`'s lookup and fallback behavior.

- [ ] Yes
- [x] No

## Other information


projects/angular/icon/icon.service.ts
```diff
@@ -29,6 +29,21 @@ export class ClarityIcons {
...
+   * rendering: `registry` spreads the whole (potentially large) registry into a new object on
+   * every access, which becomes expensive/memory-churning when many icons render or re-render.
+   */
+  static getIconShape(name: string): IconRegistrySources[string] | undefined {
+    const iconRegistry = GlobalStateService.state.iconRegistry;
+    if (name in iconRegistry) {
+      return iconRegistry[name];
+    }
+    return name === 'unknown' ? (unknownIcon[1] as string) : undefined;
+  }
```
valentin-mladenov, line 45: Since the `unknown` icon is not in regestry, then it make sense not to search for it there first.

```suggestion
  static getIconShape(name: string): IconRegistrySources[string] | undefined {
    if (name === 'unknown') {
        return unknownIcon[1] as string;
    }

    const iconRegistry = GlobalStateService.state.iconRegistry;
    if (name in iconRegistry) {
      return iconRegistry[name];
    }

    return undefined;
  }
```

valentin-mladenov (commented):
Check my comment.

You can update all the unit tests to use the new method to reduce the footprint of `registry` internally.

projects/angular/icon/icon.service.ts
dtsanevmw: On reflection, decided not to take this — reverted in b0ebd49b3.

Checking `name === 'unknown'` before the registry means a consumer who registered their own `unknown` shape via `addIcons(['unknown', ...])` would silently stop getting it back from `getIconShape('unknown')`, always getting the built-in fallback instead. Nothing in this codebase's tests/docs/demos exercises that today, but it's a real behavior change to the public API's contract, not just an internal micro-optimization, and I'd rather not make that call silently in a perf-cleanup PR. Kept the registry-first lookup so overriding 'unknown' still works as before.

dtsanevmw:
Updated the icon.service unit tests to use getIconShape() instead of indexing into registry directly (fecfa658c), so the test suite no longer exercises the full-registry-copy path either.

valentin-mladenov (approved):

PR #2670, opened 2026-09-01
feat(datagrid): column actions

valentin-mladenov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-3014

- [ ] Yes
- [x] No


projects/angular/data/datagrid/datagrid-column-action.ts
```diff
@@ -0,0 +1,97 @@
...
+  // A disabled item still receives the keydown, so it has to swallow it before it turns into a click.
+  @HostListener('keydown.space', ['$event'])
+  @HostListener('keydown.enter', ['$event'])
+  protected onActivate(event: KeyboardEvent) {
+    if (this.disabled) {
+      event.preventDefault();
+      event.stopImmediatePropagation();
+    }
+  }
+}
```
dtsanevmw, line 97: Proposal: make this a subclass of `ClrDropdownItem` instead of re-implementing the item. It keeps the `clrDgColumnAction` selector, inherits `clrDisabled`, the host bindings, the click/keydown handling and the arrow-key registration, and only adds what a column action needs on top (re-anchoring the menu after an action that keeps it open).

```suggestion
import { Directive, HostListener, inject } from '@angular/core';
import { ClrDropdownItem } from '@clr/angular/popover/dropdown';
import { BASIC_FOCUSABLE_ITEM_PROVIDER } from '@clr/angular/utils';

import { ClrDatagridColumnActions } from './datagrid-column-actions';

/**
 * A `clrDropdownItem` for a `clr-dg-column-actions` menu. It is the dropdown item - same inputs,
 * styling, arrow key order and close-on-click behaviour - with what a column action needs on top.
 *
 * An item that keeps the menu open (`[clrCloseMenuOnClick]="false"`) re-anchors it after the click.
 * The menu is anchored to a trigger that travels with the column, so an action that moves the column
 * would otherwise leave the menu hanging where the column used to be. The built-in Pin Column action
 * behaves the same way.
 *
 * A plain `clrDropdownItem` can be projected as well; it only lacks the re-anchoring.
 */
@Directive({
  selector: '[clrDgColumnAction]',
  // Providers are not inherited from ClrDropdownItem, and without its own FocusableItem the item
  // would resolve the menu's DropdownFocusHandler instead.
  providers: [BASIC_FOCUSABLE_ITEM_PROVIDER],
  standalone: false,
})
export class ClrDatagridColumnAction extends ClrDropdownItem {
  private readonly columnActions = inject(ClrDatagridColumnActions);

  @HostListener('click')
  protected onColumnActionClick(): void {
    if (!this.disabled && this.closeMenuOnClick === false) {
      this.columnActions.repositionMenu();
    }
  }
}
```

This does not compile on its own; it needs two things that are not in the PR yet:

1. `ClrDatagridColumnActions` has to *be* the dropdown (extend `ClrDropdown` and provide itself under the `ClrDropdown` token) so a projected item can inject it. Today the `clr-dropdown` lives inside the component's template and is not on the injector path of projected nodes, which is the reason this directive exists in its current form.
2. `ClrDropdownItem` needs a per-item `clrCloseMenuOnClick` input (falling back to the dropdown's `clrCloseMenuOnItemClick`), so the built-in pin/filter items can stay open while everything else closes. That also removes the `[clrCloseMenuOnItemClick]="false"` on the menu and the `closeMenu()` method, which currently can never close the menu because it early-returns on `!isMenuClosable`.

The full working change, with the PR's suites green (core 3587, addons 1315), is on `claude/datagrid-column-actions-review-6n7y5k`: https://github.com/vmware-clarity/ng-clarity/compare/vmladenov/datagrid-column-actions_cde-3014...claude/datagrid-column-actions-review-6n7y5k

dtsanevmw (commented):
One suggestion on `datagrid-column-action.ts`: derive it from `ClrDropdownItem` rather than re-implementing the item. Details and the dependencies it needs are in the inline comment.

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -797,11 +816,17 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
 
   onColumnOrderChange(data: ColumnOrderChanged) {
     this.columns = data.columns;
-    this.visibleColumns = this.columns.filter((column: ColumnDefinition<T>) => !column.hidden);
-    this.cdr.detectChanges();
-    //Without resize when the grid is empty and column is moved
-    //the columns are not displayed correctly
-    this.resize();
+
+    if (this.columns.some((column: ColumnDefinition<T>) => column.pinned)) {
```
dtsanevmw, line 820: A pinned column that is hidden is not rendered in either container (hidden columns are left out of `visibleColumns`), so it does not stand in the way of the ordinary reorder. As written, a grid whose only pinned column is hidden still destroys and recreates every column view and every cell on each move or drop.

```suggestion
    // Only a pinned column that is actually rendered sits in the sticky container; a hidden one
    // is in neither container and does not stand in the way of the ordinary reorder.
    if (this.columns.some((column: ColumnDefinition<T>) => column.pinned && !column.hidden)) {
```

Verified with a spec: with a pinned-and-hidden third column, "Move Right" reorders through the ordinary path and `rebuildColumnViews` is not called. One thing to keep in mind for the spec: the definitions must be bound as a new array with `hidden` already set, since mutating `hidden` on an already-bound definition does not recompute `visibleColumns` and the column then stays rendered (and the move hits the `insertBefore` error the comment above describes).

projects/addons/datagrid/datagrid.component.html
```diff
@@ -99,6 +95,108 @@
...
+          column offers one way to reach its filter rather than two.
+        -->
+        <clr-dg-column-actions [clrDgKeepFilterInHeader]="true">
+          @if (
+            column.sortComparator ||
+            column.sortAndFilterByField ||
+            column.pinnable ||
+            column.stringFilter ||
+            column.filter
+          ) {
```
dtsanevmw, line 112: This divider is meant to separate the built-in items from the move actions, but with `clrDgKeepFilterInHeader` the filter item is never in this menu, so `column.stringFilter || column.filter` puts a separator above nothing. A column that has a filter and can neither sort nor pin opens a menu whose first child is `role="separator"`, followed by Move Left / Move Right.

```suggestion
          <!--
            Only what the menu renders above counts: sort actions for a sortable column and the pin
            action for a pinnable one. The filter stays in the header here, so it adds nothing.
          -->
          @if (column.sortComparator || column.sortAndFilterByField || column.pinnable) {
```

Reproduced with `{ displayName: 'Only filter', field: 'name', stringFilter: { accepts: () => true } }`: the menu's `firstElementChild` has `role="separator"`.

projects/angular/data/datagrid/datagrid-column-actions.ts
```diff
@@ -0,0 +1,394 @@
...
+          </button>
+          <button
+            type="button"
+            clrDropdownItem
+            [class.active]="sortOrder === ClrDatagridSortOrder.DESC"
+            (click)="sort(true)"
+          >
+            <cds-icon shape="arrow" direction="down" aria-hidden="true"></cds-icon>
+            {{ commonStrings.keys.sortColumnDescending }}
+          </button>
```
dtsanevmw, line 100: The `active` class is visual only: both items are announced as plain `menuitem`s, so a screen reader user cannot tell which direction is applied (the column header's `aria-sort` is not read while focus is inside the menu). The two directions are exclusive settings, which is what `menuitemradio` + `aria-checked` is for.

```suggestion
          <button
            type="button"
            clrDropdownItem
            role="menuitemradio"
            [attr.aria-checked]="sortOrder === ClrDatagridSortOrder.ASC"
            [class.active]="sortOrder === ClrDatagridSortOrder.ASC"
            (click)="sort(false)"
          >
            <cds-icon shape="arrow" direction="up" aria-hidden="true"></cds-icon>
            {{ commonStrings.keys.sortColumnAscending }}
          </button>
          <button
            type="button"
            clrDropdownItem
            role="menuitemradio"
            [attr.aria-checked]="sortOrder === ClrDatagridSortOrder.DESC"
            [class.active]="sortOrder === ClrDatagridSortOrder.DESC"
            (click)="sort(true)"
          >
            <cds-icon shape="arrow" direction="down" aria-hidden="true"></cds-icon>
            {{ commonStrings.keys.sortColumnDescending }}
          </button>
```

One dependency: `ClrDropdownItem` sets `[attr.role]="'menuitem'"` as a host binding, and a host binding runs after the template's static attribute, so the `role` here would be overwritten. The small change that makes it work is to let the directive take the role from the template:

```ts
// dropdown-item.ts
host: { '[attr.role]': 'role', ... }
/** `menuitem` unless the template says otherwise, e.g. `menuitemradio`. */
@Input() role = 'menuitem';
```

With that in place a plain `role="menuitemradio"` attribute is honoured and every existing `clrDropdownItem` keeps `menuitem`. `aria-checked` needs nothing extra.

projects/angular/data/datagrid/datagrid-column-actions.ts
```diff
@@ -0,0 +1,394 @@
...
+        }
+
+        @if (hasFilter && !keepFilterInHeader) {
+          @if (column.sortable || column.pinnable) {
+            <div class="dropdown-divider" role="separator"></div>
+          }
+          <button type="button" #trigger clrDropdownItem (click)="openFilter($event)">
+            <cds-icon [shape]="filterActive ? 'filter-grid-circle' : 'filter-grid'" solid aria-hidden="true"></cds-icon>
+            {{ commonStrings.keys.filterColumn }}
+          </button>
```
dtsanevmw: The header toggle this item replaces carried `aria-expanded` and `aria-controls` (see `datagrid-filter.ts`), and this item carries neither, so nothing tells assistive technology that it opens a dialog or which one. Now that the `active` class is gone too, the item has no state at all.

```suggestion
          <button
            type="button"
            #trigger
            clrDropdownItem
            aria-haspopup="dialog"
            [attr.aria-expanded]="filterOpen"
            [attr.aria-controls]="filterPopoverId"
            (click)="openFilter($event)"
          >
            <cds-icon [shape]="filterActive ? 'filter-grid-circle' : 'filter-grid'" solid aria-hidden="true"></cds-icon>
            {{ commonStrings.keys.filterColumn }}
          </button>
```

Two small additions go with it: a `protected get filterOpen()` returning `this.columnPopover.open`, and a `protected get filterPopoverId()` returning `this.columnActions.filter()?.popoverId ?? null`, with `popoverId?: string` added to `ColumnFilterHandle` (`ClrDatagridFilter` already has a public `popoverId`, so nothing else changes).

dtsanevmw (commented):
Three smaller findings from the review, each with a suggestion inline: the rebuild guard also firing for a hidden pinned column, a leading separator in the appfx menu for a filter-only column, and the missing ARIA state on the sort and filter items. Each is verified with a spec on `dtsanevmw/datagrid-column-actions-fixes_cde-3014` (commits `2f8e026`, `cba9b17`, `e549346`) if you want to pull the tests along with the change.

dtsanevmw:
Please create a separate PR for the new filter Icon change as this affects default filters and it's good to be individual

valentin-mladenov:
> Please create a separate PR for the new filter Icon change as this affects default filters and it's good to be individual

It doesn't affect default filters only the column actions. The new Icon is directly and only linked to the column actions.

projects/angular/data/datagrid/datagrid-column-actions.ts
valentin-mladenov: Corrected in commit: 800cbb7

projects/angular/data/datagrid/datagrid-column-actions.ts
valentin-mladenov: Corrected in commit: 800cbb7

projects/angular/data/datagrid/datagrid-column-action.ts
valentin-mladenov: Clr column actions component extends Clr dropdown and clr column action extends clr dropdown item. 
Commit 7289d98

projects/addons/datagrid/datagrid.component.html
valentin-mladenov: Corrected in commit: f73ae4a.

projects/addons/datagrid/datagrid.component.ts
valentin-mladenov: Corrected in commit: d64b84d

projects/angular/data/datagrid/datagrid-column-actions.ts
```diff
@@ -0,0 +1,465 @@
...
+   * first means it lands correctly even when the action opens a modal.
+   *
+   * Checks `isMenuClosable` for the same reason `clrDropdownItem` does on its own click handler:
+   * without it a projected item would close the menu on click while a built-in one, which this
+   * component keeps open, would not.
+   */
+  closeMenu() {
+    if (!this.isMenuClosable) {
+      return;
+    }
```
dtsanevmw, line 369: If that is always false as set in the constructor, then `canClosePopover` would not work if set to true

projects/angular/data/datagrid/datagrid-column.ts
```diff
@@ -53,23 +54,6 @@ import { WrappedColumn } from './wrapped-column';
...
-          <button
-            class="datagrid-column-pin"
-            type="button"
-            [attr.aria-label]="pinned ? commonStrings.keys.unpinColumn : commonStrings.keys.pinColumn"
-            (click)="togglePinned($event)"
-          >
-            <cds-icon [size]="'12'" [shape]="pinned ? 'unpin' : 'pin'" solid aria-hidden="true"></cds-icon>
-          </button>
-        }
-      -->
```
dtsanevmw: Deleted by accident

projects/angular/data/datagrid/datagrid-column.ts
```diff
@@ -132,19 +126,15 @@ export class ClrDatagridColumn<T = any>
...
-  //  * Shows a pin toggle in the column header, letting the user pin and unpin the column from within
-  //  * the datagrid. It only adds the control - the pinned state itself stays on `clrDgPinned`.
-  //  */
-  // @Input({ alias: 'clrDgPinnable', transform: booleanAttribute }) pinnable = false;
+  /**
+   * Lets the user pin and unpin the column from within the datagrid, through the pin action in
+   * `clr-dg-column-actions`. It only offers the control - the pinned state itself stays on
+   * `clrDgPinned`.
+   */
+  @Input({ alias: 'clrDgPinnable', transform: booleanAttribute }) pinnable = false;
```
dtsanevmw, line 134: For the normal `clr-datagrid` if that is set to true, it won't do anything if there are no column actions provided by the user.

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -1134,6 +1195,33 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
...
+   */
+  private rebuildColumnViews(): void {
+    this.visibleColumns = [];
+    this.cdr.detectChanges();
+    this.visibleColumns = this.columns.filter((column: ColumnDefinition<T>) => !column.hidden);
+    this.cdr.detectChanges();
+    // The columns are measured from scratch, and this also covers an empty grid, where the datagrid
+    // does not re-render the columns on its own.
+    this.resize();
+  }
```
dtsanevmw, line 1223: Not needed to move pinned columns for now so we can drop this

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -797,11 +816,19 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
...
-    this.cdr.detectChanges();
-    //Without resize when the grid is empty and column is moved
-    //the columns are not displayed correctly
-    this.resize();
+
+    // Only a pinned column that is actually rendered sits in the sticky container; a hidden one is
+    // in neither container and does not stand in the way of the ordinary reorder.
+    if (this.columns.some((column: ColumnDefinition<T>) => column.pinned && !column.hidden)) {
+      this.rebuildColumnViews();
+    } else {
```
dtsanevmw, line 824: This won't be needed if we drop rebuildColumnViews

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -1067,6 +1103,31 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
     }
   }
 
+  /**
+   * The actions to offer in one column's menu: the grid-wide ones first, then the column's own, so a
+   * column adds to the shared set rather than reordering it.
+   */
+  protected getColumnActions(column: ColumnDefinition<T>): ActionDefinition[] {
+    return [...(this.columnActions || []), ...(column.actions || [])];
```
dtsanevmw, line 1111: nit: easier to read
```suggestion
const actions = [...(this.columnActions ?? [])];
if (column.actions?.length) {
  actions.push(...column.actions);
}
return actions;
```

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -1011,6 +1042,11 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
   }
 
   protected onFilterChange(filterValue: unknown, column: ColumnDefinition<T>): void {
+    // The column definition is what the filter is bound to, so the active value is kept there the
+    // same way the active sort order is. Otherwise reordering a column, which rebuilds the column
+    // views, would bind the filter back to the value it started with and silently drop the filter
+    // the user typed.
+    column.defaultFilterValue = filterValue;
```
dtsanevmw, line 1049: This write-back is only there for `rebuildColumnViews`, and it doesn't actually fix that case, so it can go together with the rebuild:

```suggestion
  protected onFilterChange(filterValue: unknown, column: ColumnDefinition<T>): void {
```

- Without the rebuild a moved column keeps its filter anyway: the filter component isn't recreated, so the typed value and the filtered rows both stay.
- With the rebuild it only restores the text: the filters read `filterValue` once on creation, so the input shows the value again but it is never applied - all rows come back while the input still shows the filter.
- It also shows up after hiding and re-showing a column: the old value is back in the input, but the rows aren't filtered.
- It changes what consumers get: `defaultFilterValue` (documented as the default) is overwritten with the current value, and `appfx-datagrid-filter` stops emitting `columnFilterChange` for keyups that don't change the value, because it compares against its `filterValue` input.

The spec "keeps an applied filter value when a column is moved" only checks `defaultFilterValue` on the definition, so it passes while the rows are unfiltered - it can be dropped with this.

---

projects/addons/datagrid/datagrid.component.ts
```diff
@@ -947,16 +974,20 @@ export class DatagridComponent<T> implements OnInit, OnDestroy, AfterViewInit, O
   }
 
   protected onColumnResize(columnSize: number, column: ColumnDefinition<T>): void {
+    // The column definition is what a rebuild reads to recreate the view - see rebuildColumnViews -
+    // so the resized width has to be written back here the same way sort and filter state is.
+    // Otherwise pinning a column and then moving or dragging any column, which rebuilds every column
+    // view, would silently discard every resize made before that pin.
+    column.width = `${columnSize}px`;
```
dtsanevmw, line 981: Same as the filter write-back: for moves this is only needed because of `rebuildColumnViews`, so it can go with it.

```suggestion
  protected onColumnResize(columnSize: number, column: ColumnDefinition<T>): void {
```

- Without the rebuild a resized width survives every move on its own - the column view isn't recreated.
- Unlike the filter one this does work, but it has a side effect the comment doesn't mention: a resized width now also survives hiding and re-showing a column (on `next` it resets). That's a behaviour change outside column actions, and it overwrites `width` on the consumer's column definitions.
- Keeping widths is already available as an opt-in: the persist-settings directive writes `column.width` on resize.

If we do want widths kept across hide/show, it's better as its own change with a comment that says so. The "keeps a resized width when a column is moved/reordered by drag and drop" specs can be dropped with this.

---

projects/angular/data/datagrid/datagrid-column-actions.ts
```diff
@@ -0,0 +1,465 @@
...
+  ngAfterViewInit() {
+    // The grid owns arrow key handling for the header, so it has to stand down while either overlay
+    // has focus - the menu, or the filter this menu opens. ClrDatagridFilter normally does the second
+    // half itself, but only when it is opened through its own input, which is no longer the path.
+    if (this.keyNavigation) {
+      this.subs.push(
+        this.popoverService.openChange.subscribe(() => this.updateSkipItemFocus()),
+        this.columnPopover.openChange.subscribe(() => this.updateSkipItemFocus())
+      );
+    }
```
dtsanevmw, line 315: Neither subscription needs to be here:

- **The menu's own popover** doesn't need `skipItemFocus` at all - the menu renders in an overlay outside the grid, so its keys never reach the grid's key navigation. Without it all core tests pass, and opening by mouse/keyboard, ArrowDown and Escape land focus on exactly the same elements.
- **The filter's popover** does need it, but the filter should own it. It already subscribes to its popover's `openChange`, so setting `skipItemFocus` there (instead of in the `open` setter) covers every way it opens: its header toggle, `[clrDgFilterOpen]`, and this menu. The header toggle bypasses the setter too, so it never set the flag before either.

```ts
// datagrid-filter.ts
popoverService.openChange.subscribe(change => {
  this.ariaExpanded = change;
  this.openChange.emit(change);

  if (keyNavigation) {
    keyNavigation.skipItemFocus = change;
  }
})
```

The menu then drops both subscriptions, `updateSkipItemFocus()` and its `KeyNavigationGridController` dependency (-20 lines), and nothing is left to reset on destroy. Change: https://github.com/vmware-clarity/ng-clarity/commit/fbf9a143e89dee49db6f63431f9d579cb2830985

---

projects/angular/data/datagrid/datagrid-column-actions.ts
```diff
@@ -0,0 +1,465 @@
...
+  private set menu(menu: ClrDropdownMenu) {
+    this.menuInstance = menu;
+    this.menuItemsSubscription?.unsubscribe();
+
+    if (menu) {
+      this.menuItemsSubscription = menu.items.changes.subscribe(() => this.linkMenuItems());
+    }
+
+    this.linkMenuItems();
+  }
```
dtsanevmw, line 304: This view query plus `registerAction`/`unregisterAction`, `viewReady` and `linkMenuItems` all work around one thing: `ClrDropdownMenu` hands the focus handler only its own content query. A small focus handler for this menu can append the projected items to whatever it is handed instead:

```ts
@Injectable()
class ColumnActionsFocusHandler extends DropdownFocusHandler {
  private ownItems: FocusableItem[] = [];
  private projectedItems: FocusableItem[] = [];

  override addChildren(children: FocusableItem[]) {
    this.ownItems = children;
    super.addChildren([...children, ...this.projectedItems]);
  }

  setProjectedItems(items: FocusableItem[]) {
    this.projectedItems = items;
    this.addChildren(this.ownItems);
  }
}
```

with the projected items found by `@ContentChildren(FocusableItem, { descendants: true })` on this component. No item has to register itself any more, and it fixes a gap: a plain `clrDropdownItem` projected into the menu is skipped by the arrow keys today (`Sort Ascending › Sort Descending › Custom › wraps`); with this it is included (`… › Custom › Plain`).

Nothing covered the arrow key order of projected items - even leaving them out entirely fails no test - so the change adds a spec for it. Core 725 and addons 276 pass. Change: https://github.com/vmware-clarity/ng-clarity/commit/7eb5da8ae003a28980c12d90938e43a6b46837e1

---

projects/addons/datagrid/addons/column-ordering/datagrid-columns-order.directive.ts
```diff
@@ -25,6 +29,26 @@ import { DatagridColumnsOrderService } from './datagrid-columns-order.service';
...
+ * The host is responsible for rendering a new order by destroying and recreating the column views,
+ * rather than letting Angular relocate the existing ones. A pinned column is rendered in the
+ * datagrid's sticky container and the rest in the scrollable one, so a single declared list of
+ * columns ends up split across two DOM parents. Reordering the list then makes Angular's `@for`
+ * reconciliation relocate a column against a reference node that lives in the other container, and
+ * the DOM insert throws - which leaves the header short of columns, because change detection gives
+ * up half way through. Rebuilding gives the reconciliation nothing to relocate, which is what makes
+ * every move within a column's own group renderable, including reordering the pinned columns with
+ * each other. See `rebuildColumnViews` in DatagridComponent.
+ */
```
dtsanevmw: The directions are hardcoded strings everywhere (`'left'`, `'right'`, plus a `switch` over them), even though there is a type for them. An enum lets the code, the template and the specs say `ColumnMoveDirection.Left`. With pinned moves and Start/End dropped, only two values are left. The class comment also has to change, since the host no longer rebuilds the views:

```suggestion
/**
 * The direction of a one step column move, the same as the arrow keys while a column is grabbed.
 */
export enum ColumnMoveDirection {
  Left = 'left',
  Right = 'right',
}

/**
 * Reorders the columns of a datagrid through drag and drop, the arrow keys, or `moveColumnTo`.
 *
 * Pinned columns are not moved. The datagrid renders them in its sticky container and the rest in
 * the scrollable one, so a single declared list of columns is split across two DOM parents, and
 * reordering pinned columns makes Angular's `@for` relocate a column against a reference node in the
 * other container - the DOM insert throws. Moves between the scrollable columns are not affected.
 */
```

In the template the move buttons become `[appfxColumnMoveAction]="ColumnMoveDirection.Left"`, with `protected readonly ColumnMoveDirection = ColumnMoveDirection;` on `DatagridComponent`. The keyboard path becomes `moveLeft ? ColumnMoveDirection.Left : ColumnMoveDirection.Right`.

---

projects/addons/datagrid/addons/column-ordering/datagrid-columns-order.directive.ts
```diff
@@ -62,21 +88,80 @@ export class DatagridColumnsOrderDirective implements OnInit, OnDestroy, OnChang
...
+        // the one that was just used. It does that from a timeout of its own - see
+        // DropdownFocusHandler.moveToFirstItemWhenOpen - so putting focus back has to be queued
+        // behind it rather than done here. Both are zero delay timeouts and this one is queued while
+        // the click above is still on the stack, so it always runs second, and the intermediate focus
+        // is never painted.
+        setTimeout(() => this.focusMoveAction(direction));
+      },
+      { injector: this.injector }
     );
   }
```
dtsanevmw: `reopenColumnActions`, `focusMoveAction` and `findColumnActionsTrigger` (below) only exist because `rebuildColumnViews` destroys the menu that was open. Without the rebuild the column and its menu survive the move, and `clrCanClosePopover="false"` re-anchors the menu, the same way it already does when no column is pinned. So all three can go, together with:
- `afterNextRender`;
- the `Injector` and `DOCUMENT` constructor parameters (the constructor goes back to what it is on `next`);
- the `ElementRef` in `ColumnMoveActionDirective`.

Without the rebuild, the specs that cover this (reopen on the moved column, focus left on the used action, repeat by keyboard) no longer apply. The addon suite passes without them.

---

projects/addons/datagrid/addons/column-ordering/datagrid-columns-order.directive.ts
```diff
@@ -140,14 +240,41 @@ export class DatagridColumnsOrderDirective implements OnInit, OnDestroy, OnChang
...
+        targetGroupIndex = groupIndex - 1;
+        break;
+      case 'right':
+        targetGroupIndex = groupIndex + 1;
+        break;
+    }
+    targetGroupIndex = Math.min(Math.max(targetGroupIndex, 0), lastGroupIndex);
+
+    return this.createColumnIndices(previousColumn, group[targetGroupIndex]);
   }
```
dtsanevmw: Without pinned moves and Start/End, the neighbour is just the next column among the visible, unpinned ones. It is either a different column or nothing, so this can return `null`:

```suggestion
  /**
   * Resolves a one step move of the column at `visibleColumnIndex` (an index into the visible columns)
   * into indices into `dgColumnsOrderColumns`, or `null` when there is no move to make.
   *
   * The neighbour is taken among the scrollable columns, because those are the ones the user sees side
   * by side - a pinned column between them in the array is rendered in the other container. A pinned
   * column itself is not moved, see the class comment.
   */
  private computeTargetIndices(visibleColumnIndex: number, direction: ColumnMoveDirection) {
    const column = this.dgColumnsOrderColumns.filter(other => !other.hidden)[visibleColumnIndex];
    const scrollableColumns = this.dgColumnsOrderColumns.filter(other => !other.hidden && !other.pinned);
    const index = scrollableColumns.indexOf(column);
    const target = scrollableColumns[index + (direction === ColumnMoveDirection.Left ? -1 : 1)];

    if (index < 0 || !target) {
      return null;
    }

    return this.createColumnIndices(column, target);
  }
```

`isMoveApplicable` then isn't needed: its "different column" check can't fail any more. `canMoveColumn` becomes `return !!this.computeTargetIndices(visibleColumnIndex, direction);` and `moveColumnTo` checks `if (!indices) { return false; }`. `isReorderAllowed` stays, because a drop can still land across a pinned column.

I checked that the pinned specs actually test this: when pinned columns are let back into the neighbour list, 10 of them fail, including the `insertBefore` throw.

---

projects/addons/datagrid/addons/column-ordering/column-move-action.directive.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    // Normally the menu survives the move and `clrCanClosePopover="false"` re-anchors it to the
+    // trigger, so it is still open with focus on this item, which is where it belongs. Once a column
+    // is pinned though, applying the move rebuilds the column views, and this very button is
+    // destroyed with the column it belonged to - there is no menu left to re-anchor, so the one on
+    // the column in its new place is opened instead to end up in the same state.
+    if (!this.elementRef.nativeElement.isConnected) {
+      this.columnsOrderDirective.reopenColumnActions(column, this.direction);
+    }
+  }
```
dtsanevmw: With the reopen gone (see the comment on `reopenColumnActions`), the click only has to ask for the move:

```suggestion
  constructor(private readonly columnsOrderDirective: DatagridColumnsOrderDirective) {}

  get disabled(): boolean {
    return !this.columnsOrderDirective.canMoveColumn(this.columnIndex, this.direction);
  }

  @HostListener('click')
  protected onClick() {
    // A disabled menu item still receives the click, so the move is only asked for when it applies.
    if (!this.disabled) {
      this.columnsOrderDirective.moveColumnTo(this.columnIndex, this.direction);
    }
  }
}
```

The class comment above also mentions start/end and "the pinned-column guard". It could read: "Moves this column one step left or right…" and "…because the column order array (`dgColumnsOrderColumns`) that resolves the move lives on the directive, not the service."

---

projects/addons/datagrid/datagrid.component.html
```diff
@@ -99,6 +95,102 @@
...
+              [clrCanClosePopover]="false"
+              appfxColumnMoveAction="end"
+              #moveToEnd="appfxColumnMoveAction"
+              [columnIndex]="index"
+              [clrDisabled]="moveToEnd.disabled"
+            >
+              <cds-icon shape="angle-double" direction="right" aria-hidden="true"></cds-icon>
+              {{ dgStrings.moveColumnToEnd }}
+            </button>
+          -->
```
dtsanevmw: The Move to Start / Move to End buttons are commented out, but everything behind them still ships:
- the `'start'` and `'end'` directions;
- the public `DatagridStrings.moveColumnToStart` / `moveColumnToEnd`, which are in the API report even though nothing renders them.

I'd suggest removing the commented buttons, the two strings and the start/end directions now, and adding them back when the actions are needed, rather than shipping public strings for a feature that isn't there. The two remaining buttons would use the enum: `[appfxColumnMoveAction]="ColumnMoveDirection.Left"` / `.Right`.

---

dtsanevmw (commented):
Follow-up on the column ordering files now that `rebuildColumnViews` and pinned moves are dropped. The reopen logic, Start/End and `isMoveApplicable` can go too, and the directions become an enum. The whole change, with the specs updated (addon suite 267/267): https://github.com/vmware-clarity/ng-clarity/commit/c7ebf8bbcb89b77d9d9861cfc82d8c31cd5a1e1e

---

projects/addons/datagrid/addons/column-ordering/datagrid-columns-order.directive.ts
```diff
@@ -170,6 +297,41 @@ export class DatagridColumnsOrderDirective implements OnInit, OnDestroy, OnChang
...
+    const renderedIndex = renderedColumns.findIndex(other => isEqualColumns(column, other));
+
+    if (renderedIndex < 0) {
+      return null;
+    }
+
+    return this.elementRef.nativeElement
+      .querySelectorAll<HTMLElement>('.datagrid-column-actions-toggle')
+      .item(renderedIndex);
+  }
```
dtsanevmw: The structure of this one is the problem. To find a column's menu trigger it:
- rebuilds the datagrid's rendered order by hand: pinned first, then the rest. That copies the datagrid's layout rule into the ordering directive, so it silently goes wrong if that rule ever changes;
- queries every `.datagrid-column-actions-toggle` in the grid and picks one by position. That only works while each rendered column has exactly one toggle and the DOM order matches the order rebuilt above;
- goes from a `ColumnDefinition` to a DOM node, even though the column's own `appfxColumnOrder` directive already has both its `columnData` and its `clr-dg-column-actions`.

It exists only for `reopenColumnActions`, which exists only for `rebuildColumnViews`. With the rebuild dropped it can be deleted together with `focusMoveAction`, `afterNextRender` and the `Injector`/`DOCUMENT` constructor parameters, rather than restructured: https://github.com/vmware-clarity/ng-clarity/commit/c7ebf8bbcb89b77d9d9861cfc82d8c31cd5a1e1e

If a reopen is ever needed again (for example if moving pinned columns comes back), the column shouldn't be looked up from the outside. The move can be reported on `DatagridColumnsOrderService`, and the `appfxColumnOrder` directive whose `columnData` matches opens its own `@ContentChild(ClrDatagridColumnActions)`. That's what https://github.com/vmware-clarity/ng-clarity/commit/dedbf9cd176f48b2e05fe2d9a396d448e7e7c23b prototyped.

---

projects/addons/datagrid/datagrid.component.html
```diff
@@ -99,6 +95,102 @@
...
+                  (click)="onColumnActionClick(action, column)"
+                >
+                  @if (action.icon) {
+                    <cds-icon [shape]="action.icon" aria-hidden="true"></cds-icon>
+                  }
+                  {{ action.label }}
+                </button>
+              }
+            }
+          }
```
dtsanevmw: Four things in this block, each reproduced with a spec at this commit:

1. **The comment is wrong.** These actions don't close the menu on click. After clicking one, the trigger still has `aria-expanded="true"` and the menu is still in the DOM, the same as after "Sort Ascending". `closeMenu()` returns early because `isMenuClosable` is always `false`. Staying open matches the default we want, so only the comment needs to change.
2. **`[title]="action.tooltip"` renders `title="undefined"`** for every action without a tooltip, so hovering shows "undefined" and it can be read out as the item's description. `[attr.title]` leaves the attribute off when the value is `undefined`.
3. **`track action.id` warns on duplicate ids.** The grid-wide and per-column lists are merged, so the same `id` in both (for example a grid "copy" and a column "copy") triggers NG0955 (duplicate keys). `track action` doesn't depend on ids being unique across the two lists.
4. **The outer `@if` does nothing:** `getColumnActions` always returns an array, so `as customColumnActions` is always truthy. `@let` plus the one `@if` that matters is simpler.

```suggestion
          @let customColumnActions = getColumnActions(column);
          @if (customColumnActions.length) {
            <div class="dropdown-divider" role="separator"></div>
            <!--
              Application actions come last, so the built-in ones keep the same place in every menu
              whatever a column adds. They leave clrCanClosePopover at its default: they do not move
              the column, so the menu has nothing to re-anchor, and it stays open like the built-in items.
            -->
            @for (action of customColumnActions; track action) {
              <button
                type="button"
                clrDgColumnAction
                [clrDisabled]="!action.enabled"
                [attr.aria-label]="action.ariaLabel"
                [attr.title]="action.tooltip"
                (click)="onColumnActionClick(action, column)"
              >
                @if (action.icon) {
                  <cds-icon [shape]="action.icon" aria-hidden="true"></cds-icon>
                }
                {{ action.label }}
              </button>
            }
          }
```

With this applied, `title` is absent, NG0955 is gone, and the addon suite passes.

---

projects/addons/datagrid/datagrid.component.html
```diff
@@ -99,6 +95,102 @@
...
+                menu on click - unlike the move actions, they are not expected to be repeated.
+              -->
+              @for (action of customColumnActions; track action.id) {
+                <button
+                  type="button"
+                  clrDgColumnAction
+                  [clrDisabled]="!action.enabled"
+                  [attr.aria-label]="action.ariaLabel"
+                  [title]="action.tooltip"
+                  (click)="onColumnActionClick(action, column)"
```
dtsanevmw: `ActionDefinition.children` is silently ignored here. An action with `children` renders as a plain item, clicking it reports the parent through `actionClick`, and the children can't be reached. Reproduced with `{ id: 'more', label: 'More', enabled: true, children: [{ id: 'child', label: 'Child action', enabled: true }] }`: the menu shows "More", and a click emits `more`.

A nested menu isn't needed now, but the interface should say so, the same way it already marks `class` and `isVisible` as ActionBar-only and `icon` as column-actions-only:

```ts
/**
 * When an action has children, it is displayed like a dropdown and
 * the children are the dropdown items.
 * @note Not applicable for column actions, which are rendered as single items.
 */
children?: ActionDefinition[];
```

---

dtsanevmw (commented):
Review of the column actions block in `datagrid.component.html`: the comment about closing is wrong, `title="undefined"`, duplicate `track` keys, a no-op outer `@if`, and `children` being silently ignored. Details and a suggestion are inline.

---

projects/addons/datagrid/datagrid.component.ts
dtsanevmw: @valentin-mladenov you were right that the rebuild is needed for hide/show. At this commit, though, it is only called from `onColumnOrderChange`, so the case you described still throws here. Reproduced at this commit, in the advanced `appfx-datagrid` with rows: pin C1 and C2, hide both, show C2, then show C1 → `NotFoundError: Failed to execute 'insertBefore'`, and C1 never renders.

**What causes it:** it's the row cells, not the headers (with no rows it doesn't throw). Clarity moves each pinned cell into the row's `.datagrid-pinned-cells`, out of the parent the cells are projected into (`datagrid-row.html`). When a cell `@for` has to create a cell in front of a pinned one, Angular inserts it next to a node that isn't there any more. Loose neighbours aren't moved, which is why unpinned columns are fine. The `appfx-datagrid` hides a column by removing it from `visibleColumns`, so showing it creates new cells. That's why only the advanced grid hits this.

**Other ways to trigger it** (all reproduced here):
- with C1 and C2 pinned: hide C1, show C1;
- a loose column shown again in front of a pinned one;
- the same columns passed in as new objects while a column is pinned (the cell loops `track column`).

**Suggestion:** keep the rebuild, but call it from the `columns` setter, and only when a pinned column is rendered and the new list has a column that isn't rendered yet. That's the only case where `@for` creates a view. Moves and hides reuse or remove views, so they don't rebuild and the actions menu keeps re-anchoring after a move:

```suggestion
  /**
   * Whether showing `visibleColumns` would make the `@for` loops create a column or cell in front of a
   * rendered pinned one. Clarity moves pinned cells into the row's pinned container, out of the parent
   * the cells are projected into, so inserting in front of one throws `NotFoundError` from
   * `insertBefore` and the new column never renders. This happens when a column is shown again, or when
   * the column definitions are replaced by new objects, while a column is pinned. Moves and hides only
   * reuse or remove views, so they never need it.
   */
  private needsColumnViewsRebuild(visibleColumns: ColumnDefinition<T>[]): boolean {
    const rendered = this.visibleColumns || [];
    return (
      !!this.clrDatagrid?.columns &&
      rendered.some((column: ColumnDefinition<T>) => column.pinned) &&
      visibleColumns.some((column: ColumnDefinition<T>) => !rendered.includes(column))
    );
  }

  /**
   * Renders `visibleColumns` by destroying every column and cell view first, so the `@for` loops have
   * nothing to insert in front of. Column state that has to survive this is kept on the column
   * definitions - `defaultSortOrder`, `defaultFilterValue` and `width` - so the new views bind it back.
   */
  private rebuildColumnViews(visibleColumns: ColumnDefinition<T>[]): void {
    this.visibleColumns = [];
    this.cdr.detectChanges();
    this.visibleColumns = visibleColumns;
    this.cdr.detectChanges();
    this.resize();
  }
```

with the setter (not in the diff, so no suggestion there):

```ts
set columns(columns: ColumnDefinition<T>[]) {
  this.#dgColumns = columns;
  if (columns) {
    const visibleColumns = columns.filter((column: ColumnDefinition<T>) => !column.hidden);
    if (this.needsColumnViewsRebuild(visibleColumns)) {
      this.rebuildColumnViews(visibleColumns);
    } else {
      this.visibleColumns = visibleColumns;
    }
    this.columnDefsChange.emit(this.#dgColumns);
  }
  // ... dataChanged() as before
}
```

`onColumnOrderChange` then only needs its non-rebuild branch, since pinned columns aren't moved any more.

**The two write-backs need to stay.** Showing a column now rebuilds the others, so their state has to come back from the definitions. Without the `defaultFilterValue` write-back an active filter on another column is lost (1 row → 4). Without the `width` write-back a resized column loses its width (293px → 94px). The setter's `dataChanged()` re-applies the restored filter, so here the filter really does survive. That corrects what I wrote on those two threads, which only holds for moves.

The full change with specs: https://github.com/vmware-clarity/ng-clarity/commit/aa308aaadde3472e12353eed2fb9d693c049de64. Addon suite 274/274. Each part is covered: removing the rebuild fails the 4 show specs, rebuilding on every change fails "does not rebuild for a move or a hide", and removing either write-back fails its spec.

---

projects/addons/datagrid/datagrid.component.ts
dtsanevmw: Correction: this only holds for moves. Showing a column while another one is pinned throws, and the fix needs `rebuildColumnViews` from the `columns` setter (see the new comment on `rebuildColumnViews`). With that in place this write-back has to stay: without it, showing a column clears an active filter on another column (1 row → 4). On that path the setter's `dataChanged()` re-applies the restored value, so the filter really does survive.

---

projects/addons/datagrid/datagrid.component.ts
dtsanevmw: Correction: same as the filter one, this only holds for moves. With the rebuild called from the `columns` setter to fix showing a column while another is pinned (see the new comment on `rebuildColumnViews`), this write-back has to stay. Without it a resized column loses its width when another column is shown (293px → 94px).

---

PR #2677, opened 2026-09-01
chore(website): add visual regression tests for the website

dtsanevmw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] CI related changes

Playwright visual regression tests cover only Storybook stories. The website application (`projects/website`) — the component documentation pages, content pages, home, and the theme builder — has no visual test coverage, so regressions on clarity.design can only be caught by manual review.

Issue Number: N/A

## What is the new behavior?

A parallel visual regression pipeline for the website that shares its infrastructure with the Storybook one:

- **`tests/helpers/create-vrt-config.ts`** — shared Playwright config factory; `playwright.config.ts` (Storybook) and `playwright-website.config.ts` are thin wrappers stating only what differs (testDir, timeouts, web server command/port).
- **`scripts/start-static-server.ts`** — one parameterized static server (root, port, optional `--spa` fallback to `index.html`, matching the redirect rule in `netlify-website.toml`), replacing the previous `start-storybook-server.ts`.
- **`tests/helpers/vrt.ts`** and **`tests/helpers/screenshot-options.interface.ts`** — the matrix env vars, screenshot assertion options, and the `ScreenshotOptions` override type shared by both suites (the Storybook suite gains `maskSelectors` support as a result).
- **`tests/website/visual-snapshots.spec.ts`** — generates one test per page from `componentlist.json` plus the markdown content pages, home, and the theme builder (82 pages), per browser/theme/density. Each test captures the default overview tab, every additional documentation tab (code, api, accessibility, usage, ...) discovered at runtime from the rendered tab bar, and every section subpage a tab links to one level below it (for example the 26 datagrid `code/...` sections and the 6 vertical-nav ones; deeper links are an embedded demo's own navigation states and are skipped). Screenshots are stored as `{browser}/website/{page}/{view}-{theme}-{density}.png` — one directory per page, mirroring the Storybook suite's one directory per component. Theme and density are seeded through the website's localStorage keys. For determinism it blocks all off-origin requests, lets Clarity's root scroll container grow (scoped so demos' embedded `.main-container` shells keep their natural size), resizes the viewport to the settled content height instead of using `captureBeyondViewport` (re-tested on Playwright 1.62: plain `fullPage` still reflows virtual-scroll datagrids mid-capture and fails 8 of 10 runs), and masks or hides regions that can never render deterministically (animated GIFs, indeterminate progress bars, `app-animated-example` demos, the modal animation demo, the translate demo's live clock, the virtual-scroll section's randomly generated rows).
- **`tests/website/screenshot-options.ts`** — per-view overrides (exclude, viewport, full-page opt-out, masking), mirroring `tests/screenshot-options.ts`.
- **`pr-build.yml`** — new `pr-website-visual-regression-test` job (chromium × light/dark × default/compact) with a cached website build. Its diff artifacts follow the `visual-snapshot-changes-*.diff` naming, so the existing snapshot update bot picks them up with no bot changes.
- 944 baseline snapshots (chromium, both themes, both densities, all tabs and sections).

Known limitations, documented in the spec and options files:

- The table of contents (`app-table-of-contents`) is hidden in all captures: it renders nondeterministically (#2678 — `TableOfContentsComponent` races the lazy-loaded content and never re-queries; CI and local runs of this PR captured the datagrid page in opposite states). Once #2678 is fixed, the hide rule can be removed to restore TOC coverage.
- `/documentation/accessibility` is excluded — it redirects to `/pages/accessibility`, which is already covered.
- `/documentation/theme-builder` is excluded — it throws a router NG04005 error on load and intermittently renders as a blank shell, which the CI runners hit consistently (#2680). The theme-builder tool page itself stays covered.

Every browser/theme/density combination was verified locally with retries disabled across repeated runs plus burn-ins of the historically flaky pages, with zero failures; the Storybook suite passes unchanged against its existing snapshots after the infrastructure refactor.

Local usage: `npm run _build:website`, then `CLARITY_VRT_BROWSER=chromium CLARITY_VRT_THEME=light CLARITY_VRT_DENSITY=default npm run test:vrt:website` (or `test:vrt:website:update` to update snapshots).

- [ ] Yes
- [x] No


tests/website/visual-snapshots.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+ * flaky captures; resizing up front lets such content settle first. The loop exits once a
+ * measurement taken 250ms after the last resize still matches the viewport height.
+ */
+async function fitViewportToContent(page: Page, viewportWidth: number) {
+  const measureContentHeight = () =>
+    page.evaluate(() => Math.min(Math.max(document.body.scrollHeight, document.documentElement.scrollHeight), 20000));
+
+  let viewportHeight = await measureContentHeight();
+
+  for (let attempt = 0; attempt < 8; attempt++) {
```
valentin-mladenov: 2 magic numbers here 

- 8 attempts, no more. Why?
- 20000 is the maximum minimum. Everything above that height will be cut. Why do we need such cut position?

Looks like the 8 attempts for loop should be a while. 8 attempts by 250ms is 2 seconds, but still the attempts are not derivered from the total wait time.

valentin-mladenov (commented):
Check my comments.

tests/website/visual-snapshots.spec.ts
dtsanevmw: Reworked in 45e37cd: the loop is now bounded by a named settle timeout (`resizeSettleTimeoutMs`, 2s re-measured every 250ms) instead of a fixed attempt count.

The 20000px cap is now a named constant (`maxViewportHeightPx`) with its reason documented: some demos size themselves off the viewport, so each resize can grow the content again — the bound keeps such a resize feedback loop finite instead of letting the page grow without limit. No page legitimately reaches it (the tallest capture is ~18.7k after pinning the login demo, which was the one offender).

valentin-mladenov (dismissed):

valentin-mladenov (approved):

PR #2681, opened 2026-09-02
feat(tree-view): add expand all and collapse all to the tree and tree nodes

dtsanevmw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

Expansion state lives on each `clr-tree-node` only. There is no way to expand or collapse a whole tree, or a whole subtree, at once: applications have to bind `[clrExpanded]="true"` on every single node, and neither `clr-tree` nor `clr-tree-node` can reach nested nodes to do it for them.

## What is the new behavior?

`clr-tree` gets `expandAll()` and `collapseAll()`, applying to every node of the tree.

`clr-tree-node` gets `expandDescendants()` and `collapseDescendants()`, applying to the node and all of its descendants.

Both are reached through a template reference variable:

```html
<button type="button" (click)="tree.expandAll()">Expand all</button>
<clr-tree #tree>...</clr-tree>
```

This is the shape the Angular CDK tree exposes (`expandAll`, `collapseAll`, `expandDescendants`, `collapseDescendants`). There is deliberately no tree-level input or output: per-node expansion stays on `[(clrExpanded)]`, the equivalent of the CDK's `isExpanded`, which is the only two-way binding the CDK tree has either. An earlier revision of this PR added `[(clrAllExpanded)]` and `[(clrDescendantsExpanded)]` two-way bindings; they are gone, along with the tri-state bookkeeping they required.

How it works:

- Bulk operations walk the model tree, which already knows parents and children for declarative and recursive trees alike. `TreeNodeModel` gets an internal `_setExpandedRecursive()` and a `_node` back-reference to the rendered node, so the walk can apply the change. A bulk operation is O(subtree), with no per-node subscriptions and no Subjects.
- Nodes created while their subtree is expected to be expanded (lazy-loaded children, dynamic nodes) check a boolean cascade flag in `ngAfterContentInit`, when their children are known, so leaves are never expanded and lazy-loaded trees cascade level by level. That path goes through the same internal entry point as the walk, so those nodes skip the animation too. The flag is cleared as soon as any node in scope is collapsed, so a node re-expanded afterwards does not cascade; expanding in bulk again turns it back on.
- Leaves are skipped when expanding, so they never emit a meaningless `clrExpandedChange`. Disabled nodes and their subtrees are left untouched, like for selection propagation, and a bulk call on a disabled node is a no-op.
- Bulk changes skip the expand animation through dedicated `expandedInstant` / `collapsedInstant` animation states, so a large tree does not run hundreds of height animations at once. The next regular toggle animates as usual.
- A collapsed subtree is `inert`, and the tree host drops its own `tabindex` the first time it is focused. So a collapse that hides the node holding the roving tab stop hands that tab stop to a node that is still visible, without moving focus, instead of leaving the tree unreachable by keyboard.
- Everything the walk needs across objects is `_`-prefixed per the coding guidelines, so the added public surface is exactly the four methods.

Also included:

- `RecursiveTreeNodeModel` now tolerates a children fetch resolving to `null` or `undefined`, which the `AsyncArray` type already allows but previously crashed on.
- Specs for the tree and node operations, disabled subtrees, nodes and leaves added later, repeated expand/collapse cycles, destroying the tree right after a bulk operation, the animation states, the tab stop after a collapse, the lazy-loaded cascade and its stop condition, and the model-level walk.
- Storybook stories for declarative, eager recursive, subtree and lazy-loaded trees.
- Website documentation: a new "Expanding or collapsing all nodes" section with a live demo.
- A configurable large-tree page in the demo app for measuring bulk expansion.
- Regenerated API reports.

- [ ] Yes
- [x] No

Nothing that existed in a released version changes. The bindings removed during review were only ever added earlier on this branch.

## Other information

- On a lazy-loaded tree, expanding all fetches every level of the tree. This is documented on the website page.
- Disabled branches are excluded from bulk operations rather than remembered, so enabling one later does not expand it retroactively. Call `expandDescendants()` on it if you need it opened.


projects/angular/data/tree-view/tree-node.ts
```diff
@@ -250,6 +304,39 @@ export class ClrTreeNode<T> implements OnInit, AfterContentInit, AfterViewInit,
...
+  expandDescendants() {
+    this.setDescendantsExpanded(true);
+  }
+
+  /**
+   * Collapses this node and every node below it, without animation. Disabled nodes are left untouched.
+   */
+  collapseDescendants() {
+    this.setDescendantsExpanded(false);
+  }
```
valentin-mladenov, line 306: `expandDescendants` and `collapseDescendants` are not really needed. `clrExpandDescendants` setter id fully capable of handle that logic. Check my other comment for it.

```suggestion
```

projects/angular/data/tree-view/tree-node.ts
```diff
@@ -165,6 +181,20 @@ export class ClrTreeNode<T> implements OnInit, AfterContentInit, AfterViewInit,
...
+  @Input('clrExpandDescendants')
+  get descendantsExpanded(): boolean {
+    return this._model.descendantsExpanded;
+  }
+  set descendantsExpanded(value: boolean) {
+    value = !!value;
+    if (value !== this._model.descendantsExpanded) {
+      this.setDescendantsExpanded(value);
+    }
+  }
```
valentin-mladenov: Linked to comment: https://github.com/vmware-clarity/ng-clarity/pull/2681/changes#r4023608533 
Naming of the Input's setter and getter doesn't sound correct, should follow the Input name.
Although `expandDescendants` sound like an action it's important to be consistent.

The other option is to change the Input's and Output's names to `clrDescendantsExpanded` and `clrDescendantsExpandedChange`.

```suggestion
  @Input('clrExpandDescendants')
  get expandDescendants(): boolean {
    return this._model.descendantsExpanded;
  }
  set expandDescendants(value: boolean) {
    value = !!value;
    if (value !== this._model.descendantsExpanded) {
      this.setDescendantsExpanded(value);
    }
  }
```

projects/angular/data/tree-view/models/tree-node.model.ts
```diff
@@ -9,9 +9,33 @@ import { BehaviorSubject } from 'rxjs';
...
 
+/**
+ * Implemented by the component rendering a node, so that bulk expand operations driven by the model tree
+ * can be applied to the rendered node.
+ */
+export interface TreeNodeExpander {
+  /**
+   * Expands or collapses the rendered node as part of a bulk operation.
+   */
+  setExpandedInBulk(expanded: boolean): void;
```
valentin-mladenov: Why the need for set word?

```suggestion
  expandInBulk(expanded: boolean) {
```

projects/angular/data/tree-view/models/tree-node.model.ts
```diff
@@ -9,9 +9,33 @@ import { BehaviorSubject } from 'rxjs';
...
   expanded: boolean;
+  /**
+   * True while this node and all of its descendants are expected to be expanded, see `expandDescendants()`.
+   * Descendants created afterwards (lazy-loaded children, dynamic nodes) come in expanded while this is set.
+   */
+  descendantsExpanded = false;
+  /*
+   * Internal, registered by the component rendering this model.
+   */
+  _expander: TreeNodeExpander | null = null;
```
valentin-mladenov: `TreeNodeExpander` by extension is a `ClrTreeNode` and that creates a circular depondency.
ClrTreeNode creates the model in it's constructor and assign itself as the _expander.
Is there a way for not to have the circular dependency, but the communication between the Node and the Model to happen through signals?

```suggestion
```

projects/website/src/app/documentation/demos/tree-view/expand-all-tree/example.component.html
```diff
@@ -0,0 +1,7 @@
+<!--
+  ~ Copyright (c) 2016-2026 Broadcom. All Rights Reserved.
+  ~ The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+<p>StackBlitz example placeholder</p>
```
valentin-mladenov: What is the point of this file?
```suggestion
```

projects/website/src/app/documentation/demos/tree-view/expand-all-tree/example.component.scss
```diff
@@ -0,0 +1,6 @@
+/*
+ * Copyright (c) 2016-2026 Broadcom. All Rights Reserved.
+ * The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+ * This software is released under MIT license.
+ * The full license information can be found in LICENSE in the root directory of this project.
+ */
```
valentin-mladenov, line 6: Empty file. Remove it.
```suggestion
```

projects/angular/data/tree-view/tree.ts
```diff
@@ -90,6 +115,31 @@ export class ClrTree<T> implements AfterContentInit, OnDestroy {
...
+  expandAll() {
+    this.setAllExpanded(true);
+  }
+
+  /**
+   * Collapses every node of the tree, without animation. Disabled nodes are left untouched.
+   */
+  collapseAll() {
+    this.setAllExpanded(false);
+  }
```
valentin-mladenov, line 108: Similar to my comment: https://github.com/vmware-clarity/ng-clarity/pull/2681/changes#r4023608533.

`expandAll` and `collapseAll` are not really needed. clrExpandAll setter id fully capable of handle that logic. 

```suggestion
```

projects/angular/data/tree-view/tree.ts
```diff
@@ -64,13 +73,29 @@ export class ClrTree<T> implements AfterContentInit, OnDestroy {
...
+  get allExpanded(): boolean {
+    return this.featuresService.allExpanded;
+  }
+  set allExpanded(value: boolean) {
+    value = !!value;
+    if (value !== this.featuresService.allExpanded) {
+      this.setAllExpanded(value);
+    }
+  }
+
```
valentin-mladenov: Related to: https://github.com/vmware-clarity/ng-clarity/pull/2681/changes#r4024811931
Naming of the Input's setter and getter doesn't sound correct, should follow the Input name.
Although expandAll sound like an action it's important to be consistent.

The other option is to change the Input's and Output's names to clrAllExpanded and clrAllExpandedChange.

```suggestion
  /**
   * Two-way binding to expand or collapse every node of the tree at once, see `expandAll()`.
   */
  @Input('clrExpandAll')
  get expandAll(): boolean {
    return this.featuresService.allExpanded;
  }
  set expandAll(value: boolean) {
    value = !!value;
    if (value !== this.featuresService.allExpanded) {
      this.setAllExpanded(value);
    }
  }

```

valentin-mladenov (commented):
Check my comments.

projects/angular/data/tree-view/tree-node.ts
dtsanevmw: Kept these. The setter only runs when the bound value actually changes, so it can't re-apply. Concrete case: a subtree with a few nodes expanded through `[clrExpanded]`, the binding sits at `false`, user hits "collapse subtree" — we write `false`, Angular sees no change, nothing happens. Same after any manual collapse, because the binding resets itself to `false` at that point.

Threw a quick spec at it to be sure:

```
BA expanded, binding false
after setting the input to false: still expanded
after collapseDescendants():     collapsed
```

CDK's tree control exposes the same four methods for the same reason, and Modal/Wizard here already pair `open()`/`close()` with `clrModalOpen`/`clrWizardOpen`. Took the naming point though, see the thread below.

---

projects/angular/data/tree-view/tree-node.ts
dtsanevmw: Agreed, went with your second option: `[(clrDescendantsExpanded)]` and `clrDescendantsExpandedChange`, accessors unchanged. Couldn't use the first one, `get expandDescendants()` collides with the `expandDescendants()` method (TS2300).

For what it's worth the alias-to-property rule isn't really enforced here, roughly 100 of 133 aliased inputs don't follow it (`clrCloseMenuOnItemClick` to `isMenuClosable`, `clrMulti` to `multiSelect`). But every other two-way binding in the library is named after the state rather than the action, so this one was genuinely off.

---

projects/angular/data/tree-view/models/tree-node.model.ts
dtsanevmw: Kept the `set` prefix. It takes the state as an argument, same as `setSelected(state, ...)` a few lines down on the model, and `expandInBulk(false)` to collapse reads backwards.

The interface itself moved to `TreeFeaturesService` though, see your other comment.

---

projects/angular/data/tree-view/models/tree-node.model.ts
dtsanevmw: Fixed, good catch. The model doesn't reference the component at all anymore. `TreeFeaturesService` holds a `WeakMap` of model to node and does the walk itself, same shape as `ModalStackService` and `Closable`. `TreeNodeExpander` moved there with it, so it's out of the public API too and the models are back to plain data.

Skipped signals: there's no signal usage anywhere in the library yet, so tree-view would be the first one, and it would still need an effect per node, which is the fan-out I was trying to get rid of in the first place. Happy to revisit as part of a wider migration.

---

projects/website/src/app/documentation/demos/tree-view/expand-all-tree/example.component.html
dtsanevmw: Removed, along with the scss next to it. Nothing reads them, the stackblitz component builds its files from the `componentTemplate` and `componentClass` inputs. Plenty of other demo folders have no such files at all.

---

projects/website/src/app/documentation/demos/tree-view/expand-all-tree/example.component.scss
dtsanevmw: Removed.

---

projects/angular/data/tree-view/tree.ts
dtsanevmw: Same answer as the node thread: `collapseAll()` after a partial collapse would be a no-op through the input, since the binding is already `false` by then, and a tree with pre-expanded nodes can never be collapsed through it at all. Keeping both, input renamed to `clrAllExpanded`.

---

projects/angular/data/tree-view/tree.ts
dtsanevmw: Same, it's `[(clrAllExpanded)]` and `clrAllExpandedChange` now, with `allExpanded`/`allExpandedChange` behind them.

---

projects/angular/data/tree-view/models/tree-node.model.ts
dtsanevmw: Correction to what I wrote above, we went back to a direct reference after talking it through.

The registry was holding the same model-to-node link the field held, one indirection further away, so it wasn't really breaking the cycle, just relocating it. It's a plain `componentRef: ClrTreeNode<T> | null` on the model now, through a type-only import so the models still pull in nothing from the components at runtime. The extra interface and the register/unregister pair are gone with it.

The walk needs the node rather than just the expand service, because it also checks `isExpandable()`, sets the flag that skips the animation, and emits the output.

Still no signals, for the reasons above.

---

projects/angular/data/tree-view/models/tree-node.model.ts
dtsanevmw: Small correction to the second half of that: there's no interface anymore. The model holds the node directly now, see the other thread. The `setExpandedInBulk` name stands, for the `setSelected` reason above.

---

projects/angular/data/tree-view/models/tree-node.model.ts
dtsanevmw: Dropped the interface. The model holds the node directly now, `node: ClrTreeNode<T> | null`, assigned in the node's constructor. It's a type-only import, so the models still pull in nothing from the components at runtime.

It does need the node rather than just the expand service, since the walk also checks `isExpandable()`, sets the flag that skips the animation on bulk changes, and emits the output.

I did try holding it in a `WeakMap` on `TreeFeaturesService` instead, but that stores the same model-to-node link one indirection further away, so it doesn't really remove the cycle, it just moves it somewhere less obvious.

No signals for now. There's no signal usage anywhere in the library yet, so tree-view would be the first one, and it would still need an effect per node, which is the same fan-out. Better as part of a wider migration than smuggled in here.

---

projects/angular/data/tree-view/tree-node.ts
valentin-mladenov: Well then the Input doesn't work as intended.
When setting the input `descendantsExpanded` to false should collapse all descendant nodes, is it not?

By your example.
```
BA expanded, binding false
after setting the input to false: still expanded
```

Expected (by me) behaviour.
`after setting the input to false: all descendants should be collapsed, not only if they are all expanded previously`

valentin-mladenov: Thsi line of reasoning is leading me to the conclusion that the model's `descendantsExpanded` should not have a initial value (should be null or undefined).

 Similarly to the chackbox indeterminate state.

1. true if all are expanded.
2. false if all are collapsed.
3. null or undefined, if some are expanded.

valentin-mladenov: `expandDescendants` and `collapseDescendants` are ok to stay. Do not remove them.

projects/angular/data/tree-view/tree.ts
```diff
@@ -64,13 +73,30 @@ export class ClrTree<T> implements AfterContentInit, OnDestroy {
...
+  @Input('clrAllExpanded')
+  get allExpanded(): boolean {
+    return this.featuresService.allExpanded;
+  }
+  set allExpanded(value: boolean) {
+    value = !!value;
+    if (value !== this.featuresService.allExpanded) {
+      this.setAllExpanded(value);
+    }
+  }
```
valentin-mladenov: Based on my other comment: https://github.com/vmware-clarity/ng-clarity/pull/2681/changes#r4025984732

Why do we have the `(value !== this.featuresService.allExpanded)` check?

The inital value is false.

When setting the input `allExpanded` to false should collapse all nodes, not only if they are all expanded previously.

This line of reasoning is leading me to the conclusion that the model's descendantsExpanded should not have a initial value (should be null or undefined).

Similarly to the chackbox indeterminate state.

1. true if all are expanded.
2. false if all are collapsed.
3. null or undefined, if some are expanded.

```suggestion
  set allExpanded(value: boolean) {
    this.setAllExpanded(!!value);
  }
```

projects/angular/data/tree-view/tree-node.ts
dtsanevmw: Good call, done. Both `clrAllExpanded` and `clrDescendantsExpanded` are tri-state now: `true` when everything in scope is expanded, `false` when none of it is, `null` when only part of it is.

That turned out to fix your earlier point too. A node toggled on its own moves every scope claiming the opposite to `null`, so asking for `false` from there is a real change and the subtree does collapse:

```
pre-expanded A     : a=true  state=null
after writing false: a=false state=false
```

And writing `null` back is ignored, so the mixed state coming out of a two-way binding is never read as an instruction, which is what made the unguarded version collapse everything when you touched one node.

Two details worth flagging:

Both outputs are async now, like `clrIfExpandedChange`. The mixed state is reached as a consequence of rendering, so emitting it synchronously changed a binding that had already been checked and threw NG0100 on any tree with a pre-expanded node.

They are also annotated explicitly rather than inferred. The library builds with `strict` off, where an inferred `boolean | null` is emitted as plain `boolean`, so consumers would never have learned that `null` can come out.

One case I deliberately left alone: a disabled branch is skipped by a bulk operation and does not make the scope mixed, so enabling it later does not expand it retroactively. That matches how selection leaves disabled nodes out of its own state. `expandDescendants()` on that branch opens it if needed.

---

PR #946, opened 2023-09-11
chore(demo): add stepper scenario with clrIfExpanded directive and fix icon in select

mivaylo:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If applicable, have a visual design approval

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [x] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #2691, opened 2026-09-08
fix(stepper): step title max width without description

valentin-mladenov:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Stepper's step header title is always have 264px width.

Issue Number: CDE-3000

## What is the new behavior?
Stepper's step header title will have 264px width when any step header have description, otherwise the title will span the whole stepper width.

- [ ] Yes
- [x] No


dtsanevmw:
I think this can be CSS only, without the new subscriptions / `ChangeDetectorRef` / API change.

The panel template already sets `clr-stepper-header-has-description` on the header button, the same way the accordion does. The only reason the stepper title never stretches is the hardcoded `max-width` on `.clr-step-title-wrapper` in `_stepper.clarity.scss`. Moving that cap under the existing class is enough:

```scss
.clr-stepper-header > .clr-stepper-header-button {
  > .clr-step-title-wrapper {
    // max-width removed
  }

  &.clr-stepper-header-has-description > .clr-step-title-wrapper {
    max-width: mixins.baselinePx(264);
  }
}
```

Note: just deleting the hardcoded line is not enough, because the stepper resets `max-width: initial` on `.clr-stepper-title` inside the wrapper, so the mixin's rule doesn't apply here.

This keeps the accordion and stepper consistent (per header, no description = full width)

Example on `dtsanevmw/quirky-curie-4mexja`: https://github.com/vmware-clarity/ng-clarity/compare/next...dtsanevmw/quirky-curie-4mexja

projects/angular/stepper/_stepper.clarity.scss
```diff
@@ -76,6 +76,10 @@
     display: block;
     margin-bottom: density.$clr-base-vertical-offset-xl;
 
+    &:has(.clr-stepper-description) .clr-stepper-header > .clr-stepper-header-button > .clr-step-title-wrapper {
```
idonchev-vmw, line 79: **Should-fix:** this `:has(.clr-stepper-description)` selector is unscoped, so it matches any `.clr-stepper-description` in the DOM, not just a header description of *this* stepper. `clr-step-description` gets the `.clr-stepper-description` class via a host binding, so it also appears on descriptions rendered by a nested stepper inside `clr-step-content` (a supported, demoed pattern — see `NestedStepper` story / `stepper.demo.html:192-258`). Concrete failure: an outer stepper whose own steps have no descriptions, but which contains an inner stepper that does, will have its outer titles wrongly capped at 264px — reintroducing the bug this PR fixes. Suggest scoping both sides of the selector via child combinators keyed off `.clr-stepper-header-has-description` instead, e.g. `&:has(> clr-stepper-panel > * > .clr-stepper-header.clr-stepper-header-has-description) > clr-stepper-panel > * > .clr-stepper-header > .clr-stepper-header-button > .clr-step-title-wrapper`.

projects/demo/src/app/stepper/stepper.demo.ts
```diff
@@ -7,7 +7,7 @@
 
 import { Component, ElementRef, ViewChild } from '@angular/core';
 import { FormControl, FormGroup, Validators } from '@angular/forms';
-import { ClrLoadingState } from '@clr/angular';
+import { ClrLoadingState, ClrStepper } from '@clr/angular';
```
idonchev-vmw, line 10: **Should-fix:** `ClrStepper` import only exists to type the unused `@ViewChild` below (line 19) — dead leftover from the reverted TS approach.

projects/demo/src/app/stepper/stepper.demo.ts
```diff
@@ -16,6 +16,7 @@ import { ClrLoadingState } from '@clr/angular';
 })
 export class StepperDemo {
   @ViewChild('nextBtn') nxtBtn?: ElementRef<HTMLElement>;
+  @ViewChild(ClrStepper) hasStepDescriptionStepper?: ClrStepper;
```
idonchev-vmw, line 19: **Should-fix:** `hasStepDescriptionStepper` is never read anywhere in `stepper.demo.ts` or `stepper.demo.html`. Leftover from the earlier TS-based `hasStepDescription` approach that was otherwise reverted — should be removed along with the `ClrStepper` import (line 10) and the stale comment at line 38 referencing the same abandoned approach.

projects/demo/src/app/stepper/stepper.demo.html
```diff
@@ -448,12 +471,7 @@ <h6>Programmatically change step after async call</h6>
 @if (!loading) {
 <form clrStepper [formGroup]="form" (ngSubmit)="submit()" [clrInitialStep]="initialStep">
   <clr-stepper-panel formGroupName="name">
-    <clr-step-title>Legal Name</clr-step-title>
-    <clr-step-description>
-      @if (!stepsExpandedState.name) {
-      <span>Closed</span>
-      } @else { Opened }
-    </clr-step-description>
+    <clr-step-title>Legal Name First Name Looong Loooooong Name</clr-step-title>
```
idonchev-vmw, line 474: **Should-fix:** this looks like unrelated manual-test scaffolding on the main reactive-forms demo. The title was renamed to a long test string and the panel's `clr-step-description` was deleted — that description demonstrated the `[(clrIfExpanded)]` two-way binding (added deliberately in #946). Since the PR already adds a purpose-built toggle demo higher up the page, and since sibling panels in this same form still have descriptions (so this title stays capped anyway and doesn't even illustrate the new uncapped behavior), suggest reverting this edit.

projects/angular/stepper/STYLES.md
```diff
@@ -0,0 +1,40 @@
...
+| CSS Custom Property                          | Description                                                      |
+| -------------------------------------------- | ---------------------------------------------------------------- |
+| --clr-stepper-error-color                    | Color of the error state icon and indicator.                     |
+| --clr-stepper-complete-color                 | Color of the complete state icon and indicator.                  |
+| --clr-stepper-header-left-indicator          | Left box-shadow indicator of a step header in its default state. |
+| --clr-stepper-header-left-current-indicator  | Left box-shadow indicator of the currently open step header.     |
+| --clr-stepper-header-left-complete-indicator | Left box-shadow indicator of a completed step header.            |
+| --clr-stepper-header-left-error-indicator    | Left box-shadow indicator of a step header with an error.        |
+
+## CSS Classes
```
idonchev-vmw, line 14: **Nit:** a couple of gaps in the new class table: (a) `clr-stepper-panel-inactive` is missing though it's a real emitted class (`stepper-panel.ts` `getPanelStateClasses` / `StepperPanelStatus.Inactive`, styled at `_stepper.clarity.scss:165`); (b) unlike accordion, stepper doesn't alias `--clr-collapsible-panel-*` into `--clr-stepper-*`, so those are what consumers actually need to override for colors/borders — a pointer to `collapsible-panel/STYLES.md` would help; (c) the new conditional-max-width behavior this PR introduces isn't documented here at all, which is the one thing a consumer hitting this change would come looking for.

.storybook/stories/stepper/stepper.stories.ts
```diff
@@ -107,6 +108,50 @@ export const StepperAlignmentTest: StoryObj = {
   },
 };
 
+const StepperSingleDescriptionTemplate: StoryFn = args => ({
+  template: `
+    <form clrStepper [formGroup]="form">
+      <clr-stepper-panel formGroupName="step1">
+        <clr-step-title>Step 1: Have a really long title that will wrap if any step have description.</clr-step-title>
```
idonchev-vmw, line 115: **Nit:** "…that will wrap if any step **have** description." → "has a description". Worth fixing now since this text is baked into 16 committed snapshots — fixing it later forces a regen.

.storybook/stories/stepper/stepper.stories.ts
```diff
@@ -39,6 +39,7 @@ export default {
     ngSubmit: action('ngSubmit'),
     alignmentTest: false,
     showPreviousButton: false,
+    showDescriptions: true,
```
idonchev-vmw, line 42: **Nit:** `showDescriptions: true` at meta-level `args` applies the control to every story in this file, including the ones that ignore it. Low priority — matches the existing pattern for `alignmentTest`/`showPreviousButton` — but could be scoped to just the two new stories instead.

dtsanevmw (approved):

PR #2692, opened 2026-09-10
feat(card): create angular component

valentin-mladenov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: CDE-3120

## What is the new behavior?
Adds a new `clr-card` Angular composite component wrapping the existing `.card` CSS, matching the header/body/footer structure already documented for the CSS-only card:

- `clr-card` / `clr-card-header` / `clr-card-body` (`clr-card-body-title`, `clr-card-body-text`) / `clr-card-footer` sub components.
- `clr-card-image`, `clr-card-divider`, `clr-card-media-block` (`clrCardMediaImage`, `clrCardMediaTitle`, `clrCardMediaText`) sub components.
- Optional per-card collapsible behavior `clrCardCollapsible), independent of any other card on the page — no shared/group state
- **SCSS reorganized**: moved the card styles `_card.clarity.scss`, `_variables.card.scss`, `_properties.card.scss` and the "Card" section of `STYLES.md` from `layout/` into `card/`, so all card-related styles/docs live in one place.

- [ ] Yes
- [x] No


projects/angular/card/card.ts
```diff
@@ -0,0 +1,71 @@
...
+} from '@angular/core';
+import { collapsiblePanelAnimation } from '@clr/angular/collapsible-panel';
+import { IfExpandService, uniqueIdFactory } from '@clr/angular/utils';
+import { Subscription } from 'rxjs';
+
+@Component({
+  selector: 'clr-card',
+  templateUrl: './card.html',
+  host: { '[class.card]': 'true', '[class.clr-card]': 'true' },
+  changeDetection: ChangeDetectionStrategy.OnPush,
```
dtsanevmw: **Animations:** this builds the new component on `collapsiblePanelAnimation` / `@toggle` / `@skipInitialRender`, which #2693 deprecates (Angular deprecated `@angular/animations` in 20.2 and we are moving everything to CSS). Shipping a brand new component that still requires `provideAnimations()` and then rewriting it in the next minor is avoidable, and the two PRs will conflict at merge.

I put together a prototype that does the collapse with CSS only, so it also works for the CSS-only card without the Angular components: https://github.com/vmware-clarity/ng-clarity/compare/vmladenov/collapsible-cards_cde-3120...dtsanevmw/festive-lamport-ieuinp

The idea: toggling `.card-collapsed` on a `.card.card-collapsible` transitions `.card-collapsible-content` between `grid-template-rows: 1fr` and `0fr` (same technique #2693 uses for tree node and vertical nav). No height measurement, no JS in the library, animates both directions, `prefers-reduced-motion` handled. The Angular component then just toggles the class plus `inert` / `aria-hidden` on the region, keeps the content in the DOM, and drops `animations`, `IfExpandService`, the subscriptions and the `NoopAnimationsModule` in specs. The chevron rotation keys off the toggle's `aria-expanded`, so ARIA state and visuals cannot drift apart. Feel free to take it as is or cherry-pick parts of it.

---

projects/angular/card/card.ts
```diff
@@ -0,0 +1,71 @@
...
+  @Output('clrCardCollapsedChange') collapsedChange = new EventEmitter<boolean>();
+
+  readonly cardId = uniqueIdFactory();
+  readonly headerId = `clr-card-header-${this.cardId}`;
+  readonly contentId = `clr-card-content-${this.cardId}`;
+
+  private subscription: Subscription;
+
+  constructor(
+    public expandService: IfExpandService,
```
dtsanevmw: **Public surface:** `expandService`, `cardId`, `headerId` and `contentId` are all public and end up in `card.api.md` (with an `ae-forgotten-export` warning for `IfExpandService`). The specs already rely on `cardInstance.expandService.toggle()`, which cements the service as API. Once this ships in a minor we cannot take it back.

Suggestion: keep the service private (or drop it, see the animations comment), expose a public `toggle()` on `ClrCard` for programmatic use, and keep only `collapsible`, `collapsed`, `footerCollapsible` and the output as the contract. Same for `ClrCardHeader`: `card` and `commonStrings` can be `protected`.

---

projects/angular/card/card.ts
```diff
@@ -0,0 +1,71 @@
...
+    private cdr: ChangeDetectorRef
+  ) {
+    expandService.expanded = true;
+  }
+
+  @Input('clrCardCollapsed')
```
dtsanevmw: **Bug (confirmed with a test):** this input has no `booleanAttribute` transform, so the bare attribute form `<clr-card clrCardCollapsible clrCardCollapsed>` receives `''`, `!''` is `true`, and the card renders expanded, while the sibling inputs accept the bare form. Please add `@Input({ alias: 'clrCardCollapsed', transform: booleanAttribute })` on the setter and a spec for the bare attribute.

**Naming (decide now, hard to change later):** every existing expand state in Clarity is positive: `clrExpanded`, `clrDgExpanded`, `clrSbExpanded`, `clrAccordionPanelOpen`. `clrCardCollapsed` inverts that convention. If there is a design reason to keep "collapsed" that is fine, but it should be a deliberate choice. Also `clr-card-img` is abbreviated while `clrCardMediaImage` is not.

---

projects/angular/card/card.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  ngOnInit() {
+    // The toggle button lives in the content-projected ClrCardHeader, a separate component, so a
+    // click there mutating this.expandService doesn't automatically mark this component dirty.
+    // Subscribing here ensures our own template (region visibility, animation trigger) refreshes.
+    this.subscription = this.expandService.expandChange.subscribe(expanded => {
+      this.collapsedChange.emit(!expanded);
+      this.cdr.markForCheck();
+    });
```
dtsanevmw: **Output echoes input changes (confirmed with a test):** setting `[clrCardCollapsed]` from the parent fires `clrCardCollapsedChange` synchronously during the parent's change detection, because the setter writes the service and this subscription emits for every change. A consumer doing analytics or a request in `(clrCardCollapsedChange)` gets it fired by their own programmatic change. Datagrid row and accordion only emit on user interaction (tree node does echo, so there is mixed precedent). For a new component I would emit only from the user toggle.

---

projects/angular/card/card-header.html
```diff
@@ -0,0 +1,24 @@
+<!--
+  ~ Copyright (c) 2016-2026 Broadcom. All Rights Reserved.
+  ~ The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+  ~ This software is released under MIT license.
+  ~ The full license information can be found in LICENSE in the root directory of this project.
+  -->
+@if (card?.collapsible) {
```
dtsanevmw: **Bug (confirmed with a test):** the header is OnPush and reads `card.collapsible` here, but nothing marks it dirty when the card's input changes. Flipping `[clrCardCollapsible]` from `false` to `true` at runtime renders the region on the card but no toggle button in the header (the `expandChange` subscription only covers the expanded state, not `collapsible`).

Simplest fix: keep the card state in signals and read them through the template (the prototype does this), which also removes both `markForCheck` subscriptions. Alternatively the card can `markForCheck()` its `@ContentChild(ClrCardHeader)` in `ngOnChanges`. Either way, please add a spec that toggles `clrCardCollapsible` at runtime.

---

dtsanevmw (changes requested):
Thanks for this, the sub-components and the overall shape are good. Since this is a new public component and everything we expose now is very hard to change in a minor release, I went through it strictly. The individual comments have the details, the short version:

**Confirmed bugs** (reproduced with specs on this branch): the header does not render the toggle when `clrCardCollapsible` is switched on at runtime, a bare `clrCardCollapsed` attribute leaves the card expanded, `clrCardCollapsedChange` fires on programmatic input changes, and the collapsed card renders a double border.

**Animations:** we are removing `@angular/animations` in #2693, so the card should not start its life on the deprecated `collapsiblePanelAnimation`. I pushed a prototype that does the collapse with CSS only so it also works for the CSS-only card, with a demo page and website example: https://github.com/vmware-clarity/ng-clarity/compare/vmladenov/collapsible-cards_cde-3120...dtsanevmw/festive-lamport-ieuinp (all 45 card specs pass there, API report regenerated). Take what is useful from it.

**API surface / a11y:** `expandService` and the ids should not be public, we need a heading level option like the accordion has, and the region / toggle accessible names need an inner header content element.

**Docs:** the demo mixes clickable and collapsible, JSDoc for the new i18n keys.

projects/angular/card/card.ts
dtsanevmw, line 27: **Animations:** this builds the new component on `collapsiblePanelAnimation` / `@toggle` / `@skipInitialRender`, which #2693 deprecates (Angular deprecated `@angular/animations` in 20.2 and we are moving everything to CSS). Shipping a brand new component that still requires `provideAnimations()` and then rewriting it in the next minor is avoidable, and the two PRs will conflict at merge.

I put together a prototype that does the collapse with CSS only, so it also works for the CSS-only card without the Angular components: https://github.com/vmware-clarity/ng-clarity/compare/vmladenov/collapsible-cards_cde-3120...dtsanevmw/festive-lamport-ieuinp

The idea: toggling `.card-collapsed` on a `.card.card-collapsible` transitions `.card-collapsible-content` between `grid-template-rows: 1fr` and `0fr` (same technique #2693 uses for tree node and vertical nav). No height measurement, no JS in the library, animates both directions, `prefers-reduced-motion` handled. The Angular component then just toggles the class plus `inert` / `aria-hidden` on the region, keeps the content in the DOM, and drops `animations`, `IfExpandService`, the subscriptions and the `NoopAnimationsModule` in specs. The chevron rotation keys off the toggle's `aria-expanded`, so ARIA state and visuals cannot drift apart. Feel free to take it as is or cherry-pick parts of it.

projects/angular/card/card.ts
dtsanevmw, line 44: **Public surface:** `expandService`, `cardId`, `headerId` and `contentId` are all public and end up in `card.api.md` (with an `ae-forgotten-export` warning for `IfExpandService`). The specs already rely on `cardInstance.expandService.toggle()`, which cements the service as API. Once this ships in a minor we cannot take it back.

Suggestion: keep the service private (or drop it, see the animations comment), expose a public `toggle()` on `ClrCard` for programmatic use, and keep only `collapsible`, `collapsed`, `footerCollapsible` and the output as the contract. Same for `ClrCardHeader`: `card` and `commonStrings` can be `protected`.

projects/angular/card/card.ts
dtsanevmw, line 50: **Bug (confirmed with a test):** this input has no `booleanAttribute` transform, so the bare attribute form `<clr-card clrCardCollapsible clrCardCollapsed>` receives `''`, `!''` is `true`, and the card renders expanded, while the sibling inputs accept the bare form. Please add `@Input({ alias: 'clrCardCollapsed', transform: booleanAttribute })` on the setter and a spec for the bare attribute.

**Naming (decide now, hard to change later):** every existing expand state in Clarity is positive: `clrExpanded`, `clrDgExpanded`, `clrSbExpanded`, `clrAccordionPanelOpen`. `clrCardCollapsed` inverts that convention. If there is a design reason to keep "collapsed" that is fine, but it should be a deliberate choice. Also `clr-card-img` is abbreviated while `clrCardMediaImage` is not.

projects/angular/card/card.ts
dtsanevmw, line 65: **Output echoes input changes (confirmed with a test):** setting `[clrCardCollapsed]` from the parent fires `clrCardCollapsedChange` synchronously during the parent's change detection, because the setter writes the service and this subscription emits for every change. A consumer doing analytics or a request in `(clrCardCollapsedChange)` gets it fired by their own programmatic change. Datagrid row and accordion only emit on user interaction (tree node does echo, so there is mixed precedent). For a new component I would emit only from the user toggle.

projects/angular/card/card-header.html
dtsanevmw, line 7: **Bug (confirmed with a test):** the header is OnPush and reads `card.collapsible` here, but nothing marks it dirty when the card's input changes. Flipping `[clrCardCollapsible]` from `false` to `true` at runtime renders the region on the card but no toggle button in the header (the `expandChange` subscription only covers the expanded state, not `collapsible`).

Simplest fix: keep the card state in signals and read them through the template (the prototype does this), which also removes both `markForCheck` subscriptions. Alternatively the card can `markForCheck()` its `@ContentChild(ClrCardHeader)` in `ngOnChanges`. Either way, please add a spec that toggles `clrCardCollapsible` at runtime.

projects/angular/card/card.html
```diff
@@ -0,0 +1,41 @@
...
+  <ng-content select="clr-card-footer"></ng-content>
+</ng-template>
+@if (collapsible) {
+<div
+  @skipInitialRender
+  role="region"
+  class="clr-card-collapsible-region"
+  [id]="contentId"
+  [attr.aria-hidden]="!expandService.expanded"
+  [attr.aria-labelledby]="headerId"
```
dtsanevmw, line 21: **Accessible names:** `aria-labelledby` points at the whole `clr-card-header`, which contains the toggle button, so the region's computed name is "Collapse card Header". Every toggle on a page is also named identically "Expand card" / "Collapse card", so a screen reader user listing buttons cannot tell the cards apart.

Suggestion: wrap the projected header content in an inner element with its own id, point `aria-labelledby` at that, and add `aria-describedby` from the toggle button to it. That inner wrapper also fixes the header layout issue (see the SCSS comment).

**Heading semantics:** the CSS card used `h3.card-header` and `h4.card-title`; the new elements have no heading role, and nesting an `<h3>` inside `clr-card-header` picks up the global heading margins instead of the card typography. The accordion solved this with `clrAccordionPanelHeadingLevel` rendering `role="heading"` + `aria-level`. The card needs the same for WCAG heading navigation, ideally on both `clr-card-header` and `clr-card-body-title`.

projects/angular/card/_card-elements.clarity.scss
```diff
@@ -0,0 +1,74 @@
...
+    display: block;
+    color: tokens.$cds-alias-object-interaction-color;
+    transition: all tokens.$cds-global-animation-duration-quick tokens.$cds-global-animation-easing-in-out;
+    transform: rotate(90deg);
+
+    @include mixins.min-equilateral(density.$clr-base-icon-size-m);
+
+    &.expanded {
+      transform: rotate(180deg);
+    }
```
dtsanevmw, line 72: A few things in the styles:

- **Header becomes a flex container for every card**, not only component ones. Text and inline children (a label, a badge) become separate flex items and stop flowing inline, which differs from the CSS card. Wrapping the projected content in an inner element with `flex: 1 1 auto; min-width: 0` keeps normal inline flow.
- **Double border when collapsed:** the header keeps its `border-bottom` because the empty region follows it, so the collapsed snapshot has two border rows at the bottom of the card versus one in every other state (checked pixel by pixel on the chromium snapshot). Needs a rule for the collapsed state; easiest with a state class on the card.
- **Header height changes when collapsible:** the toggle padding + icon make the collapsible header 4px taller than a plain one (snapshot rows 105 vs 109), so cards side by side with mixed settings misalign. A `padding: 0` inline-flex button with the icon sized to fit the line height avoids that.
- Commented-out `&:focus` block should either be a real `:focus-visible` style or removed. With `border: 0; background: transparent` the button still needs a visible focus indicator.
- `transition: all` should be `transition: transform`.
- If the CSS-only card is the base (see the animations comment), the collapsible structure and state classes should live in the unprefixed `card-*` family in `layout/_card.clarity.scss`, so both variants share one stylesheet and one doc table. The `clr-card-*` host duplicates then only exist as Angular hooks.

All of these are addressed in the prototype branch if you want to compare.

projects/demo/src/app/card/angular/card-angular.html
```diff
@@ -0,0 +1,266 @@
...
+<h3>Angular component</h3>
+
+<h4>Collapsible card</h4>
+
+<p>Current state: {{ collapsed ? 'collapsed' : 'expanded' }}</p>
+
+<div class="clr-example nomargin">
+  <div class="clr-row">
+    <div class="clr-col-lg-5 clr-col-md-8 clr-col-12">
+      <clr-card class="clickable" clrCardCollapsible [(clrCardCollapsed)]="collapsed">
```
dtsanevmw, line 17: This demo combines `class="clickable"` with a collapsible card that contains a footer button and an app-level alert inside the header. The docs on the same page say clickable cards must not contain clickable elements, and an app-level alert does not belong in a card header. Suggest a plain collapsible card here (and maybe a separate "alert in a collapsible card" example with a regular alert in the body if that is the scenario you wanted to show).

projects/angular/utils/i18n/common-strings.interface.ts
```diff
@@ -361,4 +361,8 @@ export interface ClrCommonStrings {
   // Breadcrumbs
   breadcrumbsLabel: string;
   expandBreadcrumbsLabel: string;
+
+  // Card
+  expandCardAriaLabel: string;
+  collapseCardAriaLabel: string;
```
dtsanevmw, line 367: Nit: please add JSDoc for the two new keys like the documented ones above (they show up as `(undocumented)` in `utils.api.md`). Something like "Aria label of the toggle button of a collapsible card while it is collapsed / expanded".

projects/angular/card/card.spec.ts
```diff
@@ -0,0 +1,244 @@
+/*
```
dtsanevmw, line 1: Missing coverage that would have caught the issues above:

- `clrCardCollapsed` set as a bare attribute
- `clrCardCollapsible` toggled at runtime (region and header toggle appear)
- `clrCardCollapsedChange` not firing when the input is changed programmatically
- ids unique across two cards on the same page
- initial render with `clrCardCollapsed` true

The prototype branch has versions of these you can reuse.

dtsanevmw (commented):
As this is a new component and we already heading towards zoneless, signals, standalone components, please look forward this direction so we don't have to re-work this component again once we get to this point.

projects/angular/utils/i18n/common-strings.interface.ts
valentin-mladenov: Corrected in commit a59c483.

projects/demo/src/app/card/angular/card-angular.html
valentin-mladenov: Corrected in commit a59c483.

projects/angular/card/_card-elements.clarity.scss
valentin-mladenov: Corrected in commit a59c483.

projects/angular/card/card.html
valentin-mladenov: Corrected in commit a59c483.

projects/angular/card/card-header.html
valentin-mladenov: Corrected in commit a59c483.

projects/angular/card/card.ts
valentin-mladenov: Corrected in commit a59c483.

projects/angular/card/card.ts
valentin-mladenov: Corrected in commit a59c483.

projects/angular/card/card.ts
valentin-mladenov: Corrected in commit a59c483.

projects/angular/card/card.ts
valentin-mladenov: Corrected in commit a59c483.

projects/angular/card/card.spec.ts
valentin-mladenov: Corrected in commit a59c483.

valentin-mladenov:
> As this is a new component and we already heading towards zoneless, signals, standalone components, please look forward this direction so we don't have to re-work this component again once we get to this point.

The card components are zoneless, use signals, but are not a standlaone. From my current point of view keeping them as module make more sense, because are tightly coupled.

dtsanevmw (approved):

